PR #3, opened 2021-09-06
OEL-450: Theme the site branding block.

abel-santos-corral:
Add twig to override block--system-branding-block and add classes to it.


escuriola (dismissed):

tests/Functional/SiteBrandingBlockTest.php
```diff
@@ -0,0 +1,68 @@
+<?php
```
drishu: Don't use a functional test just to check some classes and/or markup.
Please use something similar to https://github.com/openeuropa/oe_theme/blob/2.x/tests/Kernel/RenderingTest.php#L103

drishu (changes requested):

tests/Functional/SiteBrandingBlockTest.php
drishu: in the end that one was incompatible with blocks

drishu (approved):

escuriola (approved):

PR #9, opened 2021-09-14
Oel 462: facet list template overriden.

Maxfire:


tests/src/Kernel/SearchFacetsTest.php
```diff
@@ -0,0 +1,72 @@
...
+    'facets',
+    'facets_summary',
+  ];
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp(): void {
+    parent::setUp();
+    /** @var \Drupal\Core\Extension\ThemeInstallerInterface $theme_installer */
```
drishu: this is not needed

tests/src/Kernel/SearchFacetsTest.php
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  /**
+   * Tests facets summary.
+   *
+   * @covers ::setFacets
+   * @covers ::getFacets
+   * @covers ::removeFacet
+   */
+  public function testFacets() {
```
drishu:
```suggestion
  public function testFacetsRendering(): void {
```

tests/src/Kernel/SearchFacetsTest.php
```diff
@@ -0,0 +1,72 @@
...
+    $entity = new FacetsSummary(['description' => 'Owls', 'name' => 'owl'], 'facets_summary');
+
+    $this->assertEmpty($entity->getFacets());
+
+    $facets = ['foo' => 'bar'];
+    $entity->setFacets($facets);
+    $this->assertEquals($facets, $entity->getFacets());
```
drishu: we need to render the build of the facets summary block, and assert the markup with Crawler

drishu (changes requested):
see the Sitebranding as an example or any of the other

templates/overrides/search/facets-summary-item-list.html.twig
```diff
@@ -0,0 +1,20 @@
+{#
+/**
+ * @file
+ * Template for a facet summary item list.
+ */
+#}
+{%- if items -%}
+  {%- for item in items -%}
+    {% if item.value['#type'] == 'link' %}
+      {% include '@oe-bcl/bcl-badge/badge.html.twig' with {
```
drishu:
```
{{ pattern('badge', {
      'label': 'label',
      'assistive_text': 'assistive_text',
      'url': '#',
      'icons_path': bcl_icon_path,
      'dismissible': true
    }) }}
```

drishu (changes requested):

tests/src/Kernel/SearchFacetsTest.php
Maxfire: Tests were excluded from this task as agreed.

tests/src/Kernel/SearchFacetsTest.php
Maxfire: Tests were excluded from this task as agreed.

tests/src/Kernel/SearchFacetsTest.php
Maxfire: Tests were excluded from this task as agreed.

templates/overrides/search/facets-summary-item-list.html.twig
Maxfire: :+1:  done.

GilNovacomm (approved):
the summary title is not in the scope of this ticket, the current implementation is correct.

drishu (approved):

PR #12, opened 2021-09-14
OEL-463: Pager templates overriden and tests added accordingly.

Maxfire:


tests/src/Kernel/PagerKernelTest.php
```diff
@@ -0,0 +1,107 @@
...
+    $view->displayHandlers->get('default')->setOption('pager', [
+      'type' => 'mini',
+      'options' => ['items_per_page' => 5],
+    ]);
+    $view->storage->save();
+
+    $output = $view->preview();
+
+    \Drupal::service('renderer')->renderPlain($output);
+    $this->assertSame(CacheBackendInterface::CACHE_PERMANENT, $output['#cache']['max-age']);
```
drishu: we dont need to check for views internals, we check our rendering
remve everything cache related

drishu (changes requested):
Actually, dont test at all yet, we will have a separate ticket for the rendering test in oe_bootstrap_theme;
Please move this code into oe_bootstrap_theme

tests/src/Kernel/PagerKernelTest.php
Maxfire: Hi Andras, I removed it from the code in order to push it in the oe_bootstrap_theme repo.
Thanks.

PR #19, opened 2021-09-17
OEL-002: Fix summary results and facets badges.

GilNovacomm:
Added two new templates for facets:
Results (badges)
Count
and fixed the template for summary item list.


drishu (approved):

Maxfire (approved):
:+1:

PR #22, opened 2021-09-23
OEL-001: Better badge styling in summary.

drishu:


templates/overrides/search/facets-result-item.html.twig
```diff
@@ -16,10 +16,13 @@
  * @ingroup themeable
  */
 #}
+{% set attributes = create_attribute() %}
+{% set attributes = attributes.addClass('me-2') %}
```
Maxfire: Suggestion, you could do both to simplify as follows:
 set attributes = create_attribute({'class': ['me-2']})

abel-santos-corral (approved):

Maxfire (approved):

PR #27, opened 2021-10-05
OEL-543: Styling components.

escuriola:
Jira issue:
- https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-543?


abel-santos-corral (dismissed):
Reviewed. Checked that all changes required at ticket were on place on the form at search page.

oe_whitelabel.theme
```diff
@@ -64,3 +66,22 @@ function oe_whitelabel_preprocess_links__language_block(&$variables) {
     ],
   ];
 }
+
+/**
+ * Implements hook_form_alter().
+ */
+function oe_whitelabel_form_alter(&$form, FormStateInterface $form_state, $form_id) {
```
drishu: I'm thinking that adding classes might change other forms in ways we might not want, lets use hook_form_FORM_ID_alter to target the facets form specifically

oe_whitelabel.theme
```diff
@@ -64,3 +66,22 @@ function oe_whitelabel_preprocess_links__language_block(&$variables) {
...
+  if (isset($form['actions'])) {
+    $form['actions']['#attributes']['class'][] = 'mt-4';
+  }
+  if (isset($form['actions']['submit'])) {
+    $form['actions']['submit']['#attributes']['class'][] = 'me-2';
+    array_push($form['actions']['submit']['#attributes']['class'], 'btn', 'btn-primary');
+    $form['actions']['submit']['#attributes']['class'] = array_unique($form['actions']['submit']['#attributes']['class']);
```
drishu: the browser can handle duplicates, if there are many, then the source of the duplication needs to be fixed

oe_whitelabel.theme
```diff
@@ -64,3 +66,22 @@ function oe_whitelabel_preprocess_links__language_block(&$variables) {
...
+  }
+  if (isset($form['actions']['reset'])) {
+    $form['actions']['reset']['#attributes']['class'][] = 'btn';
+    $form['actions']['reset']['#attributes']['class'][] = 'btn-secondary';
+    $form['actions']['reset']['#attributes']['class'] = array_unique($form['actions']['reset']['#attributes']['class']);
```
drishu: the browser can handle duplicates, if there are many, then the source of the duplication needs to be fixed

drishu (changes requested):

composer.json
```diff
@@ -7,6 +7,7 @@
     "prefer-stable": true,
     "require": {
         "php": ">=7.3",
+        "drupal/better_exposed_filters": "^5.0@beta",
```
drishu: I don't think this should be a hard requirement, rather require-dev

modules/oe_whitelabel_helper/src/Plugin/better_exposed_filters/sort/FloatEndSortWidget.php
```diff
@@ -0,0 +1,29 @@
...
+namespace Drupal\oe_whitelabel_helper\Plugin\better_exposed_filters\sort;
+
+use Drupal\better_exposed_filters\Plugin\better_exposed_filters\sort\DefaultWidget;
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * OpenEuropa custom better exposed filters widget implementation.
+ *
+ * @BetterExposedFiltersSortWidget(
+ *   id = "float_end_sort",
```
drishu: prefix with oe_whitelabel

modules/oe_whitelabel_helper/src/Plugin/better_exposed_filters/sort/FloatEndSortWidget.php
```diff
@@ -0,0 +1,29 @@
...
+ *   label = @Translation("Float End Sort Better Exposed Filter"),
```
drishu:
```suggestion
 *   label = @Translation("Float End Sort"),
```

drishu (changes requested):
- we now have oe_whitelabel_search, we should move this widget into there
- add a kernel test for the widget

drishu (dismissed):

drishu (dismissed):

drishu (approved):

GilNovacomm (approved):

PR #28, opened 2021-10-06
OEL-534: Counter to 0 included in the template.

Maxfire:
Fixes:
 - https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-534

### DIFF
--- a/templates/overrides/search/facets-summary-count.html.twig
+++ b/templates/overrides/search/facets-summary-count.html.twig
@@ -4,5 +4,4 @@
  * Default theme implementation to show facets summary count.
  */
 #}
-{% set count = count ?? 0 %}
 <span class="source-summary-count">({{ count }})</span>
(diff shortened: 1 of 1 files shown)


abel-santos-corral (dismissed):
Reviewed and tested. When no element is found the total live summary is set to 0 instead of empty.

templates/overrides/search/facets-summary-count.html.twig
```diff
@@ -4,6 +4,5 @@
  * Default theme implementation to show facets summary count.
  */
 #}
-{% if count %}
-  <span class="source-summary-count">({{ count }})</span>
-{% endif %}
+{% set count = count ?? 0 %}
```
drishu: if the result from our live total is 0 when it should be, then there is not point in this check

drishu (changes requested):

drishu (approved):

abel-santos-corral (approved):

PR #30, opened 2021-10-11
OEL-587: Update header template.

escuriola:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-587


abel-santos-corral (dismissed):
For this part of the change is ok, reviewed and added a phpunit.xml.dist change. Waiting for response to know if we can activate at runner.yml modules (views, views_ui and menu_ui).

abel-santos-corral (dismissed):

PR #31, opened 2021-10-12
OEL-589: Adapt footer to BCL 0.8.0.

escuriola:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-589

### DIFF
--- a/modules/oe_whitelabel_search/src/Form/SearchForm.php
+++ b/modules/oe_whitelabel_search/src/Form/SearchForm.php
@@ -58,6 +58,7 @@ class SearchForm extends FormBase {
       '#type' => 'textfield',
       '#title' => $config['input']['label'],
       '#title_display' => 'invisible',
+      '#size' => 20,
       '#attributes' => [
         'placeholder' => $config['input']['placeholder'],
         'class' => [


templates/patterns/footer/pattern-footer.html.twig
```diff
@@ -10,9 +10,23 @@
...
+          contact_eu_links: section.contact_eu_links,
+        } %}
+      </div>
+    {% endfor %}
+  </div>
+{% endif %}
 {% if bottom_sections is defined and bottom_sections is not empty and bottom_sections is iterable %}
-  <hr/>
-  <div class="row">
+  <div class="pb-4 pb-lg-5 mt-4 mt-lg-5 bcl-footer__bordered-row row">
```
abel-santos-corral: This is used 2 times, it can be defined as a set classes = [values] and then add the classes variable here.

templates/patterns/footer/footer.ui_patterns.yml
```diff
@@ -8,3 +8,6 @@ footer:
     bottom_sections:
       type: array
       label: Sections in the footer
+    middle_sections:
+      type: array
+      label: Sections in the footer
```
abel-santos-corral: Change the labels, all have same label when they actually are bottom and middle sections.

abel-santos-corral (changes requested):

abel-santos-corral (dismissed):

PR #34, opened 2021-10-18
OEL-648: BCL 0.9.0 Language switch modal adaptation.

escuriola:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-648


abel-santos-corral (commented):
Please Carlos, confirm with Andras because it seems to me that there's no need to put the different EU and Non-EU languages.

abel-santos-corral (changes requested):
Please review comments.

abel-santos-corral (approved):
Reviewed.

PR #35, opened 2021-10-18
OEL-547: Add links to corporate logos.

escuriola:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-547

### DIFF
--- a/modules/oe_whitelabel_search/src/Form/SearchForm.php
+++ b/modules/oe_whitelabel_search/src/Form/SearchForm.php
@@ -58,6 +58,7 @@ class SearchForm extends FormBase {
       '#type' => 'textfield',
       '#title' => $config['input']['label'],
       '#title_display' => 'invisible',
+      '#size' => 20,
       '#attributes' => [
         'placeholder' => $config['input']['placeholder'],
         'class' => [


modules/oe_whitelabel_helper/src/Plugin/Block/CorporateEcLogoBlock.php
```diff
@@ -80,15 +81,26 @@ public function build(): array {
...
       '#uri' => $logo_path . '/logo-ec--' . $language . '.svg',
-      '#width' => '240px',
-      '#height' => '60px',
       '#alt' => $title,
       '#title' => $title,
     ];
 
+    $build = [
+      '#type' => 'link',
+      '#url' => Url::fromUri('https://ec.europa.eu/info/index_' . $language, [
```
abel-santos-corral: From a maintenance point of view, this path should be placed on a configuration settings. What if the url changes? Then there should be change at git level to put new value here. How is done in OE modules and themes?

modules/oe_whitelabel_helper/src/Plugin/Block/CorporateEuLogoBlock.php
```diff
@@ -83,19 +84,21 @@ public function build(): array {
...
+
     $build = [
-      '#type' => 'inline_template',
-      '#template' => '<picture><source media="(max-width: 25em)" srcset="{{ mobile }}">{{ image }}</picture>',
-      '#context' => [
-        'image' => $image,
-        'mobile' => file_url_transform_relative($mobile_url),
-      ],
+      '#type' => 'link',
+      '#url' => Url::fromUri('https://europa.eu/european-union/index_' . $language, [
```
abel-santos-corral: Same as above. Would be good to consider to use a path set at config and not hard-coded. Let's see if this can be set as config.

PR #38, opened 2021-10-20
OEL-601: Theming search box button with icon

escuriola:
Jira:
- https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-601

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         "drupal/core": "^8.9 || ^9.1",
         "drupal/ui_patterns": "^1.0",
         "drupal/ui_patterns_settings": "^1.0",
-        "openeuropa/oe_bootstrap_theme": "OEL-601"
+        "openeuropa/oe_bootstrap_theme": "dev-OEL-601"
     },
     "require-dev": {
         "composer/installers": "^1.11",
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         "drupal/core": "^8.9 || ^9.1",
         "drupal/ui_patterns": "^1.0",
         "drupal/ui_patterns_settings": "^1.0",
-        "openeuropa/oe_bootstrap_theme": "dev-OEL-601"
+        "openeuropa/oe_bootstrap_theme": "1.0.0-alpha4"
     },
     "require-dev": {
         "composer/installers": "^1.11",
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         "drupal/core": "^8.9 || ^9.1",
         "drupal/ui_patterns": "^1.0",
         "drupal/ui_patterns_settings": "^1.0",
-        "openeuropa/oe_bootstrap_theme": "1.0.0-alpha4"
+        "openeuropa/oe_bootstrap_theme": "OEL-601"
     },
     "require-dev": {
         "composer/installers": "^1.11",
--- a/modules/oe_whitelabel_search/src/Form/SearchForm.php
+++ b/modules/oe_whitelabel_search/src/Form/SearchForm.php
@@ -95,6 +95,7 @@ class SearchForm extends FormBase {
             'd-flex',
             'btn btn-light',
             'btn-md',
+            'py-2',
           ],
         ],
       ],
--- a/modules/oe_whitelabel_search/src/Form/SearchForm.php
+++ b/modules/oe_whitelabel_search/src/Form/SearchForm.php
@@ -9,6 +9,8 @@ use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Language\LanguageManagerInterface;
 use Drupal\Core\Url;
 use Symfony\Component\DependencyInjection\ContainerInterface;
+use Drupal\Core\Template\Attribute;
+use Drupal\Core\Routing\TrustedRedirectResponse;
 
 /**
  * Configurable search form class.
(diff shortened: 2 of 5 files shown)


modules/oe_whitelabel_search/src/Form/SearchForm.php
```diff
@@ -48,40 +48,62 @@ public function buildForm(array $form, FormStateInterface $form_state, array $co
...
-      '#prefix' => '<div class="ms-2">',
-      '#suffix' => '</div>',
-      '#type' => 'submit',
-      '#name' => FALSE,
-      '#value' => $config['button']['label'],
-      '#attributes' => [
-        'class' => [
-          'btn-md',
-          $config['button']['classes'],
+    $button = [
```
abel-santos-corral: We have to research on this solution. In case finally we succeed on using this approach, here must be added a comment explaining that this approach is used only here.

modules/oe_whitelabel_search/src/Plugin/Block/SearchBlock.php
```diff
@@ -138,13 +138,7 @@ public function blockForm($form, FormStateInterface $form_state): array {
       '#tree' => TRUE,
       '#description' => $this->t('Fill in the settings of the Button field.'),
     ];
-    $form['button']['button_label'] = [
```
abel-santos-corral, line 141: Better put label back and add the config to decide upon three cases: icon, label and icon+label. Regarding this, move the value at the pattern above.

modules/oe_whitelabel_search/src/Form/SearchForm.php
```diff
@@ -48,40 +48,62 @@ public function buildForm(array $form, FormStateInterface $form_state, array $co
...
+      '#type' => 'pattern',
+      '#id' => 'button',
+      '#variant' => 'light',
+      '#fields' => [
+        'settings' => [
+          'type' => 'submit',
+        ],
+        'icon' => [
+          'name' => 'search',
+          'size' => 'xs',
```
abel-santos-corral, line 86: Add this here to check if attributes are working:
'attributes' => new Attribute('class', ['align-self-center']),

At the beginning of file add:
use Drupal\Core\Template\Attribute;

modules/oe_whitelabel_search/src/Form/SearchForm.php
```diff
@@ -48,40 +48,62 @@ public function buildForm(array $form, FormStateInterface $form_state, array $co
...
+    $button = [
+      '#type' => 'pattern',
+      '#id' => 'button',
+      '#variant' => 'light',
+      '#fields' => [
+        'settings' => [
+          'type' => 'submit',
+        ],
+        'icon' => [
+          'name' => 'search',
```
abel-santos-corral, line 85: Add a field in the block for the name of the icon. Remember to add it too at the tests and at the schema.

abel-santos-corral (changes requested):
As commented, please check how to manage variants (icon, label, and icon+label). In case label is brought back, please update the schema and test for the block.

composer.json
```diff
@@ -11,7 +11,7 @@
         "drupal/core": "^8.9 || ^9.1",
         "drupal/ui_patterns": "^1.0",
         "drupal/ui_patterns_settings": "^1.0",
-        "openeuropa/oe_bootstrap_theme": "1.0.0-alpha4"
+        "openeuropa/oe_bootstrap_theme": "dev-OEL-601"
```
drishu: if this PR overrides the input submit then we can't use it

modules/oe_whitelabel_search/src/Form/SearchForm.php
drishu:
```suggestion
   $form['submit'] = [
      '#input' => TRUE,
      '#is_button' => TRUE,
      '#executes_submit_callback' => TRUE,
      '#type' => 'pattern',
      '#id' => 'button',
      '#variant' => 'light',
      '#fields' => [
        'settings' => [
          'type' => 'submit',
        ],
        'attributes' => [
          'id' => 'submit',
          'class' => [
            'border-start-0',
            'rounded-0 rounded-end',
            'd-flex',
            'btn btn-light',
            'btn-md',
            'py-2',
            $config['button']['classes'],
          ],
        ],
      ],
    ];
```

notice the first 3 keys, those will tell drupal this is a form element that submitted the form

modules/oe_whitelabel_search/src/Form/SearchForm.php
```diff
@@ -48,13 +48,17 @@ public function buildForm(array $form, FormStateInterface $form_state, array $co
     $form_state->set('oe_whitelabel_search_config', $config);
     $input_value = '';
 
+    $form['#attributes']['class'][] = 'bcl_search_form';
+
     if (!empty($config['input']['name'])) {
       $input_value = $this->getRequest()->get($config['input']['name']);
     }
 
     $form['#action'] = $config['form']['action'];
```
drishu: remove this, it works better without it

modules/oe_whitelabel_search/src/Form/SearchForm.php
```diff
@@ -63,26 +67,51 @@ public function buildForm(array $form, FormStateInterface $form_state, array $co
...
+      $button['#fields']['label'] = $config['button']['label'];
+    }
+    if ($config['button']['label_icon'] != 'label') {
+      $button['#fields']['icon'] = [
+        'name' => 'search',
+        'size' => 'xs',
+      ];
+    }
+
+    $form['submit'] = $button;
```
drishu: I dont see the value in this, you can use $form['submit'] = [.. directly

drishu (changes requested):

drishu (approved):

abel-santos-corral (approved):

PR #41, opened 2021-10-29
OEL-726: Add header light/standard configuration and apply to pattern.

escuriola:
- https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-726

Also upgrade oe_bootstrap_theme to BCL 0.11.0

### DIFF
--- a/modules/oe_whitelabel_search/src/Plugin/Block/SearchBlock.php
+++ b/modules/oe_whitelabel_search/src/Plugin/Block/SearchBlock.php
@@ -129,8 +129,7 @@ class SearchBlock extends BlockBase implements ContainerFactoryPluginInterface {
       '#type' => 'textfield',
       '#title' => $this->t('Input placeholder text'),
       '#description' => $this->t('The placeholder that will be shown inside the input field.'),
-      '#default_value' => $this->t('Search'),
-      '#value' => $config['input']['placeholder'],
+      '#default_value' => $config['input']['placeholder'],
     ];
     $form['button'] = [
       '#type' => 'details',
--- a/modules/oe_whitelabel_search/src/Plugin/Block/SearchBlock.php
+++ b/modules/oe_whitelabel_search/src/Plugin/Block/SearchBlock.php
@@ -129,7 +129,7 @@ class SearchBlock extends BlockBase implements ContainerFactoryPluginInterface {
       '#type' => 'textfield',
       '#title' => $this->t('Input placeholder text'),
       '#description' => $this->t('The placeholder that will be shown inside the input field.'),
-      '#default_value' => $config['input']['placeholder'],
+      '#default_value' => $config['input']['placeholder'] ?? $this->t('Search'),
     ];
     $form['button'] = [
       '#type' => 'details',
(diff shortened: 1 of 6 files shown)


modules/oe_whitelabel_search/src/Plugin/Block/SearchBlock.php
```diff
@@ -129,8 +129,7 @@ public function blockForm($form, FormStateInterface $form_state): array {
       '#type' => 'textfield',
       '#title' => $this->t('Input placeholder text'),
       '#description' => $this->t('The placeholder that will be shown inside the input field.'),
-      '#default_value' => $this->t('Search'),
-      '#value' => $config['input']['placeholder'],
+      '#default_value' => $config['input']['placeholder'],
```
drishu: I think a default value is fine

oe_whitelabel.theme
```diff
@@ -111,11 +111,22 @@ function oe_whitelabel_form_system_theme_settings_alter(&$form, FormStateInterfa
...
   ];
+  $form['header_style'] = [
+    '#type' => 'select',
+    '#title' => t('Background style'),
+    '#options' => [
+      'standard' => t('Blue header'),
+      'light' => t('Light header'),
+    ],
+    '#default_value' => theme_get_setting('header_style'),
+    '#description' => t('Choose "European Commission" for websites hosted under the ec.europa.eu domain and "European Union" for websites hosted under the europa.eu domain.'),
```
drishu: is this the right description for this option/setting ?

templates/overrides/navigation/block--system-branding-block.html.twig
```diff
@@ -20,7 +20,8 @@
     </a>
   {% endif %}
   {% if site_name %}
-    <p class="site-name"><a class="text-white text-decoration-none align-bottom" href="{{ path('<front>') }}" rel="home">{{ site_name }}</a></p>
+  {% set text_color_class = bcl_header_style =='light' ? 'text-dark': 'text-white' %}
```
drishu: Since we start adding classes dynamically it starts making sense to create a attributes object and use addClass method, add by default "ext-decoration-none align-bottom" and put the if after adding correct text- class.
Add also href, rel.

templates/overrides/navigation/block--system-branding-block.html.twig
```diff
@@ -20,7 +20,8 @@
...
+    <p class="site-name"><a class="{{ text_color_class }} text-decoration-none align-bottom" href="{{ path('<front>') }}" rel="home">{{ site_name }}</a></p>
```
drishu: after we contruct the attributes, use it here instead of all this

templates/patterns/header/pattern-header--variant-ec.html.twig
```diff
@@ -7,7 +7,11 @@
       </ul>
     </div>
   </nav>
-  <div class="bcl-header__project">
+  {% set _project_classes = 'bcl-header__project' %}
```
drishu: lets use attributes class instead of this

templates/patterns/header/pattern-header--variant-eu.html.twig
```diff
@@ -7,7 +7,11 @@
       </ul>
     </div>
   </nav>
-  <div class="bcl-header__project">
+  {% set _project_classes = 'bcl-header__project' %}
```
drishu: I think we have a lot of duplicate code here, lets take the chance and use extend template and block content

drishu (changes requested):
testing we will do in the dedicated ticket

drishu (approved):

Maxfire (approved):

PR #48, opened 2021-11-18
OEL-477: Create User profile card pattern, and field group custom form…

GilNovacomm:
Override the card pattern for the user profile.
Create a custom Field Formatter for field group to print fields evenly in two columns


modules/oe_whitelabel_helper/src/Plugin/field_group/FieldGroupFormatter/MultiColumn.php
```diff
@@ -0,0 +1,266 @@
...
+    $element['fieldset']['data']['first_column']['#attributes'] = [
+      'class' => $this->getSetting('first_column') ? $first_column_classes : $classes,
+    ];
+    $element['fieldset']['data']['second_column']['#type'] = 'container';
+    $element['fieldset']['data']['second_column']['#attributes'] = [
+      'class' => $this->getSetting('second_column') ? $second_column_classes : $classes,
+    ];
+
+    // Number of items in the first columns.
+    $first_column_elements = array_slice($children, 0, ceil(count($children) / 2));
```
Maxfire, line 203: I think ceil is waiting for a float not an int

modules/oe_whitelabel_helper/oe_whitelabel_helper.module
```diff
@@ -7,9 +7,81 @@
...
+
+/**
+ * Implements hook_entity_display_build_alter().
+ */
+function oe_whitelabel_helper_entity_display_build_alter(&$build, $context) {
+  $entity = $context['entity'];
+
+  // Append RDF term mappings on displayed taxonomy links.
+  foreach (Element::children($build) as $field_name) {
+    $field = &$build[$field_name];
```
Maxfire: If it's not returning anything, is this being used?

Maxfire (changes requested):
Small changes.

modules/oe_whitelabel_helper/src/Plugin/field_group/FieldGroupFormatter/MultiColumn.php
```diff
@@ -0,0 +1,266 @@
...
+    $summary = parent::settingsSummary();
+    $summary[] = $this->t('Display results as a 2 column table.');
+
+    return $summary;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function preRender(&$element, $rendering_object): void {
```
Maxfire, line 149: We have an issue with this according to the complexity, please have a look at https://drone.fpfis.eu/openeuropa/oe_whitelabel/483/20

modules/oe_whitelabel_helper/src/Plugin/field_group/FieldGroupFormatter/MultiColumn.php
```diff
@@ -0,0 +1,266 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function defaultContextSettings($context): array {
+    return [
+      'first_column' => '',
+      'second_column' => '',
+      'hide_table_if_empty' => FALSE,
+    ] + parent::defaultSettings($context);
```
Maxfire: The parent doesn't need a variable, do we need it here? just a question.

Maxfire (changes requested):

modules/oe_whitelabel_helper/src/Plugin/field_group/FieldGroupFormatter/MultiColumn.php
GilNovacomm: In fact ceil can receive both int and floats, but in this case ceil is rounding up for then we have un-even elements.

PR #55, opened 2021-12-20
OEL-772: Fix EC logo not showing in mobile

escuriola:
- https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-772


drishu (approved):
we will account for the tests in OEL-486 so its ok for me

Maxfire (approved):

PR #69, opened 2022-02-11
OEL-1147: Footer refactor

escuriola:


modules/oe_whitelabel_helper/src/Plugin/Block/NeutralFooterBlock.php
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Provides the Neutral footer block.
+ *
+ * @Block(
+ *   id = "oe_corporate_blocks_neutral_footer",
+ *   admin_label = @Translation("Neutral Footer block"),
+ *   category = @Translation("Corporate blocks"),
+ * )
+ */
+class NeutralFooterBlock extends FooterBlockBase implements ContainerFactoryPluginInterface {
```
drishu: I like the block, but now our helper module is now using (depending on) classes from oe_corporate_blocks, so add the dependency to its info file

modules/oe_whitelabel_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -45,6 +46,7 @@ public function getFunctions(): array {
     return [
       new TwigFunction('bcl_footer_links', [$this, 'bclFooterLinks'], ['needs_context' => TRUE]),
       new TwigFunction('bcl_block', [$this, 'bclBlock']),
```
drishu: I think this is a good occasion to deprecate bcl_block and use twig_tweak::drupal_block

modules/oe_whitelabel_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -105,6 +107,22 @@ public function bclTimeAgo(string $timestamp): PluralTranslatableMarkup {
...
+   * @param string $title
+   *   The title to be formatted.
+   * @param string $classes
+   *   Classes to add to the title.
+   *
+   * @return array
+   *   Title prepared to be rendered in the footer template.
+   */
+  public function bclTitle(string $title, string $classes): array {
+    $title = \Drupal::translation()->translate($title)->render();
```
drishu: with or without this line, same result

modules/oe_whitelabel_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -105,6 +107,22 @@ public function bclTimeAgo(string $timestamp): PluralTranslatableMarkup {
...
+    return ['#markup' => '<p class="' . $classes . '">' . $title . '</p>'];
```
drishu: Since this is used in twig we can use https://www.drupal.org/node/2311123, BUT the real question is: why not use a macro ?
When the function just wraps stuff into html then its a good candidate for a macro
```
{% macro bcl_title(title, classes) %}
  <p class="{{ classes }}">{{ title }}</p>
{% endmacro %}
```

oe_whitelabel.theme
```diff
@@ -105,10 +105,11 @@ function oe_whitelabel_form_system_theme_settings_alter(&$form, FormStateInterfa
     '#type' => 'select',
     '#title' => t('Component library'),
     '#options' => [
+      'neutral' => t('Neutral'),
       'ec' => t('European Commission'),
       'eu' => t('European Union'),
     ],
-    '#default_value' => theme_get_setting('component_library'),
+    '#default_value' => theme_get_setting('component_library') ?? 'eu',
```
drishu:
```suggestion
    '#default_value' => theme_get_setting('component_library') ?? 'neutral',
```
Let's default to neutral, this way we bring to the attention of the site builders that they have to chose a flavour

modules/oe_whitelabel_helper/src/Plugin/Block/NeutralFooterBlock.php
```diff
@@ -0,0 +1,44 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build() {
+    $cache = new CacheableMetadata();
+    $cache->addCacheContexts(['languages:language_interface']);
+
+    $config = $this->configFactory->get('oe_corporate_blocks.ec_data.footer');
```
drishu: do we know what data (links) need to be loaded for the neutral block ? I hear neutral is for agencies so EC looks good at first glance

drishu (changes requested):

modules/oe_whitelabel_helper/src/Plugin/Block/NeutralFooterBlock.php
escuriola: Not yet, but yes, EC is really similar to neutral in several things

modules/oe_whitelabel_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -44,7 +45,6 @@ public function getFilters(): array {
   public function getFunctions(): array {
     return [
       new TwigFunction('bcl_footer_links', [$this, 'bclFooterLinks'], ['needs_context' => TRUE]),
-      new TwigFunction('bcl_block', [$this, 'bclBlock']),
```
drishu: if we remove it we have to remove also the bclBlock method, but don't do that yet, just mark it as [deprecated](https://www.drupal.org/docs/develop/standards/api-documentation-and-comment-standards#deprecated)

drishu (changes requested):
The reason drone is failing in lowest is d9 is installed instead of d8, try a rebase

templates/overrides/navigation/oe-corporate-blocks-eu-footer.html.twig
```diff
@@ -1,110 +1,88 @@
...
-        'path': url('<front>')|render
+  {% set _bottom_sections =  {
+    'extra_classes': 'pb-4 pb-lg-5 mt-4 mt-lg-5 bcl-footer__bordered-row row',
+    'cols': [
+      {
+        'classes': 'col-12 col-lg-4 pb-4',
+        'items': [
+        {
+          'type': 'content',
+          'content': drupal_block('whitelabel_eu_logo_block'),
```
drishu: you are deleting the logos in the header refactor so you need to add the logo directly

drishu (changes requested):

PR #73, opened 2022-02-17
OEL-1133: Add Contact forms theming

Maxfire:
Theming for contact form and testing done.


modules/oe_whitelabel_helper/oe_whitelabel_helper.module
```diff
@@ -7,9 +7,106 @@
...
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\contact\MessageInterface;
+
 /**
  * Implements hook_locale_translation_projects_alter().
  */
 function oe_whitelabel_helper_locale_translation_projects_alter(&$projects) {
   $projects['oe_whitelabel_helper']['info']['interface translation server pattern'] = drupal_get_path('module', 'oe_whitelabel_helper') . '/translations/%project-%language.po';
 }
+
```
drishu: the helper module is for general purpose code like field formatters, helper functions so on, this should go into ints own submodule oe_whitelabel_contact_forms, enable the new module in the runner.yml and provide a second PR that enables the new module in oe_showcase.info.yml and replace this line https://github.com/openeuropa/oe_showcase/blob/EPIC-1136-contact_solution/modules/oe_showcase_contact_forms/oe_showcase_contact_forms.info.yml#L8 with the new module.
The new module will require (have dependencies on) oe_contact_forms of course and describe in the info file that openeuropa/oe_contact_forms needs to be added in composer.

composer.json
```diff
@@ -28,7 +28,10 @@
         "phpspec/prophecy-phpunit": "^1 || ^2",
         "easyrdf/easyrdf": "1.0.0 as 0.9.1",
         "drupal/search_api": "^1.21",
-        "drupal/search_api_autocomplete": "^1.5"
+        "drupal/search_api_autocomplete": "^1.5",
+        "openeuropa/oe_contact_forms": "~1.1",
+        "openeuropa/rdf_skos": "~1.0",
```
drishu: remove this, we already have it here https://github.com/openeuropa/oe_contact_forms/blob/master/composer.json#L14

composer.json
```diff
@@ -28,7 +28,10 @@
...
+        "openeuropa/oe_corporate_countries": "~2.0"
```
drishu: remove this, we already have it here https://github.com/openeuropa/oe_contact_forms/blob/master/composer.json#L15

modules/oe_whitelabel_helper/oe_whitelabel_helper.module
```diff
@@ -7,9 +7,106 @@
...
+/**
+ * Implements hook_form_FORM_ID_alter().
+ */
+function oe_whitelabel_helper_form_contact_message_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  // Add 'hr' line on top of the checkboxes.
+  $form['line'] = [
+    '#type' => 'markup',
+    '#markup' => '<hr class="mt-5 mb-5"/>',
+    '#weight' => 6,
```
drishu: this is not good, becasue if I add 6 fields through the UI then it wont appear in the right place, rather add a wrapping element, like a fieldset or whatever and add the hr to it, then move the elements 'copy', 'privacy_policy', 'actions' into it, and set the wrapper weight to a higher number so it is always on the bottom

modules/oe_whitelabel_helper/oe_whitelabel_helper.module
```diff
@@ -7,9 +7,106 @@
...
+      'name',
+      'mail',
+      'subject',
+      'message',
+      'oe_topic',
+    ];
+
+    // Add new fields if exists.
+    foreach ($contact_message->getFields() as $contact_form_field) {
+      if (strpos($contact_form_field->getName(), 'field_') === 0) {
```
drishu: this way is unreliable, the correct way of checking if a field is a managed field and not a base field is by checking the definition
```suggestion
      if (!$contact_form_field->getFieldDefinition() instanceof FieldConfigInterface){
        continue;
      }
      $fields[] = $contact_form_field->getName();
```

drishu (changes requested):

modules/oe_whitelabel_contact_forms/README.md
```diff
@@ -0,0 +1,11 @@
+# OpenEuropa Whitelabel contact forms
+
+This module offers the ability to add contact forms.
+
+## Requirements
+
+To enable this you will have to provide dependent modules in your projects composer.json
+
+```
```
drishu:
```suggestion
```json
```

modules/oe_whitelabel_contact_forms/oe_whitelabel_contact_forms.module
```diff
@@ -0,0 +1,113 @@
...
+      '#orientation' => 'horizontal',
+      '#fields' => [
+        'items' => $items,
+      ],
+    ];
+  }
+}
+
+/**
+ * Implements hook_form_FORM_ID_alter().
```
drishu:
```suggestion
 * Implements hook_form_FORM_ID_alter() for contact_message_form().
```

modules/oe_whitelabel_contact_forms/tests/src/Kernel/ContactFormRenderTest.php
```diff
@@ -0,0 +1,130 @@
...
+    $this->installEntitySchema('path_alias');
+    $this->installEntitySchema('contact_message');
+    $this->installEntitySchema('user');
+    $this->installSchema('system', 'sequences');
+
+    \Drupal::configFactory()
+      ->getEditable('system.theme')
+      ->set('default', 'oe_whitelabel')
+      ->save();
+    $this->container->set('theme.registry', NULL);
```
drishu: no need for this

modules/oe_whitelabel_contact_forms/tests/src/Kernel/ContactFormRenderTest.php
```diff
@@ -0,0 +1,130 @@
...
+    $entity_type_manager = $this->container->get('entity_type.manager');
+    $form = $this->container->get('entity.form_builder')->getForm($message, 'corporate_default');
+    $crawler = new Crawler($this->render($form));
+
+    // Assert classes and contact form.
+    $actual = $crawler->filter('hr.mt-5.mb-5');
+    $this->assertCount(1, $actual);
+    $actual = $crawler->filter('div.mb-2');
+    $this->assertCount(6, $actual);
+    $actual = $crawler->filter('form.contact-message-oe-contact-form-corporate-default-form');
```
drishu: this line should get the form, and we work with that after, so in fact this should be the first assert
```suggestion
    $form = $crawler->filter('form.contact-form');
```

modules/oe_whitelabel_contact_forms/tests/src/Kernel/ContactFormRenderTest.php
```diff
@@ -0,0 +1,130 @@
...
+      'oe_topic' => 'Topic name',
+    ]);
+
+    /** @var \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager */
+    $entity_type_manager = $this->container->get('entity_type.manager');
+    $form = $this->container->get('entity.form_builder')->getForm($message, 'corporate_default');
+    $crawler = new Crawler($this->render($form));
+
+    // Assert classes and contact form.
+    $actual = $crawler->filter('hr.mt-5.mb-5');
```
drishu, line 97: after we get the form as mentioned below we assert only the elements we alter, but in the context of the form
```suggestion
    $wrapper = $form->filter('fieldset');
    $hr = $wrapper->filter('hr.mt-5.mb-5');
```

modules/oe_whitelabel_contact_forms/tests/src/Kernel/ContactFormRenderTest.php
```diff
@@ -0,0 +1,130 @@
...
+    $full_view = $entity_type_manager->getViewBuilder('contact_message')->view($message, 'full');
+    $messenger->addMessage($full_view);
+    $messages = $messenger->messagesByType('status');
+    $html = $this->render($messages);
+    $crawler = new Crawler($html);
+
+    // Assert message success and values.
+    $actual = $crawler->filter('div.alert-success');
+    $this->assertCount(1, $actual);
+    $actual = $crawler->filter('div.oe-contact-form__name .field__label');
```
drishu: rather assert that we have a description list, but just the general structure
```
<div>
  <dt>field label</dt>
</div>
<dd>field value</dd>
```

drishu (changes requested):

modules/oe_whitelabel_contact_forms/tests/src/Kernel/ContactFormRenderTest.php
Maxfire: I appended it into the fieldset prefix so it will fail in this way:
image

modules/oe_whitelabel_contact_forms/oe_whitelabel_contact_forms.module
```diff
@@ -0,0 +1,113 @@
...
+
+    if (in_array('oe_country_residence', $optional_fields)) {
+      $fields[] = 'oe_country_residence';
+    }
+
+    if (in_array('oe_telephone', $optional_fields)) {
+      $fields[] = 'oe_telephone';
+    }
+
+    // Build the label body pairs for the field_list pattern.
```
drishu:
```suggestion
    // Build the label body pairs for the description_list pattern.
```

modules/oe_whitelabel_contact_forms/tests/src/Kernel/ContactFormRenderTest.php
```diff
@@ -0,0 +1,131 @@
...
+    $this->assertStringContainsString("<dd>test@example.com</dd>", $html);
+    $this->assertStringContainsString("<dt>Subject</dt>", $html);
+    $this->assertStringContainsString("<dd>subject</dd>", $html);
+    $this->assertStringContainsString("<dt>Message</dt>", $html);
+    $this->assertStringContainsString("<dd>welcome_message</dd>", $html);
+    $this->assertStringContainsString("<dt>Phone</dt>", $html);
+    $this->assertStringContainsString("<dd>0123456</dd>", $html);
+    $this->assertStringContainsString("<dt>Topic</dt>", $html);
+    $this->assertStringContainsString("<dd>Topic name</dd>", $html);
+
```
drishu: remove this newline

modules/oe_whitelabel_contact_forms/tests/src/Kernel/ContactFormRenderTest.php
```diff
@@ -0,0 +1,131 @@
...
+    $messenger->addMessage($full_view);
+    $messages = $messenger->messagesByType('status');
+    $html = $this->render($messages);
+    $crawler = new Crawler($html);
+
+    // Assert message success and values.
+    $actual = $crawler->filter('div.alert-success');
+    $this->assertCount(1, $actual);
+    $html = $crawler->html();
+    $this->assertStringContainsString("<dt>The sender's name</dt>", $html);
```
drishu:
```suggestion
    $dt = $html->filter('dt');
    $this->assertEquals("The sender's name", $dt->eq(0)->text());
```
Like here https://github.com/openeuropa/oe_bootstrap_theme/blob/8b535fa36e5663a394df610f7e5d87962a2f29fe/tests/src/Kernel/DropbuttonTest.php#L63

drishu (changes requested):

drishu (approved):

abel-santos-corral (approved):

PR #83, opened 2022-02-24
OEL-1191: Fix News meta and spacing

GilNovacomm:


modules/oe_whitelabel_news/templates/node--oe-news--full.html.twig
```diff
@@ -4,17 +4,24 @@
...
 <article{{attributes}}>
   {{ pattern('content_banner', {
     background: 'gray',
     title: label,
     content: content.oe_summary,
     image: image,
-    meta: [
-      content.oe_publication_date,
-    ]
+    meta: _meta,
```
drishu, line 22: there is the same problem here as we had for events
Screenshot 2022-02-28 at 17 31 28
so lets add the fix, also mention somewhere here in a comment that this is to be removed once OEL-1227 is implemented, same for the events one
```suggestion
    meta: _meta,
    attributes: create_attribute().addClass(['ps-0']),
```
I also see the content is not in align with the title, not sure what these changes fix..

modules/oe_whitelabel_news/templates/node--oe-news--full.html.twig
```diff
@@ -4,17 +4,24 @@
...
   }) }}
-  
+
   <div class="container mt-md-4-75 mt-4">
```
drishu: as with the events, this wrapper is actually already present, so we don't need the container, just the margins (see screenshot)
```suggestion
  <div class="mt-md-4-75 mt-4">
```

drishu (changes requested):

modules/oe_whitelabel_news/templates/node--oe-news--full.html.twig
```diff
@@ -4,17 +4,24 @@
  * News full display.
  */
 #}
+
+{% if content.oe_publication_date|render is not empty %}
```
drishu:
```suggestion
{% if not content.oe_publication_date.isEmpty() %}
```

drishu (changes requested):

drishu (dismissed):

Maxfire (approved):

drishu (approved):

PR #84, opened 2022-02-24
OEL-1213: New inline address formatter.

Maxfire:
Address new formatter created to inline the field, tests also done.

### DIFF
--- a/modules/oe_whitelabel_helper/src/Plugin/Field/FieldFormatter/AddressInlineFormatter.php
+++ b/modules/oe_whitelabel_helper/src/Plugin/Field/FieldFormatter/AddressInlineFormatter.php
@@ -21,6 +21,8 @@ use Drupal\Core\Language\LanguageInterface;
  *     "address",
  *   },
  * )
+ *
+ * @see https://github.com/openeuropa/oe_theme/blob/3.x/modules/oe_theme_helper/src/Plugin/Field/FieldFormatter/AddressInlineFormatter.php
  */
 class AddressInlineFormatter extends AddressDefaultFormatter {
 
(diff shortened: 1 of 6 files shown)


modules/oe_whitelabel_helper/src/Plugin/Field/FieldFormatter/AddressInlineFormatter.php
```diff
@@ -0,0 +1,155 @@
...
+use CommerceGuys\Addressing\Locale;
+use Drupal\address\AddressInterface;
+use Drupal\address\Plugin\Field\FieldFormatter\AddressDefaultFormatter;
+use Drupal\Core\Field\FieldItemListInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Language\LanguageInterface;
+
+/**
+ * Format an address inline with locale format and a configurable separator.
+ *
```
drishu, line 16: add a @see so we know where we copied from

modules/oe_whitelabel_helper/tests/src/Kernel/Plugin/Field/FieldFormatter/AddressInlineFormatterTest.php
```diff
@@ -0,0 +1,123 @@
...
+
+namespace Drupal\Tests\oe_whitelabel_helper\Kernel\Plugin\Field\FieldFormatter;
+
+use Drupal\entity_test\Entity\EntityTest;
+use Drupal\Tests\address\Kernel\Formatter\FormatterTestBase;
+
+/**
+ * Test AddressInlineFormatter plugin.
+ *
+ * @group batch2
```
drishu: the batch2 refers to this https://github.com/openeuropa/oe_theme/blob/01b6a8c25ca808bc6b8b7fe968c2452a8706732c/.drone.yml#L154 which we don't have and don't need, please remove

drishu (changes requested):

drishu (approved):

GilNovacomm (approved):

PR #85, opened 2022-02-28
OEL-1190: Avoid spacing after title when no publication date.

escuriola:


modules/oe_whitelabel_news/templates/node--oe-news--full.html.twig
```diff
@@ -4,17 +4,16 @@
  * News full display.
  */
 #}
+{% set _meta = content.oe_publication_date|field_value %}
```
drishu, line 7: not good, you need set the value only if the field is not empty, but I saw we already handle this in a different PR (related issue) here https://github.com/openeuropa/oe_whitelabel/pull/83/files#diff-b281335256f12b69ec30324b5df363050850c5e404adf897ddf3c7ab5c0c64f2R8 so we can close this PR

PR #89, opened 2022-03-08
OEL-486: Improve test coverage

drishu:


escuriola (dismissed):

escuriola (dismissed):

modules/oe_whitelabel_helper/tests/src/Kernel/Plugin/better_exposed_filters/sort/FloatEndSortWidgetTest.php
```diff
@@ -0,0 +1,63 @@
...
+      ],
+    ]);
+
+    // Render the exposed form.
+    $this->renderExposedForm($view);
+    $crawler = new Crawler($this->content->__toString());
+
+    $widget = $crawler->filter('form.bef-exposed-form');
+    $this->assertCount(1, $widget);
+    $this->assertStringContainsString('float-lg-end d-none d-md-flex align-items-baseline', $widget->attr('class'));
```
donquixote, line 60: Is this the only way? Do we want to test the order of classes?

drishu: you'r right this could be made agnostic of order, but since its our own code that controls the order this is the simplest way

modules/oe_whitelabel_helper/tests/src/Kernel/Plugin/better_exposed_filters/sort/FloatEndSortWidgetTest.php
```diff
@@ -0,0 +1,63 @@
...
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_whitelabel_helper\Kernel\Plugin\better_exposed_filters\sort;
+
+use Drupal\Tests\better_exposed_filters\Kernel\BetterExposedFiltersKernelTestBase;
+use Drupal\views\Views;
+use Symfony\Component\DomCrawler\Crawler;
+
+/**
+ * Tests the FloatEndSort exposed form sort widget rendering.
```
donquixote: Let's use the correct machine name or a human name. FloatEndSort is neither :)
For my taste, we could use a human name and then put a `@covers` doc below for the full machine name.
Also, I am not a friend of the "rendering" at the end of the phrase.

```suggestion
 * Tests theming of the 'Float End Sort' widget for "better exposed filters".
 *
 * @see \Drupal\oe_whitelabel_helper\Plugin\better_exposed_filters\sort\FloatEndSortWidget
```
(not saying this is perfect, feel free to improve)

donquixote: Actually `@covers` would be wrong because we are not really testing the functionality, just the theming in whitelabel. So `@see` is more suitable.

donquixote: Or no, `@covers` actually works.
All we do in this class is add classes to the parent widget.
I don't really know if this is a good idea, but this is not the test's fault. So ok for me.

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
```diff
@@ -84,7 +82,7 @@ protected function setUp(): void {
   }
 
   /**
-   * Tests the rendering of blocks.
+   * Tests the rendering of the whitelabel search block.
    */
   public function testBlockRendering(): void {
     $entity_type_manager = $this->container
```
donquixote: The variable name should be `$block_entity_storage`, not `$entity_type_manager`.
(Up to you if you want to fix it in this PR, because it was not introduced here)

modules/oe_whitelabel_helper/tests/src/Kernel/Plugin/better_exposed_filters/sort/FloatEndSortWidgetTest.php
```diff
@@ -0,0 +1,63 @@
...
+
+    // Render the exposed form.
+    $this->renderExposedForm($view);
+    $crawler = new Crawler($this->content->__toString());
+
+    $widget = $crawler->filter('form.bef-exposed-form');
+    $this->assertCount(1, $widget);
+    $this->assertStringContainsString('float-lg-end d-none d-md-flex align-items-baseline', $widget->attr('class'));
+
+    $view->destroy();
```
donquixote, line 62: Not sure if needed here but ok.

tests/src/Kernel/AuthenticationBlockTest.php
```diff
@@ -32,17 +32,15 @@ class AuthenticationBlockTest extends KernelTestBase {
...
       ->set('default', 'oe_whitelabel')
       ->save();
   }
 
   /**
-   * Tests the rendering of blocks.
+   * Tests the rendering of the authentication block.
    */
   public function testBlockRendering(): void {
     $entity_type_manager = $this->container
```
donquixote: Again this should be `$block_storage` or `$block_entity_storage` - up to you if you want to fix.

tests/src/Kernel/FacetsFormTest.php
```diff
@@ -0,0 +1,117 @@
...
+    $this->config('system.theme')
+      ->set('default', 'oe_whitelabel')
+      ->save();
+  }
+
+  /**
+   * Tests the rendering of the Facets Form block.
+   */
+  public function testBlockRendering(): void {
+    $entity_type_manager = $this->container
```
donquixote: `$block_storage` or `$block_entity_storage`.

tests/src/Kernel/FacetsFormTest.php
```diff
@@ -0,0 +1,117 @@
...
+    $this->assertCount(1, $body);
+    $form = $body->filter('form.facets-form');
+    $this->assertCount(1, $form);
+    $button = $crawler->filter('button.btn-light.btn-lg');
+    $this->assertSame('button', $button->attr('type'));
+    $this->assertSame('#bcl-offcanvas', $button->attr('data-bs-target'));
+    $this->assertSame('offcanvas', $button->attr('data-bs-toggle'));
+    $this->assertStringContainsString('Facets form', $button->text());
+    $icon = $button->filter('svg');
+    $this->assertStringContainsString('/assets/icons/bcl-default-icons.svg#filter', $icon->html());
```
donquixote, line 117: Some blank lines for visual separation?

I am not sure about the assertions and selectors, but I can't think of any better. Let's keep them.

tests/src/Kernel/FacetsFormTest.php
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  /**
+   * Tests the rendering of the Facets Form block.
+   */
+  public function testBlockRendering(): void {
+    $entity_type_manager = $this->container
+      ->get('entity_type.manager')
+      ->getStorage('block');
+    $entity = $entity_type_manager->create([
```
donquixote: This could be `$block` instead of `$entity`?

modules/oe_whitelabel_helper/tests/src/Kernel/Plugin/better_exposed_filters/sort/FloatEndSortWidgetTest.php
drishu: It's being used in core kernel tests wherever a view is being created, seems that some references need to be manually unset for proper garbage collection

tests/src/Kernel/FacetsFormTest.php
drishu: I also dont like some selectors in one or two tests, but I think the ones in this test are fine, I grab parents like header or body or the offcanvas outer one and then assert the presence of some elements

modules/oe_whitelabel_helper/tests/src/Kernel/Plugin/better_exposed_filters/sort/FloatEndSortWidgetTest.php
drishu: we use coversDefaultClass see https://github.com/openeuropa/europa-search-client/commit/fcf1f46811ea031fd5b6d33d2d55b17289262024

donquixote: From what I understand this is only useful in combination with `@covers` annotations on methods.
See https://phpunit.readthedocs.io/en/9.5/annotations.html#coversdefaultclass
Otherwise we could simply use `@covers` directly.
But no strong opinion.

donquixote (dismissed):

donquixote (dismissed):

escuriola (dismissed):

escuriola (approved):

donquixote (approved):

PR #91, opened 2022-03-10
OEL-1307: Update oe_whitelabel to BCL 0.20.0

escuriola:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1307


drishu (changes requested):
could not go very deep with the review, for some reason the oe_bootstrap_theme/assets folder is empty for me, can you check if it installes ok for you ?

Also pls update the bcl version in this themes package.json

templates/patterns/modal/modal-language.html.twig
```diff
@@ -1,89 +1 @@
...
-  header: _header,
-  body: _body,
-  footer: _footer,
-} %}
-
-<div class="bcl-language">
-  {% include '@oe-bcl/bcl-modal/modal.html.twig' with _data only %}
-</div>
-{% endspaceless %}
+{% include '@oe-bcl/bcl-language-list/language-list.html.twig' with language_list only %}
```
drishu, line 1: this is very nice, but the close buttons are gone, and there is no modal title

drishu (changes requested):
One more thing, this probably comes from the 0.20 BCL itself, but the ec logos are gone, except for one, see:
Screenshot 2022-03-17 at 22 21 49
Screenshot 2022-03-17 at 22 22 05
Lets ask about this from FRONT and adapt

templates/patterns/modal/modal-language.html.twig
escuriola: image

@drishu  Title and close button are there. I'm checking the header composition, not the language list composition.

oe_whitelabel.theme
```diff
@@ -31,10 +31,11 @@ function oe_whitelabel_preprocess_links__language_block(&$variables) {
     $label = $link['link']['#title'];
 
     $link = [
-      'href' => $href,
+      'path' => $href,
       'name' => $label,
       'id' => 'link_' . $language_code,
-      'hreflang' => $language_code,
+      'label' => $label,
+      'lang' => $language_code,
```
drishu: if we are using 'lang' now then we should use the internal ids https://github.com/openeuropa/oe_theme/blob/3.x/oe_theme.theme#L656

drishu (changes requested):

drishu (dismissed):

brummbar (dismissed):

drishu (dismissed):

drishu (approved):

abel-santos-corral (approved):

brummbar (approved):

PR #94, opened 2022-03-14
OEL-1167: Theme the content language switcher

GilNovacomm:


modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
```diff
@@ -0,0 +1,123 @@
...
+ */
+function oe_whitelabel_multilingual_preprocess_links__oe_multilingual_content_language_block(array &$variables): void {
+  // Generate the label for the unavailable language.
+  /** @var \Drupal\Core\Language\LanguageInterface[] $languages */
+  $languages = \Drupal::service('language_manager')->getNativeLanguages();
+  $currentLanguage = \Drupal::languageManager()->getCurrentLanguage();
+  $current_language_id = $currentLanguage->getId();
+  $variables['unavailable_language'] = $languages[$current_language_id]->getName();
+
+  // Normalize the links to an array of optional languages suitable for the ECL.
```
drishu:
```suggestion
  // Normalize the links to an array of optional languages suitable for the BCL.
```
no copy pasta prego

modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
```diff
@@ -0,0 +1,123 @@
...
+      'path' => "/icons.svg",
+      'size' => "xs",
+      '#attributes' => [
+        'class' => ["me-2"],
+      ],
+    ],
+  ];
+
+  // Generate required ids.
+  $variables['expandable_id'] = Html::getUniqueId('ecl-expandable');
```
drishu: where is this used ?

modules/oe_whitelabel_multilingual/templates/links--oe-multilingual-content-language-block.html.twig
```diff
@@ -0,0 +1,57 @@
+{% spaceless %}
```
drishu: we are in drupal 9 now we should do apply

modules/oe_whitelabel_multilingual/templates/links--oe-multilingual-content-language-block.html.twig
```diff
@@ -0,0 +1,57 @@
...
+
+  {# Parameters:
+  - id (string) (default: collapse-{random number 0-1000}) - collapse custom id - optional
+  - message (alert object) (default: {})
+  - expandable (button object) (default: [])
+  - language (array of link object) (default: [])
+  #}
+
+  {% set id = "dropdown-languages" %}
```
drishu: you are setting the id, so why do you write that it's a parameter ?

modules/oe_whitelabel_multilingual/templates/links--oe-multilingual-content-language-block.html.twig
```diff
@@ -0,0 +1,57 @@
...
+    icon: {
+      name: "caret-down-fill",
+      path: "/icons.svg",
+    },
+    outline: "true",
+  } %}
+
+  {% set _id = id|default('collapse-' ~ random(1000)) %}
+  {% set _message = message|default({}) %}
+  {% set _expandable = expandable|default({}) %}
```
drishu: if you set these values then this doesn't make sense

modules/oe_whitelabel_multilingual/templates/links--oe-multilingual-content-language-block.html.twig
```diff
@@ -0,0 +1,57 @@
...
+      >
+        {% for language in languages %}
+          {% include '@oe-bcl/bcl-link/link.html.twig' with language|merge({
+            attributes: create_attribute().addClass(["btn", "btn-link", "d-block", "ps-0", "text-start"])
+          }) only %}
+        {% endfor %}
+      </div>
+    </div>
+  </div>
+{% endspaceless %}
```
drishu: please move this template to its proper place in /templates/overrides/navigation because like this it creates the confusion that if you disable this module the template wont be picked up by drupal which is not the case.

modules/oe_whitelabel_multilingual/templates/links--oe-multilingual-content-language-block.html.twig
```diff
@@ -0,0 +1,57 @@
+{% spaceless %}
+
+  {# Parameters:
```
drishu: instead of this, leave a comment saying where you copied it from, what you changed/differs, and a todo, with what should be replaced when.

modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
```diff
@@ -0,0 +1,123 @@
...
+  $multilingual_helper = \Drupal::service('oe_multilingual.helper');
+  $entity = $multilingual_helper->getEntityFromCurrentRoute();
+  /** @var \Drupal\Core\Entity\EntityInterface $translation */
+  $translation = $multilingual_helper->getCurrentLanguageEntityTranslation($entity);
+  $variables['languages'][] = [
+    'path' => $translation->toUrl()->setAbsolute(TRUE)->toString(),
+    'hreflang' => $translation->language()->getId(),
+    'label' => $languages[$translation->language()->getId()]->getName(),
+    'current' => TRUE,
+    'icon_position' => "before",
```
drishu: let's not mix single with double quotes where not needed, lets keep to single

modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
```diff
@@ -0,0 +1,123 @@
...
+    'remove_icon_spacers' => TRUE,
+    'icon' => [
+      'name' => "check-lg",
+      'path' => "/icons.svg",
```
drishu: bcl_icon_path is where we store the icon path

modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
```diff
@@ -0,0 +1,123 @@
...
+      'size' => "xs",
+      '#attributes' => [
+        'class' => ["me-2"],
```
drishu: this is not applied, the class is missing, please check why

modules/oe_whitelabel_multilingual/templates/links--oe-multilingual-content-language-block.html.twig
```diff
@@ -0,0 +1,57 @@
...
+  - id (string) (default: collapse-{random number 0-1000}) - collapse custom id - optional
+  - message (alert object) (default: {})
+  - expandable (button object) (default: [])
+  - language (array of link object) (default: [])
+  #}
+
+  {% set id = "dropdown-languages" %}
+  {% set message = {
+    message: "This page is not available in %s."|format(unavailable_language)|t,
+    icon_path: "/icons.svg",
```
drishu: wrong icon path

modules/oe_whitelabel_multilingual/templates/links--oe-multilingual-content-language-block.html.twig
```diff
@@ -0,0 +1,57 @@
...
+    variant: "warning",
+  }%}
+  {% set expandable = {
+    label: "Choose another language"|t,
+    icon: {
+      name: "caret-down-fill",
+      path: "/icons.svg",
```
drishu: wrong icon path.

Where all these icons showing for you ?

Please check that the message, the button and the active item have icons visible

modules/oe_whitelabel_multilingual/templates/links--oe-multilingual-content-language-block.html.twig
```diff
@@ -0,0 +1,57 @@
...
+  {# Parameters:
+  - id (string) (default: collapse-{random number 0-1000}) - collapse custom id - optional
+  - message (alert object) (default: {})
+  - expandable (button object) (default: [])
+  - language (array of link object) (default: [])
+  #}
+
+  {% set id = "dropdown-languages" %}
+  {% set message = {
+    message: "This page is not available in %s."|format(unavailable_language)|t,
```
drishu: you dont need to call format when the translate function does all you need `|t({'@language...`

tests/src/Kernel/ContentLanguageSwitcherTest.php
```diff
@@ -0,0 +1,172 @@
...
+    $actual = $crawler->filter('#dropdown-languages');
+    $this->assertCount(1, $actual);
+
+    // Warning message doesn't contain the unavailable language, the translation
+    // will have it.
+    $this->assertUnavailableLanguage($crawler, 'This page is not available in български.');
+
+    // Make sure that available languages are properly rendered.
+    $this->assertTranslationLinks($crawler, ['español', 'English']);
+  }
```
drishu, line 159: why not continue with https://github.com/openeuropa/oe_theme/blob/a33ffc0b726fd3be5985c6c8ccd7033350c06e44/tests/src/Kernel/ContentLanguageSwitcherTest.php#L84 ?

tests/src/Kernel/ContentLanguageSwitcherTest.php
```diff
@@ -0,0 +1,172 @@
...
+    'block',
+    'content_translation',
+    'language',
+    'locale',
+    'node',
+    'oe_bootstrap_theme_helper',
+    'oe_multilingual',
+    'oe_whitelabel_helper',
+    'oe_whitelabel_multilingual',
+    'system',
```
drishu, line 31: can you double check this test is running in drone, like make it fail and see it runs. Asking because like this it was failing in my local, had to add 'oe_corporate_blocks'

tests/src/Kernel/ContentLanguageSwitcherTest.php
```diff
@@ -0,0 +1,172 @@
...
+      'oe_multilingual',
+      'oe_whitelabel_helper',
+      'system',
+    ]);
+    $this->container->get('module_handler')->loadInclude('oe_multilingual', 'install');
+    oe_multilingual_install(FALSE);
+
+    $this->container->get('theme_installer')->install(['oe_whitelabel']);
+    $this->config('system.theme')->set('default', 'oe_whitelabel')->save();
+    $this->container->set('theme.registry', NULL);
```
drishu: you dont need to empty the registry

drishu (changes requested):
- also place the block so it appears above the content;
- replace the enabling of oe_multilingual in the runner with this module;
- and update the branch of course

tests/src/Kernel/ContentLanguageSwitcherTest.php
GilNovacomm: It did pass on drone, but locally after rebasing it I do need the same dependency.

modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
```diff
@@ -0,0 +1,116 @@
...
+  $translation = $multilingual_helper->getCurrentLanguageEntityTranslation($entity);
+  $variables['languages'][] = [
+    'path' => $translation->toUrl()->setAbsolute(TRUE)->toString(),
+    'hreflang' => $translation->language()->getId(),
+    'label' => $languages[$translation->language()->getId()]->getName(),
+    'current' => TRUE,
+    'icon_position' => 'before',
+    'remove_icon_spacers' => FALSE,
+    'icon' => [
+      'name' => "check-lg",
```
drishu: single quotes please

modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
```diff
@@ -0,0 +1,116 @@
...
+      'path' => $variables['bcl_icon_path'],
+      'size' => "xs",
```
drishu: same

templates/overrides/navigation/links--oe-multilingual-content-language-block.html.twig
```diff
@@ -0,0 +1,52 @@
+{% apply spaceless %}
+  {# This template uses the BCL composition for
+  language switcher with following components:
```
drishu:
```suggestion
  language switcher. _id, _message and _expandable are overwritten.
  @todo Call include once BCL 0.20 is available.
```
or something of sorts. You could even add a @see and a path to the file in git

templates/overrides/navigation/links--oe-multilingual-content-language-block.html.twig
```diff
@@ -0,0 +1,52 @@
...
+  - bcl-alert
+  - bcl-button
+  - bcl-link
+  #}
+  {% set _id = "dropdown-languages" %}
+  {% set _message = {
+    message: "This page is not available in @language."|t({'@language': unavailable_language}),
+    icon_path: "/icons.svg",
```
drishu: bcl_icon_path

tests/src/Kernel/ContentLanguageSwitcherTest.php
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+ * Test content language switcher rendering.
+ */
+class ContentLanguageSwitcherTest extends KernelTestBase {
+  /**
+   * {@inheritdoc}
+   */
+  public static $modules = [
+    'block',
```
drishu, line 21: please sort this list of modules

drishu (changes requested):
Again:

- [x] also place the block so it appears above the content (to make sure of this also place the content block, see oe_theme);
- [x] replace the enabling of oe_multilingual in the runner with this module;

tests/src/Kernel/ContentLanguageSwitcherTest.php
```diff
@@ -0,0 +1,214 @@
...
+    // Verify that the content has been rendered in the fallback language.
+    $this->assertSelectedLanguage($crawler, 'English');
+    
+    // Re-render the block assuming a request to the base language English
+    // version of the node.
+    $this->setCurrentRequest('/en/node/' . $node->id());
+
+    $html = (string) $this->container->get('renderer')->renderRoot($render);
+    $crawler = new Crawler($html);
+    $render = $plugin_block->build();
```
drishu: this is duplicate code from above, please remove

tests/src/Kernel/ContentLanguageSwitcherTest.php
```diff
@@ -0,0 +1,212 @@
...
+
+    // Verify that the content has been rendered in the fallback language.
+    $this->assertSelectedLanguage($crawler, 'English');
+
+    // Simulate a request to the canonical route of the node base language.
+    $this->setCurrentRequest('/node/' . $node->id());
+
+    $html = (string) $this->container->get('renderer')->renderRoot($render);
+    $crawler = new Crawler($html);
+
```
drishu:
```suggestion
```

modules/oe_whitelabel_multilingual/config/optional/block.block.openeuropa_content_language_switcher.yml
```diff
@@ -0,0 +1,19 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - oe_multilingual
+  theme:
+    - oe_whitelabel
+id: openeuropa_content_language_switcher
```
drishu: lets name the file 
block.block.oe_whitelabel_language_switcher.yml
and the id
oe_whitelabel_language_switcher

modules/oe_whitelabel_multilingual/config/optional/block.block.oe_whitelabel_main_page_content.yml
```diff
@@ -0,0 +1,19 @@
+langcode: en
```
drishu: move this file into oe_whitelabel/config/optional with the rest of the block config exports

modules/oe_whitelabel_multilingual/config/optional/block.block.openeuropa_content_language_switcher.yml
```diff
@@ -0,0 +1,19 @@
+langcode: en
```
drishu: move this file into oe_whitelabel/config/optional with the rest of the block config exports

drishu (changes requested):

modules/oe_whitelabel_multilingual/config/optional/block.block.openeuropa_content_language_switcher.yml
GilNovacomm: If we move this file and rename it  as on the suggestion above it will replace the already existing block placed on the nav_right region.
How about renaming it to oe_whitelabel_MULTILINGUAL_language_switcher instead?

tests/src/Kernel/ContentLanguageSwitcherTest.php
GilNovacomm: This re-render is necessary  validate the new URI request to the node base language.

drishu (dismissed):

drishu (dismissed):

drishu (dismissed):

drishu (dismissed):

templates/overrides/navigation/links--oe-multilingual-content-language-block.html.twig
```diff
@@ -0,0 +1,51 @@
...
+    label: "Choose another language"|t,
+    icon: {
+      name: "caret-down-fill",
+      path: bcl_icon_path,
+    },
+    outline: "true",
+  } %}
+
+  {% set _languages = languages|default([]) %}
+  {% set _expandable_attributes = _expandable.attributes ?: create_attribute() %}
```
donquixote, line 23: This seems wrong: `_expandable` is being set above, and `_expandable.attributes` does not exist.

templates/overrides/navigation/links--oe-multilingual-content-language-block.html.twig
```diff
@@ -0,0 +1,51 @@
...
+    variant: "warning",
+  }%}
+  {% set _expandable = {
+    label: "Choose another language"|t,
+    icon: {
+      name: "caret-down-fill",
+      path: bcl_icon_path,
+    },
+    outline: "true",
+  } %}
```
donquixote, line 20: Can we name this `_expand_button` instead of `_expandable`?
Or, if up to me, we could use inline value instead of a variable - see below.

templates/overrides/navigation/links--oe-multilingual-content-language-block.html.twig
```diff
@@ -0,0 +1,51 @@
...
+    .setAttribute('aria-expanded', 'false')
+    .setAttribute('aria-controls', _id)
+    .setAttribute('data-bs-target', '#' ~ _id)
+  %}
+
+  {% include '@oe-bcl/bcl-alert/alert.html.twig' with _message only %}
+  <div class="mb-3">
+    {% include '@oe-bcl/bcl-button/button.html.twig' with _expandable|merge({
+      attributes: _expandable_attributes
+    }) only %}
```
donquixote, line 36: Can we we drop the variable and use an inline value?
(could be just my personal preference)
E.g.
```twig
    {# Dropdown-like button to hide/reveal the language list. #}
    {% include '@oe-bcl/bcl-button/button.html.twig' with {
      label: "Choose another language"|t,
      icon: {
        name: "caret-down-fill",
        path: bcl_icon_path,
      },
      outline: "true",
      attributes: create_attribute()
        .setAttribute('data-bs-toggle', 'collapse')
        .setAttribute('autocomplete', 'off')
        .setAttribute('aria-expanded', 'false')
        .setAttribute('aria-controls', _id)
        .setAttribute('data-bs-target', '#' ~ _id),
    } only %}
```

templates/overrides/navigation/links--oe-multilingual-content-language-block.html.twig
```diff
@@ -0,0 +1,51 @@
+{% apply spaceless %}
+  {# This template uses the BCL composition for the
+  language switcher.
+  _id, _message and _expandable are overwritten.
+  @todo Call include once BCL 0.20 is available.
+  #}
+  {% set _id = "dropdown-languages" %}
```
donquixote: Is this id unique enough?
We could use 'dropdown-content-language'.

donquixote: We could also use a randomized id, but this is bad for caching I think.

templates/overrides/navigation/links--oe-multilingual-content-language-block.html.twig
```diff
@@ -0,0 +1,51 @@
...
+  <div class="mb-3">
+    {% include '@oe-bcl/bcl-button/button.html.twig' with _expandable|merge({
+      attributes: _expandable_attributes
+    }) only %}
+
+    <div class="collapse mt-3" id="{{ _id }}">
+      <div
+        class="d-md-grid"
+        style="grid-auto-flow: column; grid-template-rows: repeat(4, 1fr)"
+      >
```
donquixote, line 42: Sad that we use inline CSS here.
But I see that netlify does the same. ¯\_(ツ)_/¯

templates/overrides/navigation/links--oe-multilingual-content-language-block.html.twig
```diff
@@ -0,0 +1,51 @@
...
+  {% set _expandable = {
+    label: "Choose another language"|t,
+    icon: {
+      name: "caret-down-fill",
+      path: bcl_icon_path,
+    },
+    outline: "true",
+  } %}
+
+  {% set _languages = languages|default([]) %}
```
donquixote, line 22: Is this default value really needed and useful?
I think if languages are empty we want to abort the entire thing.
Also, the preprocess should guarantee that languages variable exists.

templates/overrides/navigation/links--oe-multilingual-content-language-block.html.twig
donquixote: Ok, I see now that lots in here is copied from bcl-language-switcher.html.twig.
If the goal is to be as close as possible to the bcl template, then some of my comments here can be ignored.

templates/overrides/navigation/links--oe-multilingual-content-language-block.html.twig
```diff
@@ -0,0 +1,51 @@
...
+  {% set _expandable_attributes = _expandable.attributes ?: create_attribute() %}
+  {% set _expandable_attributes = _expandable_attributes
+    .setAttribute('data-bs-toggle', 'collapse')
+    .setAttribute('autocomplete', 'off')
+    .setAttribute('aria-expanded', 'false')
+    .setAttribute('aria-controls', _id)
+    .setAttribute('data-bs-target', '#' ~ _id)
+  %}
+
+  {% include '@oe-bcl/bcl-alert/alert.html.twig' with _message only %}
```
donquixote, line 32: Again I would prefer inline parameters instead of a variable.
But I am ok if the goal is to be aligned with the bcl template ([snapshot at 67447742](https://github.com/openeuropa/bootstrap-component-library/blob/6744774215866a144092f9f499901ae0416bccc9/src/compositions/bcl-language-switcher/language-switcher.html.twig))

E.g.
```twig
{% include '@oe-bcl/bcl-alert/alert.html.twig' with {
  message: "This page is not available in @language."|t({'@language': unavailable_language}),
  path: bcl_icon_path,
  variant: "warning",
} only %}
```

modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
```diff
@@ -0,0 +1,114 @@
...
+    'current' => TRUE,
+    'icon_position' => 'before',
+    'remove_icon_spacers' => FALSE,
+    'icon' => [
+      'name' => 'check-lg',
+      'path' => $variables['bcl_icon_path'],
+      'size' => 'xs',
+      '#attributes' => ['class' => ['me-2']],
+    ],
+  ];
```
donquixote: I wonder, why does `ContentLanguageBlock` from `oe_multilingual` not provide this link?

Also, when we show this link, as a visitor I think I would want to see something like: "✓ English (shown instead of Polish)" instead of just "✓ English", if Polish is the language where we don't have a translation.

But it seems this has been decided in BCL and we are not going to change it here.

donquixote:
I have some general feedback about UX for the design in BCL.
This is not really the place to fix this, but I am going to mention it anyway:
- The open/close button looks like a radio element, but it does not behave as such, defying user expectations.
- The open/close icon always points down. It should point up when the list is opened.

Not a blocker for this PR.

oe_whitelabel.info.yml
```diff
@@ -35,15 +35,16 @@ config_devel:
...
     - block.block.oe_whitelabel_local_actions
     - block.block.oe_whitelabel_local_tasks
+    - block.block.oe_whitelabel_main_navigation
+    - block.block.oe_whitelabel_main_page_content
     - block.block.oe_whitelabel_messages
-    - block.block.oe_whitelabel_page_title
-    - block.block.whitelabelsearchblock
-    - cas.settings.yml
+    - block.block.oe_whitelabel_neutral_footer
+    - block.block.oe_whitelabel_search_form
```
donquixote, line 50: I found that these entries are already wrong in 1.x.
Also in two other files:
- oe_whitelabel_starter_event.info.yml (wrong order)
- oe_whitelabel_starter_news.info.yml (".yml" suffix)

Imo we should have a separate branch to fix all of the entries, and then we rebase.
But I can accept if we want to do this here to move forward.

config/optional/block.block.oe_whitelabel_ec_corporate_footer.yml
```diff
@@ -3,6 +3,7 @@ status: true
 dependencies:
   module:
     - oe_corporate_blocks
+    - oe_whitelabel_helper
```
donquixote, line 6: I wonder where this additional dependency comes from.
Probably just noise.
Not a blocker, just curious.

modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
```diff
@@ -0,0 +1,114 @@
...
+}
+
+/**
+ * Implements hook_preprocess_links__language_block().
+ */
+function oe_whitelabel_multilingual_preprocess_links__language_block(&$variables) {
```
donquixote, line 70: TBD: We _could_ follow the pattern where we implement the preprocess hook on behalf of oe_whitelabel theme, instead of the module.
This would prevent the hook from running if oe_whitelabel is not the active theme, e.g. during a test when we only test the structure and not the theme.
I think it won't matter in this case, but we want to establish best practices that are safe to use elsewhere.

donquixote: (Arguments could be made that all the preprocess functions should be in the theme, not in a submodule, to prevent that the template is ever used without the preprocess. Still, both would depend on the exact config coming from a module.)

modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
```diff
@@ -0,0 +1,114 @@
...
+      'hreflang' => $language_code,
+      'label' => $link['link']['#title'],
+      'current' => FALSE,
+    ];
+  }
+
+  // Add the current language to the list.
+  $multilingual_helper = \Drupal::service('oe_multilingual.helper');
+  $entity = $multilingual_helper->getEntityFromCurrentRoute();
+  /** @var \Drupal\Core\Entity\EntityInterface $translation */
```
donquixote: Put a `@var` doc on the `$multilingual_helper` service above, then you don't need one here.

tests/src/Kernel/ContentLanguageSwitcherTest.php
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  /**
+   * Tests the rendering of the language switcher block.
+   */
+  public function testMultilingualLanguageSwitcherBlockRendering(): void {
+    $node = Node::create([
+      'title' => 'Hello, world!',
+      'type' => 'oe_demo_translatable_page',
+    ]);
```
donquixote, line 87: This is very interesting.
The `oe_demo_translatable_page` content type is defined in `oe_multilingual_demo`, but that module is not enabled in this test.
But the test still passes.
In fact it still passes if we insert any arbitrary string for the node type.

tests/src/Kernel/ContentLanguageSwitcherTest.php
```diff
@@ -0,0 +1,215 @@
...
+    $this->setCurrentRequest('/bg/node/' . $node->id());
+
+    // Setup and render language switcher block.
+    $block_manager = \Drupal::service('plugin.manager.block');
+    $config = [
+      'id' => 'oe_multilingual_content_language_switcher',
+      'label' => 'Content language switcher',
+      'provider' => 'oe_multilingual',
+      'label_display' => '0',
+    ];
```
donquixote, line 102: We can remove all array keys except 'label_display'. They will be filled with default values.

donquixote: Ok, perhaps we want all the keys to fully simulate a block with the typical configuration - ok for me.

donquixote (commented):
I left some comments.
Maybe some of these should lead to changes, but some maybe can be ignored, if there are good reasons why we do it this way, or decisions were already made in the past.

templates/overrides/navigation/links--oe-multilingual-content-language-block.html.twig
drishu: yes, we don't have BCL 0.20 yet, but as soon as we do it will call a template from there, so no need to QA on it

templates/overrides/navigation/links--oe-multilingual-content-language-block.html.twig
drishu: no, let's no QA on it

templates/overrides/navigation/links--oe-multilingual-content-language-block.html.twig
drishu: lets no QA BCL stuff that will be replaced, but you can check the template there and start a discussion with FRONT

templates/overrides/navigation/links--oe-multilingual-content-language-block.html.twig
drishu: report to FRONT, maybe even repat(3 is there are only 3 items and so on

templates/overrides/navigation/links--oe-multilingual-content-language-block.html.twig
drishu: its BCL, will be replaced

templates/overrides/navigation/links--oe-multilingual-content-language-block.html.twig
drishu: its BCL, will be replaced

modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
drishu: I think that just indicates the fallback language

modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
drishu: Let's keep it like this for now, and prep a discussion among leads to establish final word on this @donquixote

tests/src/Kernel/ContentLanguageSwitcherTest.php
drishu: indeed, I noticed that myself, but I kept it like that so its as close as possible to the test in oe_theme

templates/overrides/navigation/links--oe-multilingual-content-language-block.html.twig
donquixote: repeat(3) is not needed.
It already behaves that way with 3 items.

tests/src/Kernel/ContentLanguageSwitcherTest.php
donquixote: Let's add an inline comment:
```php
    $node = Node::create([
      'title' => 'Hello, world!',
      // The node type does not actually exist in this test, but it works.
      'type' => 'oe_demo_translatable_page',
    ]);
```

modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
donquixote: I see that, but it is not immediately obvious for a visitor.

donquixote: Anyway I am resolving this.

tests/src/Kernel/ContentLanguageSwitcherTest.php
```diff
@@ -109,7 +109,7 @@ public function testMultilingualLanguageSwitcherBlockRendering(): void {
     $crawler = new Crawler($html);
 
     // Make sure that content language switcher block is present.
-    $actual = $crawler->filter('#dropdown-languages');
+    $actual = $crawler->filter('div.collapse.mt-3');
```
donquixote, line 112: I don't like to see the `.mt-3` here, but I suppose we have to for disambiguation?
Worst case we have to update this test when this class changes.

modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
```diff
@@ -57,6 +59,9 @@ function oe_whitelabel_multilingual_preprocess_links__oe_multilingual_content_la
       ],
     ];
   }
+
+  // Generate required ids.
+  $variables['expandable_id'] = Html::getUniqueId('bcl-expandable');
```
donquixote, line 64: I wonder if caching (render cache) could cause the same id to be used elsewhere on the same page.
But I think this would need a very convoluted scenario, which we should ignore for the time being.
So I am ok with it.

modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
```diff
@@ -39,20 +39,24 @@ function oe_whitelabel_multilingual_preprocess_links__oe_multilingual_content_la
...
+      'icon_position' => 'before',
+      'remove_icon_spacers' => FALSE,
+      'icon' => [
+        'name' => 'check-lg',
+        'path' => $variables['bcl_icon_path'],
+        'size' => 'xs',
+        '#attributes' => ['class' => ['me-2']],
+      ],
+    ];
+  }
```
donquixote, line 61: What is the scenario where this happens?

donquixote: Btw I was considering to suggest early return pattern here (invert the if to get rid of indentation).
But then we would have to drop the early return when we add the `$variables['expandable_id']`.
This is why I prefer to use early return only in functions with a single responsibility - which is often not the case for preprocess.
So ok to keep it this way.

modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
```diff
@@ -39,20 +39,24 @@ function oe_whitelabel_multilingual_preprocess_links__oe_multilingual_content_la
...
-      '#attributes' => ['class' => ['me-2']],
-    ],
-  ];
+  // If we don't have a language id defined yet, the current translation wasn't
+  // saved, so we don't add it to the list.
+  if ($translation->language()->getId() !== 'und') {
+    $variables['languages'][] = [
+      'path' => $translation->toUrl()->setAbsolute(TRUE)->toString(),
+      'hreflang' => $translation->language()->getId(),
+      'label' => $languages[$translation->language()->getId()]->getName(),
```
donquixote, line 50: We could introduce a local var for `$translation->language()->getId()`, to avoid the repeated call, e.g. `$current_content_language_id`.

We could also check whether `$languages[$current_content_language_id]` exists.
Would be strange if it does not. Perhaps if a site language is deleted, renamed or disabled, while translations still exist? I don't think it will happen.

modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
```diff
@@ -39,20 +39,24 @@ function oe_whitelabel_multilingual_preprocess_links__oe_multilingual_content_la
...
-    'icon' => [
-      'name' => 'check-lg',
-      'path' => $variables['bcl_icon_path'],
-      'size' => 'xs',
-      '#attributes' => ['class' => ['me-2']],
-    ],
-  ];
+  // If we don't have a language id defined yet, the current translation wasn't
+  // saved, so we don't add it to the list.
+  if ($translation->language()->getId() !== 'und') {
```
donquixote, line 46: I notice you are using `'und'` string literal instead of a constant.
In Drupal 7 I would have told you to use `LANGUAGE_NONE`.
Now this has changed, see [LANGUAGE_NONE changed to LANGUAGE_NOT_SPECIFIED, LANGUAGE_NOT_APPLICABLE and LANGUAGE_MULTIPLE added
](https://www.drupal.org/node/1473200)
Not sure if we also need to support other possible values.
Perhaps safer to just check `isset($languages[$current_content_language_id])`?

modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
donquixote: Also if language is 'und', will the message "This page is not available in ***." make sense for all cases?

I think we should settle for something that works, we have no time to investigate all possible scenarios.
If you say you want to keep the current version, let's go with it.

drishu (approved):

donquixote (approved):
Good enough, we might fine-tune in follow-up.

PR #97, opened 2022-03-18
[OEL-1255] OEL-1281: Upgrade path from oe_bootstrap_theme_paragraphs

donquixote:
[OEL-1281](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1281)


modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.install
```diff
@@ -39,3 +69,129 @@ function oe_whitelabel_paragraphs_install($is_syncing): void {
...
+ * Normally this should happen through a batch process, e.g. via $sandbox in a
+ * hook_update_N(). Unfortunately, hook_install() does not support batch
+ * processes.
+ * For the known projects where this will be used, we assume that not too many
+ * paragraphs exist yet - fingers crossed.
+ *
+ * See https://atendesigngroup.com/articles/programmatically-copy-field-data-drupal-8.
+ *
+ * @param string[][] $field_names_by_bundle
+ *   Format: $[$bundle][$dest_field_name] = $source_field_name.
```
drishu: Lets use the official way of documenting an associative array https://www.drupal.org/docs/develop/standards/api-documentation-and-comment-standards#themepreprocess
I agree that your way of doing it has certain advantages but find or open an issue on d.org and let's discuss there further

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.install
```diff
@@ -39,3 +69,129 @@ function oe_whitelabel_paragraphs_install($is_syncing): void {
...
+    }
+    $paragraph_revision->setNewRevision(FALSE);
+    $paragraph_revision->save();
+  }
+}
+
+/**
+ * Removes legacy field instances from oe_bootstrap_theme_paragraphs module.
+ *
+ * @param string[][] $field_names_by_bundle
```
drishu, line 180: `string[][]` is good but the description is not the drupal best practice

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.install
```diff
@@ -39,3 +69,129 @@ function oe_whitelabel_paragraphs_install($is_syncing): void {
...
+  return $field_names_by_bundle;
+}
+
+/**
+ * Migrates field data from the old oe_bootstrap_theme_paragraphs module.
+ *
+ * Normally this should happen through a batch process, e.g. via $sandbox in a
+ * hook_update_N(). Unfortunately, hook_install() does not support batch
+ * processes.
+ * For the known projects where this will be used, we assume that not too many
```
drishu: delete this sentence pls, especially the "fingers crossed."

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.install
```diff
@@ -39,3 +69,129 @@ function oe_whitelabel_paragraphs_install($is_syncing): void {
...
+  }
+
+  return $field_names_by_bundle;
+}
+
+/**
+ * Migrates field data from the old oe_bootstrap_theme_paragraphs module.
+ *
+ * Normally this should happen through a batch process, e.g. via $sandbox in a
+ * hook_update_N(). Unfortunately, hook_install() does not support batch
```
drishu, line 131: why not use queue API?

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.install
```diff
@@ -39,3 +69,129 @@ function oe_whitelabel_paragraphs_install($is_syncing): void {
...
+  $paragraph_ids = $query->execute();
+
+  foreach ($paragraph_ids as $revision_id => $paragraph_id) {
+    $paragraph_revision = $paragraphs_storage->loadRevision($revision_id);
+    if (!$paragraph_revision) {
+      // Revision not found - this is unexpected, but survivable.
+      continue;
+    }
+    $modified = FALSE;
+    foreach ($field_names_by_bundle[$paragraph_revision->bundle()] ?? [] as $dest_field_name => $source_field_name) {
```
drishu: this reminds me of python syntax, cant we just assign `$field_names_by_bundle[$paragraph_revision->bundle()]` to a variable and add an if empty before the loop like you did before?

modules/oe_whitelabel_paragraphs/tests/modules/oe_whitelabel_legacy_paragraphs_test/oe_whitelabel_legacy_paragraphs_test.install
```diff
@@ -0,0 +1,40 @@
...
+ * Customise fields for bootstrap theme paragraphs.
+ */
+function oe_whitelabel_legacy_paragraphs_test_install($is_syncing): void {
+  // If we are installing from config, we bail out.
+  if ($is_syncing) {
+    return;
+  }
+
+  $configs = [
+    'core.entity_form_display.paragraph.oe_links_block.default',
```
drishu: sort these, plus you are missing one config if you really want to import all

modules/oe_whitelabel_paragraphs/tests/modules/oe_whitelabel_legacy_paragraphs_test/oe_whitelabel_legacy_paragraphs_test.install
```diff
@@ -0,0 +1,40 @@
...
+    'core.entity_view_display.paragraph.oe_social_media_follow.default',
+    'core.entity_form_display.paragraph.oe_description_list.default',
+    'core.entity_view_display.paragraph.oe_description_list.default',
+    'core.entity_form_display.paragraph.oe_accordion_item.default',
+    'core.entity_view_display.paragraph.oe_accordion_item.default',
+    'core.entity_form_display.paragraph.oe_facts_figures.default',
+    'core.entity_view_display.paragraph.oe_facts_figures.default',
+    'core.entity_form_display.paragraph.oe_list_item_block.default',
+    'core.entity_form_display.paragraph.oe_list_item_block.highlight',
+    'core.entity_form_display.paragraph.oe_list_item.default',
```
drishu: do all these make sense in the context of the migration we are testing ?

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.install
```diff
@@ -39,3 +69,129 @@ function oe_whitelabel_paragraphs_install($is_syncing): void {
 
   ConfigImporter::importMultiple('oe_whitelabel_paragraphs', '/config/overrides/', $configs);
 }
+
+/**
+ * Gets a map of legacy fields to be migrated on install.
+ *
+ * @return string[][]
+ *   Format: $[$bundle][$dest_field_name] = $legacy_field_name.
```
drishu: same comment like before

modules/oe_whitelabel_paragraphs/tests/src/Functional/InstallTest.php
```diff
@@ -0,0 +1,203 @@
...
+      ],
+    ];
+
+    $actual_updated = [];
+    $actual_deleted = [];
+    foreach ($ids as $name => $id) {
+      $updated_paragraph = Paragraph::load($id);
+      self::assertNotNull($updated_paragraph);
+      foreach ($expected_created[$name] as $field_name => $value) {
+        if (!$updated_paragraph->hasField($field_name)) {
```
drishu: this is a bit confusing to me, so I think a good candidate for adding a comment above the if, explaining why we continue there.

modules/oe_whitelabel_paragraphs/tests/src/Functional/InstallTest.php
```diff
@@ -0,0 +1,203 @@
...
+          continue;
+        }
+        $actual_updated[$name][$field_name] = $updated_paragraph->get($field_name)->value;
+      }
+      foreach ($expected_deleted[$name] as $field_name => $deleted) {
```
drishu, line 208: I know this is constructed using $paragraphs_data which we control, but its confusing and prone to human error, a simple listing of checks one after the other would have been simple and clear.
I prefer SOLID and KISS programming principles.
This time we keep it like this , but please read up, on above, especially on "KISS"

drishu (changes requested):

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.install
donquixote: I think the main reason for batch in hook_update_N() is to work around max_execution_time, and also to show a progress bar when doing it through the UI.
For memory concerns, we simply have to avoid keeping all data in memory at the same time, so that garbage collection can do its job.
I think the max_execution_time mostly applies in web UI, for cli it is typically disabled.

In hook_install() we don't have the luxury of splitting up the current process, because the calling code expects everything to finish in one call.
Using queue, afaik, would mean that part of this operation would happen during cron _after_ the install.
So it would depend on people actually running cron, and operations directly after the install could not rely on the operation being complete.

modules/oe_whitelabel_paragraphs/tests/src/Functional/InstallTest.php
donquixote: The purpose of this data collection is to have a more useful and more complete diff output on failure.
But yes, with individual assertions the code would look simpler.

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.install
drishu: I've tested and this brings no benefit when calling from install, ignore pls

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.install
donquixote: Actually this doc section does not apply here.
It refers to associative arrays with known/fixed keys (e.g. in psalm syntax `array{path: string, options: array}`.)
What we have here is a nested array with two levels of dynamic keys, but no fixed keys. (psalm `array<string, array<string, string>>`).
Existing documentation for such types is often poor or very verbose.
I will do my best to "fix" it.

donquixote: I searched for `[][]` in build/core, and found that the arrays keys are almost never properly documented.
Although some positive cases exist.
And one interesting snippet here :) 
https://git.drupalcode.org/project/drupal/-/blame/9.4.x/core/tests/Drupal/Tests/Core/Extension/ExtensionDiscoveryTest.php#L106
```php
   * @return string[][]
   *   Format: $[$type][$name] = $yml_file
   *   E.g. $['module']['system'] = 'system.info.yml'
   */
  protected function populateFilesystemStructure(array &$filesystem_structure) {
```

But I will use something else, no worries.

modules/oe_whitelabel_paragraphs/tests/modules/oe_whitelabel_legacy_paragraphs_test/oe_whitelabel_legacy_paragraphs_test.install
donquixote: No. I am dropping them all.

modules/oe_whitelabel_paragraphs/tests/src/Functional/InstallTest.php
donquixote: Fyi, I prefer to put comment inside the `if ()`, because it eliminates the need for conditional phrases in the comment.
I don't know of any convention that would prevent this, and it is generally better.

```php
// Skip if node not found.
if ($node === NULL) {
  return;
}
```
vs
```php
if ($node === NULL) {
  // Node was not found. Skip.
  return;
}
```

drishu (approved):

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.install
```diff
@@ -39,3 +72,128 @@ function oe_whitelabel_paragraphs_install($is_syncing): void {
...
+  $field_names_by_bundle = [];
+  foreach ($field_names_map as $dest_field_name => $source_field_name) {
+    if (!isset($fields_map[$source_field_name])) {
+      // No field to migrate from.
+      continue;
+    }
+    if (!isset($fields_map[$dest_field_name])) {
+      // No target field to migrate to - this is unexpected.
+      continue;
+    }
```
brummbar: This cannot be true, as we just imported the field configurations during install. So either we drop this, or we throw an exception.

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.install
```diff
@@ -39,3 +72,128 @@ function oe_whitelabel_paragraphs_install($is_syncing): void {
...
+    // Find bundles where both source and destination fields exist.
+    foreach (array_intersect_key(
+      $fields_map[$dest_field_name]['bundles'],
+      $fields_map[$source_field_name]['bundles'],
+    ) as $bundle) {
```
brummbar: This makes me wonder. Basically we are getting all the bundles were these fields are present. But since this code will be run only as a standalone install, this means that the (new) field instances have just being imported and they are present only there. Which means that we can rely on a static mapping, and we just check if the source field is present.

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.install
```diff
@@ -39,3 +72,128 @@ function oe_whitelabel_paragraphs_install($is_syncing): void {
...
+}
+
+/**
+ * Migrates field data from the old oe_bootstrap_theme_paragraphs module.
+ *
+ * This should happen through a batch process, e.g. via $sandbox in a
+ * hook_update_N(). Unfortunately, hook_install() does not support batch
+ * processes.
+ *
+ * See https://atendesigngroup.com/articles/programmatically-copy-field-data-drupal-8.
```
brummbar: There's no real useful information there for what we are doing here, so let's drop this.

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.install
```diff
@@ -39,3 +72,128 @@ function oe_whitelabel_paragraphs_install($is_syncing): void {
...
+  // Load all the paragraph ids.
+  $query = $paragraphs_storage->getQuery();
+  $query->allRevisions();
+  $query->condition('type', array_keys($field_names_by_bundle), 'IN');
+  $paragraph_ids = $query->execute();
+
+  foreach ($paragraph_ids as $revision_id => $paragraph_id) {
+    $paragraph_revision = $paragraphs_storage->loadRevision($revision_id);
+    if (!$paragraph_revision) {
+      // Revision not found - this is unexpected, but survivable.
```
brummbar: I don't like much this type of comment, as it sounds like we don't know how the system is supposed to behave.
We just loaded a revision that we queried for, so either we assume that it's correctly loaded and we move forward, or we throw an exception. In both scenarios the only way this is not loaded is if mysql is acting funky, so it's correct to break the otherwise partial update path.

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.install
```diff
@@ -39,3 +72,128 @@ function oe_whitelabel_paragraphs_install($is_syncing): void {
...
+      continue;
+    }
+    $field_names_map = $field_names_by_bundle[$paragraph_revision->bundle()] ?? [];
+    if (!$field_names_map) {
+      // No field names to migrate - this is unexpected.
```
brummbar: This also can never happen, as we literally loaded the paragraph after a query with a specific type in it.

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.install
```diff
@@ -39,3 +72,128 @@ function oe_whitelabel_paragraphs_install($is_syncing): void {
...
+    $paragraph_revision->save();
+  }
+}
+
+/**
+ * Removes legacy field instances from oe_bootstrap_theme_paragraphs module.
+ *
+ * @param string[][] $field_names_by_bundle
+ *   Legacy field names, indexed by paragraph type and destination field name.
+ *   (The destination field name is not relevant for this function.)
```
brummbar: This can be inferred from the code below.

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.install
```diff
@@ -39,3 +72,128 @@ function oe_whitelabel_paragraphs_install($is_syncing): void {
...
+ */
+function _oe_whitelabel_paragraphs_install_drop_legacy_fields(array $field_names_by_bundle): void {
+  foreach ($field_names_by_bundle as $bundle => $field_names) {
+    foreach ($field_names as $field_name) {
+      $definition = FieldConfig::loadByName('paragraph', $bundle, $field_name);
+      if ($definition === NULL) {
+        // Field no longer exists. This is unexpected, but can be ignored.
```
brummbar: This won't be possible, as the entry wouldn't be in here otherwise.

modules/oe_whitelabel_paragraphs/tests/src/Functional/InstallTest.php
```diff
@@ -0,0 +1,209 @@
...
+   * @var \Drupal\Core\Extension\ModuleHandlerInterface
+   */
+  protected $moduleHandler;
+
+  /**
+   * Module installer.
+   *
+   * @var \Drupal\Core\Extension\ModuleInstallerInterface
+   */
+  protected $moduleInstaller;
```
brummbar: In browser tests it's recommended to use services directly through \Drupal::service() , to ensure that they are kept in sync with the real Drupal installation.

modules/oe_whitelabel_paragraphs/tests/src/Functional/InstallTest.php
```diff
@@ -0,0 +1,209 @@
...
+    ];
+
+    $ids = [];
+    foreach ($paragraphs_data as $name => $paragraph_data) {
+      $paragraph = Paragraph::create($paragraph_data);
+      $paragraph->save();
+      $ids[$name] = $paragraph->id();
+    }
+
+    self::assertTrue($this->moduleHandler->moduleExists('oe_whitelabel_legacy_paragraphs_test'));
```
brummbar: Let's use `$this->` for the assertions. While the methods are static and can be used as such, core uses $this invocation.

modules/oe_whitelabel_paragraphs/tests/src/Functional/InstallTest.php
```diff
@@ -0,0 +1,209 @@
...
+    self::assertFalse($this->moduleHandler->moduleExists('oe_whitelabel_paragraphs'));
+    self::assertTrue($this->moduleInstaller->install(['oe_whitelabel_paragraphs']));
+    self::assertFalse($this->moduleHandler->moduleExists('oe_whitelabel_paragraphs'));
```
brummbar: What are we asserting here? That the services are out of sync? There's no need for it.

modules/oe_whitelabel_paragraphs/tests/src/Functional/InstallTest.php
```diff
@@ -0,0 +1,209 @@
...
+        $actual_updated[$name][$field_name] = $updated_paragraph->get($field_name)->value;
+      }
+      foreach ($expected_deleted[$name] as $field_name => $deleted) {
+        $actual_deleted[$name][$field_name] = !$updated_paragraph->hasField($field_name);
+      }
+    }
+
+    // Compare the reports to the expected values.
+    self::assertSame($expected_created, $actual_updated);
+    self::assertSame($expected_deleted, $actual_deleted);
```
brummbar: Tests look good, but can we add at least one revision on one of them, to test that part? One of them is enough.

brummbar (changes requested):
Tested, works good, I did a local test with around 20 paragraphs and I felt the speed was already impacted, but the code has nothing that can be improved on. We might have to ask UCPKN to run this on their database to see if it's too slow already.

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.install
donquixote: Actually it might be true, if paragraph types have been removed on the target system.
But this would have blown up earlier I suppose.
Either way, let's consider this case not supported.

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.install
donquixote: True. And we prevent running on unknown paragraph types that use the same fields.

brummbar (approved):

PR #100, opened 2022-03-24
OEL-1317: Improve navbar search form.

escuriola:


GilNovacomm (dismissed):

modules/oe_whitelabel_search/src/Form/SearchForm.php
```diff
@@ -48,14 +48,14 @@ public function buildForm(array $form, FormStateInterface $form_state, array $co
     $form_state->set('oe_whitelabel_search_config', $config);
     $input_value = '';
 
-    $form['#attributes']['class'][] = 'bcl_search_form';
+    $form['#attributes']['class'][] = 'd-flex mt-3 mt-lg-0';
```
drishu: lets add classes only in form--oe-whitelabel-search-form.html.twig

modules/oe_whitelabel_search/templates/form--oe-whitelabel-search-form.html.twig
```diff
@@ -12,6 +12,6 @@
  * @ingroup themeable
  */
 #}
-<form{{ attributes.addClass('d-flex', 'bcl-search-form', 'submittable') }}>
+<form{{ attributes.addClass('d-flex', 'mt-3', 'mt-lg-0', 'w-100', 'submittable') }}>
```
drishu: keep only d-flex mt-3 mt-lg-0

drishu (changes requested):

templates/overrides/search/form-element--search-api-autocomplete.html.twig
```diff
@@ -48,6 +48,7 @@
 #}
 {%
   set classes = [
+  'w-100',
```
drishu: if we make the changes I proposed then (from my tests) this is not needed anymore, but after you make the other changes please double check and let me know.

modules/oe_whitelabel_search/templates/block--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,34 @@
+{#
```
drishu: we should move all the templates from this folder to /templates/search

modules/oe_whitelabel_search/templates/form--oe-whitelabel-search-form.html.twig
drishu: do we need "submittable" @escuriola ?

modules/oe_whitelabel_search/templates/block--oe-whitelabel-search-form.html.twig
drishu: don't forget about this one

modules/oe_whitelabel_search/templates/block--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * @file
+ * Override theme implementation to display block oe_whitelabel-search-form.
+ *
+ * Available variables:
```
drishu: as we did in the oebt PR lets remove these as they are there in block.html.twig already

drishu (changes requested):
2 issues I saw now
Screenshot 2022-04-14 at 10 04 03
we need a template for the navbar_right region and wrap {{ content}} in a div.nav
Screenshot 2022-04-14 at 10 04 15
we need to add class ps-0 to the site name because  we have to much passing on the left, or get rid of the "bcl-header__site-name" class from the link and only leave the one on the parent (closer to what is in netlify)

templates/search/block--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * @file
+ * Theme override for the search box block.
+ *
+ * @see ./core/modules/system/templates/block.html.twig
+ */
+#}
+{% block content %}
+  {{ content }}
+{% endblock %}
```
donquixote, line 11: The `{% block %}` is not really useful if we have no wrappers or other elements in the template.
In core, block templates without other stuff don't have `{% block %}`, e.g. `build/core/modules/system/templates/block--system-messages-block.html.twig`.
In whitelabel, we have `templates/overrides/navigation/block--languageswitcherinterfacetext.html.twig`, which does have the `{% block %}` syntax.

The only reason we would keep it is if we want to add wrapper elements or a title in the future, and we want inheriting templates to also get this change. But is this a realistic scenario?

I don't want to be the one who makes the final decision on this.

templates/overrides/navigation/block--system-branding-block.html.twig
```diff
@@ -26,7 +26,7 @@
             </a>
           {% endif %}
           {% if site_name is not empty %}
-            <div class="bcl-header__site-name site-name{{ bcl_component_library == 'neutral' ? ' h5 d-inline-block d-lg-none' }}">
+            <div class="bcl-header__site-name ps-0 site-name{{ bcl_component_library == 'neutral' ? ' h5 d-inline-block d-lg-none' }}">
```
donquixote, line 29: The padding-left that we want to suppress with `ps-0` only applies in large viewports where the entire thing is hidden anyway. So do we need the `ps-0` at all?

Here is what I find in the site css (in showcase), coming from BCL:
```css
@media (min-width: 992px) {
    .bcl-header--ec .bcl-header__project.light,.bcl-header__project.light,.ec__header .bcl-header__project.light {
        background:#fff
    }
    .bcl-header__site-name {
        font-weight: 500;
        padding-left: 1.5rem
    }
```

donquixote:
Seems mostly ok.
For the two points that I found I have no final opinion.
Either way we need to rebase.

drishu (dismissed):

drishu (dismissed):

modules/oe_whitelabel_search/src/Form/SearchForm.php
```diff
@@ -48,14 +48,12 @@ public function buildForm(array $form, FormStateInterface $form_state, array $co
...
 
-    $form['#attributes']['class'][] = 'bcl_search_form';
-
     if (!empty($config['input']['name'])) {
       $input_value = $this->getRequest()->get($config['input']['name']);
     }
 
     $form['search_input'] = [
-      '#prefix' => '<div class="bcl-search-form__group">',
+      '#prefix' => '<div class="bcl-search-form__group w-100">',
```
donquixote, line 56: I think `flex-grow-1` would be more logical here. But `w-100` does the job, so let's keep it.

templates/overrides/navigation/block--system-branding-block.html.twig
donquixote: I won't block the PR for this, but still would be interested in why we do this.

donquixote (commented):
I am generally ok with the changes but I would wish some response to my questions.

modules/oe_whitelabel_search/templates/form--oe-whitelabel-search-form.html.twig
donquixote: The "submittable" is gone. Resolving this.

templates/overrides/navigation/block--system-branding-block.html.twig
donquixote: Actually the ps-0 is fine here, it is needed for EC header.

donquixote:
Happy now, but need rebase.

drishu (dismissed):

donquixote (dismissed):

drishu (approved):

donquixote (approved):

PR #101, opened 2022-03-25
OEL-1227: Move content banner into a block.

donquixote:
[OEL-1227](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1227)


modules/oe_whitelabel_helper/src/Plugin/Block/ContentBannerBlock.php
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function blockAccess(AccountInterface $account): AccessResult {
+    /** @var \Drupal\node\NodeInterface $node */
+    $node = $this->getContextValue('node');
+
+    /** @var \Drupal\Core\Entity\EntityViewModeInterface|null $view_mode_object */
+    $view_mode_object = \Drupal::entityTypeManager()
```
brummbar: Y U NO ContainerFactoryPluginInterface?

modules/oe_whitelabel_helper/src/Plugin/Block/ContentBannerBlock.php
```diff
@@ -0,0 +1,50 @@
...
+      ->getStorage('entity_view_display')
+      ->load('node.' . $node->bundle() . '.oe_w_content_banner');
+
+    return AccessResult::allowedIf($view_mode_object !== NULL);
```
brummbar: Just to be extra careful, let's add a check if the current user can view the node. Most of the time this will be rendered in the node view page, so the page access altogether will be denied, but it's safer to add the check.

modules/oe_whitelabel_helper/src/Plugin/Block/ContentBannerBlock.php
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build(): array {
+    /** @var \Drupal\Core\Entity\EntityInterface $entity */
+    $entity = $this->getContextValue('node');
+    $view_builder = \Drupal::entityTypeManager()->getViewBuilder($entity->getEntityTypeId());
```
brummbar: Same as above, dependency injection.

brummbar (changes requested):

oe_whitelabel.info.yml
```diff
@@ -23,6 +23,7 @@ regions:
   header_top: Header top
   header_left: Header left
   header_right: Header right
+  banner: Banner
```
brummbar: Maybe we could call this in a more generic way, like "Page header" or "Content top" or something similar.

brummbar (changes requested):

modules/oe_whitelabel_helper/src/Plugin/Block/ContentBannerBlock.php
donquixote: Strange, I probably wanted to do this and then I didn't..

oe_whitelabel.info.yml
donquixote: You mean use one of the existing regions?
My goal was to put this outside of `class="container"`  elements, to have natural full width.
It is not fully needed, because the element itself has `100vw`, but if we already move it to a separate region, we can go all the way.

In fact the `width: 100vw` is a bit clumsy, because it does not subtract the width of the vertical scrollbar. So it means we need an `overflow-x: hidden` somewhere.
But as long as the 100vw is in bcl, it does not really matter which region we use..

What's worse, all the regions following this one are within `class="row"`, which messes with the container padding.
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1226?jql=project%20%3D%20oel%20AND%20text%20~%20%22container%20padding%20row%22
In our current solution, this required us to add classes from outside to mitigate the problem.

donquixote: One question though is how this should be placed in relation to breadcrumb and tabs.

donquixote: Fyi, the first comment in OEL-1226 has a screenshot that shows a symptom of the 100vw problem. Part of the text for "full-width section for bcl-content-banner" is cut off on the left.

modules/oe_whitelabel_helper/src/Plugin/Block/ContentBannerBlock.php
donquixote: done

modules/oe_whitelabel_helper/src/Plugin/Block/ContentBannerBlock.php
donquixote: I added the access check.

I was wondering if we need to add the node as cacheable dependency. But it seems we don't, because the node is coming from a context, and therefore the caching is already sensitive to that.

I also wondered if we should add a test for the cache tags context. But as there are no cache tags, it would be quite pointless. Also, I did not find any such tests in oe_theme.

oe_whitelabel.info.yml
donquixote: A good experiment is to put a block into the sidebar region.
We don't want the banner to show up next to the sidebar.

brummbar: The place where you put the new region is perfect, I didn't like the name much. We had a chat with @drishu and we will follow the naming used in Olivero core theme: `hero: 'Hero (full width)'`

brummbar: The place where you put the new region is perfect, I didn't like the name much. We had a chat with @drishu and we will follow the naming used in Olivero core theme: `hero: 'Hero (full width)'`

brummbar (dismissed):

tests/src/Functional/ContentEventRenderTest.php
```diff
@@ -91,25 +87,30 @@ protected function setUp(): void {
...
+    $this->drupalGet('node/' . $node->id());
 
-    // Assert content banner title.
+    /** @var \Symfony\Component\BrowserKit\AbstractBrowser $client */
+    $client = $this->getSession()->getDriver()->getClient();
+    $crawler = $client->getCrawler();
+
+    // Select the content banner element.
     $content_banner = $crawler->filter('.bcl-content-banner');
+    $this->assertSame(1, $content_banner->count(), 'Content banner found.');
```
drishu:
```suggestion
    $this->assertSame(1, $content_banner->count(), 'Content banner not found.');
```
the message displays when this fails, same in news pls

modules/oe_whitelabel_helper/oe_whitelabel_helper.info.yml
```diff
@@ -5,7 +5,13 @@ package: OpenEuropa Whitelabel Theme
...
 dependencies:
+  - drupal:node
   - oe_bootstrap_theme:oe_bootstrap_theme_helper
   - openeuropa:oe_corporate_blocks
 
 'interface translation project': oe_whitelabel_helper
+
+config_devel:
+  optional:
+    # A view mode should not depend on a theme, so it is exported in a module.
```
drishu: needles to say, as in not needed, lets add comments only where clarifications are needed

drishu (changes requested):
Other than these two small points very nice changes.

tests/src/Functional/ContentEventRenderTest.php
donquixote: I found this in Drupal core a lot. But I now learn this is a relic from the past, and should go:
https://www.drupal.org/project/coding_standards/issues/2057905#comment-9470963
So I will change it.

modules/oe_whitelabel_helper/oe_whitelabel_helper.info.yml
donquixote: I have a habit to add a comment whenever I was not sure about something myself and had to look it up or learn it.

donquixote: But yeah let's drop, otherwise we will have comments like this all over the place.

tests/src/Functional/ContentEventRenderTest.php
donquixote: After talk with Francesco I am dropping the comment and use assertCount().

drishu (approved):

brummbar (approved):

PR #110, opened 2022-04-04
OEL-1390: Use a formatter instead of 'name' view mode for stakeholder in project detail page.

donquixote:


abel-santos-corral (approved):
Checked what's been modified, looks good. I did a test on my localhost and checked that the label is used now, and it worked fine. Passing to QA.

composer.json
```diff
@@ -81,6 +81,11 @@
                 }
             }
         },
+        "patches": {
+            "drupal/entity_reference_revisions": {
+                "https://www.drupal.org/project/entity_reference_revisions/issues/2937835": "https://www.drupal.org/files/issues/2021-03-26/entity_reference_revisions-field_formatter_label-2937835-36.patch"
```
drishu, line 86: add a line in the readme of the submodule mentioning that this patch is needed and whatever other lines from composer 
something like this https://github.com/openeuropa/oe_theme/tree/3.x/modules/oe_theme_content_project#required-contrib-modules
and for the patch something like this
https://github.com/openeuropa/oe_content/blob/2.x/modules/oe_content_project/README.md

drishu (changes requested):

drishu (approved):

PR #112, opened 2022-04-05
OEL-1159: Adapt starter content entities to card last changes.

escuriola:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1159

### DIFF
--- a/modules/oe_whitelabel_starter_event/templates/node--oe-sc-event--teaser.html.twig
+++ b/modules/oe_whitelabel_starter_event/templates/node--oe-sc-event--teaser.html.twig
@@ -13,9 +13,10 @@
     variant: 'search',
     title: _title,
     text: content.oe_summary,
-    image: image|merge({
+    image: image ? {
       path: image.src,
-    }),
+      alt: image.alt,
+    },
     meta: [
       content.oe_sc_event_dates,
     ],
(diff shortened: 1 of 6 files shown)


donquixote:
Looks good to me!

modules/oe_whitelabel_starter_event/templates/node--oe-sc-event--teaser.html.twig
```diff
@@ -7,17 +7,19 @@
...
     title: _title,
     text: content.oe_summary,
-    image: image,
-    content: _content
+    image: image ? {
+      path: image.src,
+      alt: image.alt,
+    },
+    meta: [
+      content.oe_sc_event_dates,
```
donquixote, line 21: I suspect that `content.oe_sc_event_dates` contains some element properties that should not be treated as render elements. Use `|field_value` to clean up.
```suggestion
      content.oe_sc_event_dates|field_value,
```

drishu:
```suggestion
      content.oe_sc_event_location|field_value,
      content.oe_sc_event_dates|field_value
```

and use **inline address formatter** for the location

drishu (changes requested):

drishu (dismissed):

drishu (dismissed):

modules/oe_whitelabel_starter_event/templates/node--oe-sc-event--teaser.html.twig
drishu: in the end we wont do it because the address can be quite long, so we need to be able to configure this. Postponed.

drishu (approved):

donquixote (approved):

PR #122, opened 2022-04-08
OEL-1421: Fix cosmetic changes.

Maxfire:
Fixed cosmetic, tests provided on showcase.

### DIFF
--- a/modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
+++ b/modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
@@ -119,16 +119,6 @@ function _oe_whitelabel_extra_project_preprocess_inpage_nav(array &$variables): 
   /** @var \Drupal\node\NodeInterface $node */
   $node = $variables['node'];
   $variables['inpage_navigation_links'] = [];
-  $variables['project_details_group'] = FALSE;
-
-  // Get field groups.
-  foreach ($variables['elements']['#fieldgroups'] as $group) {
-    foreach ($group->children as $field) {
-      if (isset($node->{$field}) && !empty($node->{$field}->getValue())) {
-        $variables['project_details_group'] = TRUE;
-      }
-    }
-  }
 
   $fields = [
     'oe_summary',
--- a/modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
+++ b/modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
@@ -118,14 +118,24 @@ function _oe_whitelabel_extra_project_preprocess_featured_media(&$variables) {
 function _oe_whitelabel_extra_project_preprocess_inpage_nav(array &$variables): void {
   /** @var \Drupal\node\NodeInterface $node */
   $node = $variables['node'];
-
   $variables['inpage_navigation_links'] = [];
+  $variables['project_details_group'] = FALSE;
+
+  // Get field groups.
+  foreach ($variables['elements']['#fieldgroups'] as $group) {
+    foreach ($group->children as $field) {
+      if (isset($node->{$field}) && !empty($node->{$field}->getValue())) {
+        $variables['project_details_group'] = TRUE;
+      }
+    }
+  }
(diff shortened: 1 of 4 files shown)


templates/content/node--oe-project--teaser.html.twig
```diff
@@ -13,21 +13,26 @@
     label: _item,
   }]) %}
 {% endfor %}
+{% if content.oe_project_dates|field_value is not empty %}
+  {% set _content %}
+    <span class="text-muted text-nowrap me-4-5">{{ content.oe_project_dates }}</span>
+  {% endset %}
+{% endif %}
+{% if image is not empty %}
```
abel-santos-corral: Consider if it's better to move this below as:
  'image': (image is not empty) ? image|merge({
    'path': image.src,
  }) : {},

abel-santos-corral (changes requested):

abel-santos-corral (approved):
Approved

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
```diff
@@ -118,14 +118,24 @@ function _oe_whitelabel_extra_project_preprocess_featured_media(&$variables) {
...
   /** @var \Drupal\node\NodeInterface $node */
   $node = $variables['node'];
-
   $variables['inpage_navigation_links'] = [];
+  $variables['project_details_group'] = FALSE;
+
+  // Get field groups.
+  foreach ($variables['elements']['#fieldgroups'] as $group) {
+    foreach ($group->children as $field) {
+      if (isset($node->{$field}) && !empty($node->{$field}->getValue())) {
```
drishu:
```suggestion
      if (isset($node->{$field}) && !$node->{$field}->isEmpty()) {
```

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
```diff
@@ -118,14 +118,24 @@ function _oe_whitelabel_extra_project_preprocess_featured_media(&$variables) {
 function _oe_whitelabel_extra_project_preprocess_inpage_nav(array &$variables): void {
   /** @var \Drupal\node\NodeInterface $node */
   $node = $variables['node'];
-
   $variables['inpage_navigation_links'] = [];
+  $variables['project_details_group'] = FALSE;
+
+  // Get field groups.
+  foreach ($variables['elements']['#fieldgroups'] as $group) {
```
drishu: this is not completely correct, you should target only the groups that are part of the details group

templates/content/node--oe-project--teaser.html.twig
```diff
@@ -13,21 +13,22 @@
     label: _item,
   }]) %}
 {% endfor %}
+{% if content.oe_project_dates|field_value is not empty %}
```
drishu, line 17: remove the entire if

templates/content/node--oe-project--teaser.html.twig
```diff
@@ -13,21 +13,22 @@
...
-</article>
+  <article{{attributes}}>
+    {{ pattern('card', {
+      variant: 'search',
+      title: _title,
+      text: content.oe_teaser|field_value,
+      image: (image is not empty) ? image|merge({
+        path: image.src,
+      }) : {},
+      content: _content,
```
drishu: we have meta now for this, see the latest card changes
```suggestion
      meta: [
        content.oe_project_dates|field_value
      ],
```

drishu (changes requested):

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
drishu: I looked deeper into the issue, the problem is that the html element group prints the title even if the group is empty. So remove this solution, hide the title from the group, implement the old solution but add also striptags before trim. Then the old code works, but you have to add the "Project details" title in the template.

drishu (approved):

PR #123, opened 2022-04-12
OEL-1357: README improvements

donquixote:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1357

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Some paragraphs are considered "internal", and only meant to be used inside othe
 - Listing item: To be used as item paragraph within 'Listing item block'.
 - Fact: To be used as item paragraph within 'Facts and figures'.
 
-## Usage as a dependency in another project
+## Usage as a dependency
 
 Website projects can use `oe_whitelabel` either as an active theme, or they can create a custom theme using `oe_whitelabel` as a base theme.
 
--- a/README.md
+++ b/README.md
@@ -112,6 +112,12 @@ Find more information on Docker Compose extension mechanism on [the official Doc
 
 #### Start the container
 
+If you have other (demonized) containers running, you might want to stop them first:
+
+```bash
+docker stop $(docker ps -q)
+```
+
 To start, run:
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -40,50 +40,9 @@ In order to have full working integration with paragraphs in your project, you m
 
 ## Development setup
 
-OpenEuropa Whitelabel theme uses [Webpack](https://webpack.js.org) to compile and bundle SASS and JS.
+### Using LAMP stack or similar
 
-Make sure you have Node and npm installed.
-
-You can read a guide on how to install node here: https://docs.npmjs.com/getting-started/installing-node
-
-If you prefer to use [Yarn](https://yarnpkg.com) instead of npm, install Yarn by following the guide [here](https://yarnpkg.com/docs/install).
-
-To install required Node.js dependencies run:
-
-
-`npm install` or `yarn install`.
-
-To build the final artifacts run:
-
-```bash
-npm run build
-```
-
-This will compile all SASS and JavaScript files into self-contained assets that are exposed as [Drupal libraries][11].
-
-In order to download all required PHP code run:
-
-```bash
-composer install
-```
-
-This will build a fully functional Drupal site in the `./build` directory that can be used to develop and showcase the
-theme.
-
-Before setting up and installing the site make sure to customize default configuration values by copying [runner.yml.dist](runner.yml.dist)
-to `./runner.yml` and override relevant properties.
-
-To set up the project run:
-
-```bash
-./vendor/bin/run drupal:site-setup
-```
-
-A post command hook (`drupal:site-setup`) is triggered automatically after `composer install`.
-It will make sure that the necessary symlinks are properly setup in the development site.
-It will also perform token substitution in development configuration files such as `phpunit.xml.dist`.
-
-The development site web root should be available in the `build` directory.
+This is not officially supported. You are on your own.
 
 ### Using Docker Compose
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 # The OpenEuropa Whitelabel theme
 
-## Paragraphs
+## Features
+
+### Paragraphs
 
 The paragraphs below are not yet themed therefore not recommended for usage:
 
--- a/README.md
+++ b/README.md
@@ -24,11 +24,9 @@ Website projects can use `oe_whitelabel` either as an active theme, or they can 
 
 The package is meant for Drupal projects that manage their dependencies with [Composer](https://www.drupal.org/docs/develop/using-composer/using-composer-to-manage-drupal-site-dependencies#managing-contributed).
 
-Ideally this project should be managed with [Docker](https://www.docker.com/get-docker) and
-[Docker Compose](https://docs.docker.com/compose/), but this is not a hard requirement.
+Ideally this project should be managed with [Docker](https://www.docker.com/get-docker) and [Docker Compose](https://docs.docker.com/compose/), but this is not a hard requirement.
 
 Check the [composer.json](composer.json) for required PHP version and other dependencies.
-
 
 ### Add the composer package
 
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Ideally this project should be managed with [Docker](https://www.docker.com/get-
 Check the [composer.json](composer.json) for required PHP version and other dependencies.
 
 
-### Installation (composer)
+### Add the composer package
 
 Add this manually in composer.json, or combine with existing entries:
 
@@ -141,7 +141,7 @@ docker-compose exec web rm -rf vendor/
 docker-compose exec web rm -rf build/
 ```
 
-#### Installing the package
+#### Install the package
 
 ```bash
 docker-compose exec -u node node npm install
@@ -152,7 +152,7 @@ docker-compose exec web ./vendor/bin/run drupal:site-install
 
 Using default configuration, the development site files should be available in the `build` directory and the development site should be available at: [http://127.0.0.1:8080/build](http://127.0.0.1:8080/build) or [http://web:8080/build](http://web:8080/build).
 
-#### Running the tests
+#### Run the tests
 
 To run the grumphp checks:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -108,7 +110,13 @@ If a service is defined in both files, Docker Compose merges the configurations.
 
 Find more information on Docker Compose extension mechanism on [the official Docker Compose documentation](https://docs.docker.com/compose/extends/).
 
-#### Usage
+#### Start the container
+
+If you have other (demonized) containers running, you might want to stop them first:
```
GilNovacomm: **daemonized

GilNovacomm (changes requested):
Rewrite is ok, just fix the typo requested.

README.md
```diff
@@ -123,7 +131,17 @@ However, if you'd like to daemonize it, you have to add the flag `-d`:
...
+
+If you already had the package installed, and want a clean start:
+
+```bash
+docker-compose exec web rm composer.lock
+docker-compose exec web rm -rf vendor/
+docker-compose exec web rm -rf build/
+```
+
+#### Installing the package
```
donquixote: Btw there is some inconsistency here: "Optionally purge ..." is in imperative form, "Running the tests" is in present participle form. Not sure what is preferable.

GilNovacomm: IMO present participle is always nicer to read, but the other subtopics are all in imperative form so to be consistent:
#### Start the containerS
#### Purge the existing installation (optional)
#### Run the tests

donquixote: Some interesting references
https://developers.google.com/style/headings
https://writing.stackexchange.com/questions/1545/verb-tense-for-technical-document-titles
https://ffeathers.wordpress.com/2013/10/12/death-of-the-gerund-in-technical-documentation/
https://idratherbewriting.com/2008/02/11/are-gerunds-in-topic-titles-problematic-in-search-results/

There is no final conclusion, but I find more sources that recommend the imperative or "bare infinitive".

donquixote: However, not every headline is the same:
- Some headlines describe steps in a sequence. Here the imperative form seems most suitable.
- Other headlines describe actions we can do again at any time, such as running the tests.

Still, having different verb forms in the same sequence of titles feels wrong, so I will go for consistency.

GilNovacomm (dismissed):

README.md
```diff
@@ -1,6 +1,10 @@
 # The OpenEuropa Whitelabel theme
 
-## Paragraphs
+Sub-theme of [OpenEuropa Bootstrap base theme](https://github.com/openeuropa/bootstrap-component-library), with integration for many of the Drupal modules from the OpenEuropa ecosystem.
```
drishu: its not integration its theming, also It's not only openeuropa that it themes, but also d.org modules, so I suggest:
```suggestion
Sub-theme of [OpenEuropa Bootstrap base theme](https://github.com/openeuropa/bootstrap-component-library), provides theming for OpenEuropa library features.
```

README.md
```diff
@@ -1,6 +1,10 @@
...
+
+## Features
+
+### Paragraphs
```
drishu, line 7: we have a submodule for this now, I'd rather move this there and maybe also mention any composer dependencies for it.

Then, instead of this, list the companion modules we have like here https://github.com/openeuropa/oe_theme#companion-sub-modules

README.md
```diff
@@ -12,78 +16,94 @@ Some paragraphs are considered "internal", and only meant to be used inside othe
 - Listing item: To be used as item paragraph within 'Listing item block'.
 - Fact: To be used as item paragraph within 'Facts and figures'.
 
-## Requirements
-
-This depends on the following software:
-
-* [PHP 7.3](http://php.net/)
+## Usage as a dependency in another project
```
drishu: this is a bit wordy

README.md
```diff
@@ -12,78 +16,94 @@ Some paragraphs are considered "internal", and only meant to be used inside othe
...
+```
+    "minimum-stability": "alpha",
+```
 
-If you prefer to use [Yarn](https://yarnpkg.com) instead of npm, install Yarn by following the guide [here](https://yarnpkg.com/docs/install).
+Require with composer:
 
-To install required Node.js dependencies run:
+```bash
+composer require openeuropa/oe_whitelabel
```
drishu: since we are in alpha lets mention it directly here and drop below paragraph

README.md
```diff
@@ -12,78 +16,94 @@ Some paragraphs are considered "internal", and only meant to be used inside othe
...
+```
 
+If composer.json does not allow alpha versions, a version string needs to be appended to the package name in the require command, and other dependencies with alpha versions also need to be required with explicit version strings. The version string could be `:^1.0@alpha`, or a specific version like `:^1.0.0-alpha7`.
 
-`npm install` or `yarn install`.
+```bash
+composer require openeuropa/oe_whitelabel:^1.0@alpha openeuropa/oe_bootstrap_theme:^1.0@alpha
```
drishu, line 69: whitelabel manages the oebt dependency, dont think you need to mention it explicitly. You manage min stability, you call oewt alpha, life is good

README.md
```diff
@@ -12,78 +16,94 @@ Some paragraphs are considered "internal", and only meant to be used inside othe
...
+```
 
-To build the final artifacts run:
+Review the installed package versions. Make sure that oe_whitelabel and oe_bootstrap theme are using 1.x or higher.
 
 ```bash
-npm run build
+composer show -i | grep oe_
 ```
```
drishu: think this is a bit much, lets rather mention in the readmes of the submodules what is required in composer

README.md
```diff
@@ -12,78 +16,94 @@ Some paragraphs are considered "internal", and only meant to be used inside othe
...
 
-This will build a fully functional Drupal site in the `./build` directory that can be used to develop and showcase the
-theme.
+# Required, if you use oe_paragraphs module, or if you copied any paragraph
+# types from that module.
+./vendor/bin/drush en oe_whitelabel_paragraphs
 
-Before setting up and installing the site make sure to customize default configuration values by copying [runner.yml.dist](runner.yml.dist)
-to `./runner.yml` and override relevant properties.
+# Other submodules are optional - check the /modules/ folder.
```
drishu, line 90: we will list the submodules as requested, then this bit is redundant

README.md
```diff
@@ -12,78 +16,94 @@ Some paragraphs are considered "internal", and only meant to be used inside othe
...
+./vendor/bin/drush en <modulename>
+```
 
-To set up the project run:
+Enable the theme and set as default:
 
 ```bash
-./vendor/bin/run drupal:site-setup
+./vendor/bin/drush config-set system.theme default oe_bootstrap_theme
```
drishu:
```suggestion
./vendor/bin/drush config-set system.theme default oe_whitelabel
```

drishu (changes requested):

README.md
donquixote: I don't like the construction of the sentence here. Either the entire thing should be a sentence with subject and verb, or none of it. I will try something.

donquixote: "with theming for ..."

README.md
donquixote: I agree with the direction, but I would rather do this in a separate PR where we add README files for all submodules.
I think the priority should be to merge this soon.
I don't like this section but we can postpone it.

README.md
donquixote: I just tried this again with an empty project where I only require `openeuropa/oe_whitelabel`.
The goal is a successful `require` command that gets something from `1.0*alpha` and not anything from `0.x`.

- If you have `"prefer-stable": false` and `"minimum-stability": "alpha"`, then you don't even need to specify a version string at all. I think people should avoid doing this.
- If you have `"prefer-stable": true` and `"minimum-stability": "alpha"`, then you need to specify a version string, but it is enough to require `^1.0` to avoid getting `0.x`. The cool thing is this can stay there even once 1.0 beta or stable is released.
- If you have `"prefer-stable": true` and you don't set `"minimum-stability"`, then you need to specify a version string with alpha for whitelabel _and_ oe_bootstrap_theme.

My concern was that some people/projects don't want to set `minimum-stability` because it can impact all packages.

I am trying something, let me know what you think.

README.md
donquixote: see above.

README.md
donquixote: The motivation were situations where me or other people wanted to install and got releases from 0.x, because these are considered stable.
I will keep it around for 2nd round, but I am not fully against removing it.

README.md
donquixote: Can we leave it for now, and make a follow up where we document submodules?
Also, even if we list submodules above, it should still be mentioned as an installation step.

I propose to keep it and then we can make a follow-up issue or PR where we say "Enable optional submodules - see above".

README.md
donquixote: oops :)

README.md
donquixote: Also not sure what this has to do with the submodules.

README.md
```diff
@@ -12,78 +16,92 @@ Some paragraphs are considered "internal", and only meant to be used inside othe
...
-If you prefer to use [Yarn](https://yarnpkg.com) instead of npm, install Yarn by following the guide [here](https://yarnpkg.com/docs/install).
+One option is to set `minimum-stability` and require the `^1.0` version.
+You should also set `prefer-stable` to mitigate the impact on other dependencies.
+Review this again when the site goes into production.
 
-To install required Node.js dependencies run:
+```bash
+composer config minimum-stability alpha
+composer config prefer-stable true
+composer require openeuropa/oe_whitelabel:^1.0
```
donquixote, line 63: While technically it is enough to specify `^1.0`, we might still want to write `^1.0@alpha` to be more explicit. TBD.

README.md
drishu: I'd remove it, but if you insist, its not a fault per say, so lets see what next reviewer thinks

drishu (approved):
not all is as I would like it but not blockers, so I approve

README.md
GilNovacomm: I think it would be overkill to specify ``^1.0@alpha`` since it is ready mentioning that the ``minimum-stability`` is alpha.
I am ok with this current version.

README.md
GilNovacomm: I think its ok as is, it helps to bring attention to the releases being used.

GilNovacomm (approved):

PR #125, opened 2022-04-19
OEL-1453: Consolidate theme regions

escuriola:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1453

### DIFF
--- a/oe_whitelabel.info.yml
+++ b/oe_whitelabel.info.yml
@@ -18,7 +18,7 @@ ckeditor_stylesheets:
   - assets/css/oe-bcl-default.min.css
 
 regions:
-  navbar_branding: 'Navbar branding'
+  header_branding: 'Header branding'
   header_top: 'Header top'
   navigation_left: 'Navigation left'
   navigation_right: 'Navigation right'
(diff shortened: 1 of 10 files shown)


templates/overrides/page/page.html.twig
```diff
@@ -30,15 +30,13 @@
  * Regions:
  * - page.navbar_branding: Items for the navbar brand region (logo).
  * - page.navbar_left: Items for the navbar left region (primary menu).
```
brummbar: This we don't have anymore, right?

oe_whitelabel.info.yml
```diff
@@ -18,19 +18,18 @@ ckeditor_stylesheets:
   - assets/css/oe-bcl-default.min.css
 
 regions:
-  navbar_branding: Navbar branding
-  navbar_right: Navbar right
-  header_top: Header top
-  header_left: Header left
-  header_right: Header right
+  navbar_branding: 'Navbar branding'
```
brummbar: Just realised this one now is out of "style": we removed navbar term completely but for this. Let's rename it to "header_branding".

brummbar (changes requested):
Two minor remarks.

brummbar (approved):

drishu (approved):

PR #126, opened 2022-04-22
OEL-1374: Search page look & feel update

escuriola:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1374


oe_whitelabel.theme
```diff
@@ -168,3 +138,21 @@ function oe_whitelabel_preprocess_media__document__default(&$variables) {
...
+ * Implements hook_theme_suggestions_HOOK_alter() for form_element.
+ */
+function oe_whitelabel_theme_suggestions_form_element_alter(array &$suggestions, array $variables) {
+  if (isset($variables['element']['#id']) && str_contains($variables['element']['#id'], 'sort-bef-combine')) {
+    $suggestions[] = 'form_element__sort_by_bef';
+  }
+}
+
+/**
+ * Helper function to preprocess sort by bef label.
```
drishu: this is a preprocess, not a helper

templates/overrides/page/page.html.twig
```diff
@@ -121,9 +121,9 @@
         {% endif %}
 
         {% if page.content or page.content_top %}
-          <div class="col-12 {{ page.sidebar_left ? 'col-lg-9' }}">
+          <div class="col-12 {{ page.sidebar_left ? 'col-lg-9  col-xxl-8' }}">
```
drishu, line 124: ok, but if we do this then we should also change the sidebar wrapper class from `col-md-3` to `col-12 col-lg-3`

templates/overrides/page/page.html.twig
```diff
@@ -121,9 +121,9 @@
...
             {% if page.content_top %}
-              <div class="mb-3 clearfix">
+              <div class="row">
                 {{ page.content_top }}
               </div>
               <hr class="d-none d-md-block mt-4 mb-4-75">
```
drishu: the hr needs to have `d-none d-md-block mt-4-5` now

templates/overrides/search/block--facets-summary-block.html.twig
```diff
@@ -1,46 +1,30 @@
...
- *   displayed after the main title tag that appears in the template.
- *
- * @see template_preprocess_block()
- *
- * @ingroup themeable
+ * @see ./core/themes/stable/templates/block/block.html.twig
  */
 #}
-{% if label and content['#items'][0]['#theme'] is defined and content['#items'][0]['#theme'] == 'facets_summary_count' %}
+{% if label and content['#items'][0]['#theme'] is defined and content['#items'][1]['#theme'] == 'facets_summary_count' %}
```
drishu: your checking 
content['#items']**[0]**['#theme'] defined
but then 
content['#items']**[1]**['#theme'] == 'facets_summary_count'
is this intended ?

if the issue here is that we are uncertain that the correct key is 0, then we should rather loop through
```
{% set results_count = '' %}
{% if label and content['#items'] is defined %}
  {% for item in content['#items'] %}
    {% if item['#theme'] is defined and item['#theme'] == 'facets_summary_count' %}
      {% set results_count = item %}
    {% endif %}
  {% endfor %}
{% endif %}
```
but do let me know what your issue was

templates/overrides/search/block--facets-summary-block.html.twig
```diff
@@ -1,46 +1,30 @@
...
   {% spaceless %}
     {{ title_prefix }}
     {% if label %}
-      <h4 class="mb-4">
-        {{ label }}
-        {{ results_count }}
-      </h4>
+      <div class="col-md-6 col-lg-8 align-self-center">
+        <h4 class="mb-0">
+          {{ label }}
```
drishu: dono if its recent, but I see that now the label is wrapped in a span.text-capitalize

templates/overrides/search/block--views-exposed-filter-block--showcase-search-showcase-search-page.html.twig
```diff
@@ -0,0 +1,13 @@
+{#
+/**
+ * @file
+ * Theme block implementation to display sort by exposed filter form.
+ *
+ * @see ./core/themes/stable/templates/block/block.html.twig
+ */
+#}
+{% block content %}
+  <div class="col-md-6 col-lg-4 mt-3 mt-md-0" data-bef-auto-submit-full-form="" data-bef-auto-submit="" data-bef-auto-submit-delay="500" data-drupal-selector="views-exposed-form-showcase-search-showcase-search-page" id="block-showcase-exposed-sort-form">
```
drishu: things are wrong here: 
1. the filename contains showcase (we are in whitelabel)
2. the selector contains showcase  (we are in whitelabel)
3. you cant just hardcode these selectors and ids
4. overriding a specific views template is not ok
5. overriding a more general template is also not ok :), we need a different solution here

Solution:
remove this template and change the classes `col-md-6 col-lg-4 mt-3 mt-md-0` in the oe_whitelabel_float_end_sort

templates/overrides/search/form-element--sort-by-bef.html.twig
```diff
@@ -0,0 +1,36 @@
+{#
```
drishu: remove this file, do it from oe_whitelabel_float_end_sort, you can controle the entire sort form from there, this way the view gets the theming only when the plugin is selected

templates/overrides/search/views-view-unformatted.html.twig
```diff
@@ -0,0 +1,22 @@
...
+  {% for row in rows %}
+    {% set row_classes = [
+      default_row_class ? 'views-row',
+    ] %}
+    <div{{ row.attributes.addClass(row_classes) }}>
+      {{- row.content -}}
+    </div>
+  {% endfor %}
+{% endif %}
+<hr class="d-none d-md-block mt-4-5">
```
drishu: remove this file, the pager should receive this hr, this is for 3 good reasons:
1. the line should not appear if no pager is there
2. we cant target showcase_search_page
3. this hr should be consistent in all listings where a pager is needed

modules/oe_whitelabel_search/oe_whitelabel_search.module
```diff
@@ -0,0 +1,67 @@
...
+      $variables['result_items'][] = $item;
+    }
+  }
+
+}
+
+/**
+ * Implements hook_form_FORM_ID_alter().
+ */
+function oe_whitelabel_search_form_views_exposed_form_alter(&$form, FormStateInterface $form_state, $form_id) {
```
drishu: remove this and solve in oe_whitelabel_float_end_sort

modules/oe_whitelabel_search/oe_whitelabel_search.module
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @param array $variables
+ *   Render array variables.
+ */
+function oe_whitelabel_search_preprocess_facets_summary_item_list(array &$variables) {
+  $variables['result_items'] = [];
+  foreach ($variables['items'] as $item) {
+    if ($item['value']['#type'] === 'link') {
+      $item['value']['#attributes']['class'] = 'standalone';
+      $variables['result_items'][] = $item;
```
drishu: this is not a menu, we don't need to have a separate key to avoid conflicts

oe_whitelabel.theme
```diff
@@ -168,3 +138,21 @@ function oe_whitelabel_preprocess_media__document__default(&$variables) {
     $variables['translations'][] = $wrapper->toFileValueObject();
   }
 }
+
+/**
+ * Implements hook_theme_suggestions_HOOK_alter() for form_element.
+ */
+function oe_whitelabel_theme_suggestions_form_element_alter(array &$suggestions, array $variables) {
```
drishu: you can remove this

modules/oe_whitelabel_search/oe_whitelabel_search.module
```diff
@@ -0,0 +1,67 @@
...
+    foreach ($reset_classes as $reset_class) {
+      if (!isset($form['actions']['reset']['#attributes']['class']) || !in_array($reset_class, $form['actions']['reset']['#attributes']['class'], TRUE)) {
+        $form['actions']['reset']['#attributes']['class'][] = $reset_class;
+      }
+    }
+  }
+}
+
+/**
+ * Helper function to preprocess the facets_summary item list.
```
drishu: this is a preprocess, not a helper

modules/oe_whitelabel_search/oe_whitelabel_search.module
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @param array $variables
```
drishu: remove, hooks and preprocesses are like interfaces for drupal, so no need to define the variables as they are already documented

templates/overrides/search/facets-summary-item-list.html.twig
```diff
@@ -1,28 +1,14 @@
...
+ * @see ./modules/contrib/facets/modules/facets_summary/templates/facets-summary-item-list.html
  *
- * @see facets_summary_preprocess_facets_summary_item_list()
- *
- * @ingroup themeable
  */
 #}
-{%- if items -%}
-  {%- for item in items -%}
+{%- if result_items -%}
```
drishu: use the old stuff

templates/overrides/search/facets-summary-item-list.html.twig
```diff
@@ -1,28 +1,14 @@
...
+  {%- for item in result_items -%}
     {{ item.value }}
```
drishu: can't remember why we did `item.value` and not `item`, can you check if it works with the whole render array

oe_whitelabel.theme
```diff
@@ -168,3 +138,21 @@ function oe_whitelabel_preprocess_media__document__default(&$variables) {
...
+    $suggestions[] = 'form_element__sort_by_bef';
+  }
+}
+
+/**
+ * Helper function to preprocess sort by bef label.
+ */
+function oe_whitelabel_preprocess_form_element_label(&$variables) {
+  if (str_contains($variables['element']['#id'], 'sort-bef-combine')) {
+    $variables['attributes']['class'][] = 'mb-0';
```
drishu: we can do this in the oe_whitelabel_float_end_sort

modules/oe_whitelabel_search/oe_whitelabel_search.module
```diff
@@ -0,0 +1,67 @@
...
+ */
+function oe_whitelabel_search_preprocess_facets_summary_item_list(array &$variables) {
+  $variables['result_items'] = [];
+  foreach ($variables['items'] as $item) {
+    if ($item['value']['#type'] === 'link') {
+      $item['value']['#attributes']['class'] = 'standalone';
+      $variables['result_items'][] = $item;
+    }
+  }
+
```
drishu: remove this newline

drishu (changes requested):

templates/overrides/search/block--facets-summary-block.html.twig
escuriola: Before I realised that the content was overwritten and that was the reason for ignoring the facets-summary-item-list, I changed the configuration of the facet? adding a new option and this altered the order of the facet_summary_count. After that I confirm that was not needed but I didn't revert this change.

modules/oe_whitelabel_search/oe_whitelabel_search.module
escuriola: We need because the "result count" with the number of the total is also an item.

templates/overrides/search/facets-summary-item-list.html.twig
escuriola: See above

modules/oe_whitelabel_search/oe_whitelabel_search.module
```diff
@@ -0,0 +1,54 @@
...
+    foreach ($reset_classes as $reset_class) {
+      if (!isset($form['actions']['reset']['#attributes']['class']) || !in_array($reset_class, $form['actions']['reset']['#attributes']['class'], TRUE)) {
+        $form['actions']['reset']['#attributes']['class'][] = $reset_class;
+      }
+    }
+  }
+}
+
+/**
+ * Preprocess the facets_summary item list.
```
drishu:
```suggestion
 * Implements hook_preprocess_HOOK() for facet_summary_item_list.
```
this is the proper way of commenting this, valid also for the others

drishu (changes requested):
The remaining issues should be fixed

drishu (dismissed):
Actually, lets not block it for some comments

drishu (dismissed):

drishu (approved):

Maxfire (dismissed):

modules/oe_whitelabel_search/oe_whitelabel_search.module
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+declare(strict_types =  1);
+
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Implements hook_form_FORM_ID_alter() for facets_forms.
+ */
+function oe_whitelabel_form_facets_form_alter(&$form, FormStateInterface $form_state, $form_id) {
```
drishu:
```suggestion
function oe_whitelabel_form_facets_form_alter(&$form, FormStateInterface $form_state, $form_id): void {
```

modules/oe_whitelabel_search/oe_whitelabel_search.module
```diff
@@ -0,0 +1,54 @@
...
+  if (isset($form['actions'])) {
+    $action_classes = ['mt-4', 'd-grid', 'gap-4', 'd-md-block'];
+    foreach ($action_classes as $action_class) {
```
drishu: using functions like array_merge is faster

modules/oe_whitelabel_search/oe_whitelabel_search.module
```diff
@@ -0,0 +1,54 @@
...
+        $form['actions']['reset']['#attributes']['class'][] = $reset_class;
+      }
+    }
+  }
+}
+
+/**
+ * Implements hook_preprocess_HOOK() for facet_summary_item_list.
+ */
+function oe_whitelabel_search_preprocess_facets_summary_item_list(array &$variables) {
```
drishu: void

templates/overrides/search/block--facets-summary-block.html.twig
```diff
@@ -1,46 +1,25 @@
...
-        {{ label }}
-        {{ results_count }}
-      </h4>
+      <div class="col-md-6 col-lg-8 align-self-center">
+        <h4 class="mb-0"><span class="text-capitalize">{{ label }} </span> {{ results_count }}</h4>
+      </div>
     {% endif %}
     {{ title_suffix }}
-    {{ content }}
+    <div class="col-md-12 order-md-2 mt-3">
```
drishu, line 21: check if we need to wrap this in an `if content` or if that would break the layout

modules/oe_whitelabel_helper/src/Plugin/better_exposed_filters/sort/FloatEndSortWidget.php
```diff
@@ -22,10 +22,17 @@ class FloatEndSortWidget extends DefaultWidget {
    */
   public function exposedFormAlter(array &$form, FormStateInterface $form_state) {
     parent::exposedFormAlter($form, $form_state);
-    $form['#attributes']['class'][] = 'float-lg-end';
-    $form['#attributes']['class'][] = 'd-none';
-    $form['#attributes']['class'][] = 'd-md-flex';
-    $form['#attributes']['class'][] = 'align-items-baseline';
+    $form['sort_bef_combine']['#wrapper_attributes']['class'][] = 'float-md-end';
```
drishu: this is already better than what was done before in this PR and for our purpose ok for now,

BUT we should follow up to improve this using '#theme' like done in the other plugins (see Drupal\better_exposed_filters\Plugin\better_exposed_filters\sort\) AND we should only do this once we have the listing pages sort themed and merged so we have a unified theming, no duplications (so block the follow up on the sort theming ticket)

drishu (changes requested):
check also the not fixed newline, thx

templates/overrides/search/block--facets-summary-block.html.twig
```diff
@@ -1,46 +1,27 @@
...
-        {{ label }}
-        {{ results_count }}
-      </h4>
+      <div class="col-md-6 col-lg-8 align-self-center">
+        <h4 class="mb-0"><span class="text-capitalize">{{ label }} </span> {{ results_count }}</h4>
+      </div>
     {% endif %}
     {{ title_suffix }}
-    {{ content }}
+    {% if content['#items'][1] is not empty %}
```
drishu: so the count is part of the list, so it will never be empty, we can remove the it

drishu (dismissed):

drishu (approved):

Maxfire (approved):

PR #129, opened 2022-04-25
OEL-1371: Project status and progress js

donquixote:


templates/content/node--oe-project--full.html.twig
```diff
@@ -54,12 +88,14 @@
     {{ _self.field_with_header(content.oe_cx_achievements_and_milestone) }}
   {% endblock %}
 {% endset %}
+
 <div class="mt-md-4-75 mt-4">
```
drishu: to much whitespace now between the content and the hero, remove the div, maybe replace by `<article{{attributes}}>` up to you

modules/oe_whitelabel_extra_project/config/install/core.entity_view_display.node.oe_project.full.yml
```diff
@@ -165,16 +163,6 @@ content:
...
-    type: daterange_default
-    label: hidden
-    settings:
-      timezone_override: ''
-      format_type: oe_whitelabel_project_date
-      separator: '-'
-    third_party_settings: {  }
-    weight: 1
-    region: content
   oe_project_funding_programme:
```
drishu, line 177: Ups, I forgot to check this at last review, no skos should be linked
```
settings:
      link: false
```
this is valid for any skos field!

templates/content/node--oe-project--full.html.twig
```diff
@@ -4,46 +4,80 @@
...
     'path': '#project-details',
     'label': _project_details_title,
   }]|merge(inpage_navigation_links) %}
 {% endif %}
-{% set inpage_navigation_fields %}
+
+{% set inpage_navigation_content %}
+  {#
+    Reliably absorb top margin of first element.
+    @todo Add a utility class 'eat-margin-top' with :before style.
```
drishu, line 48: was the todo discussed with FRONT colleagues? I'd skip the comment all together, up to you, but the way I see it, its something that we add to reflect bcl composition, so I dont feel the need to explain it further

templates/content/node--oe-project--full.html.twig
```diff
@@ -4,46 +4,80 @@
...
+        }],
+      }),
+    } %}
+    {% if corporate_contributions is defined %}
+      {% include '@oe-bcl/project-status/project-contributions' with {
+        'corporate_contributions': corporate_contributions,
+        'chart': true,
+        'legend': _contribution_legend_args,
+      } only %}
+    {% else %}
```
drishu: even if I don't fill out one field, I dont get into this else, lets drop it, and report to FRONT, they should take care of this situation on their side in the template

resources/js/oe_whitelabel.project_status.js
```diff
@@ -0,0 +1,135 @@
...
+          });
+        });
+
+        // Reveal the entire section.
+        $element.removeClass('d-none');
+      });
+    }
+  };
+
+})(bootstrap, Drupal, jQuery);
```
drishu, line 61: Something is wrong with the status, below it should be planned/not-started, see
Screenshot 2022-04-27 at 09 48 30
and here ongoing no?
Screenshot 2022-04-27 at 09 50 43
Also the date formats displayed are wrong, should be what we use everywhere, like 22 February 2022

resources/js/oe_whitelabel.project_status.js
```diff
@@ -0,0 +1,135 @@
...
+   * @param {int} msEnd
+   *   End timestamp in milliseconds.
+   * @param {function(0|1|2)} setStatus
+   *   Callback to set the status: 0 = planned, 1 = ongoing, 2 = closed.
+   */
+  function animateStatus(msBegin, msEnd, setStatus) {
+    const msNow = Date.now();
+    if (msNow < msBegin) {
+      setStatus(0);
+      window.setTimeout(setStatus, msBegin - msNow, 1);
```
drishu: I appreciate the intention, and viewing the bigger picture, but, a few things:
1. We don't have requirements for this, so we should not have done it in the first place
2. Scripting like this is prone to error `Uncaught TypeError: Cannot read properties of null (reading 'classList')`
3. Animating elements on the page is unexpected, is an accessibility issue, and should be limited to situations where an action has been performed (ex a progress bar in admin when a bulk action is performed, or rotating an icon when hovering or clicking on an accordion item)
4. we should leave such work to the FRONT and UX, that is their department and such things need to be requested and validated by business first in the UX side
5. this only adds unnecessary work and complexity, harder to maintain, harder to test
6. not worth, as we are dealing with days, not hours or minutes, so the effect is not even visible.
There was one more concern about usability but I forgot while writing up the rest.

The js should run only once, on page load, and set the appropriate status and progress depending on the time intervals. Thats it. Don't forget about KISS.

resources/js/oe_whitelabel.project_status.js
```diff
@@ -0,0 +1,135 @@
...
+(function (bootstrap, Drupal, $) {
+
+  const colorClasses = [
+    'bg-secondary',
+    'bg-info',
+    'bg-dark',
+  ];
+
+  /**
+   * @param {int} msBegin
```
drishu: short desc is missing

resources/js/oe_whitelabel.project_status.js
```diff
@@ -0,0 +1,135 @@
...
+    else if (msNow < msEnd) {
+      setStatus(1);
+      window.setTimeout(setStatus, msEnd - msNow, 2);
+    }
+    else {
+      setStatus(2);
+    }
+  }
+
+  /**
```
drishu: short desc

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
```diff
@@ -144,5 +160,92 @@ function _oe_whitelabel_extra_project_preprocess_inpage_nav(array &$variables):
...
+ */
+function _oe_whitelabel_extra_project_preprocess_status_and_progress(array &$variables): void {
+  /** @var \Drupal\node\NodeInterface $node */
+  $node = $variables['node'];
+  /** @var \Drupal\datetime_range\Plugin\Field\FieldType\DateRangeItem|null $date_range_item */
+  $date_range_item = $node->get('oe_project_dates')->first();
+  if ($date_range_item === NULL) {
+    return;
+  }
+  $status_keys = ['planned', 'ongoing', 'closed'];
```
drishu: not used

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
```diff
@@ -144,5 +160,92 @@ function _oe_whitelabel_extra_project_preprocess_inpage_nav(array &$variables):
...
+  $status_labels = [t('Planned'), t('Ongoing'), t('Closed')];
```
drishu: think you can move this more down, and then you can merge the 2 ifs

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
```diff
@@ -144,5 +160,92 @@ function _oe_whitelabel_extra_project_preprocess_inpage_nav(array &$variables):
...
+  /* @see \Drupal\system\TimeZoneResolver::getTimeZone() */
+  /** @var string $timezone */
+  $timezone = \Drupal::config('system.date')->get('timezone.default') ?? 'UTC';
+  $get_timestamp = static function (string $date_string) use ($timezone): int {
+    return (new DrupalDateTime($date_string, $timezone))->getTimestamp();
+  };
+  // Project starts at the beginning of the first day at 00:00.
+  $t_start = $get_timestamp($date_range_item->value);
+  // Project ends at the end of the last day at 24:00.
+  $t_end = $get_timestamp($date_range_item->end_value . ' +1 day');
```
drishu, line 199: Not sure if its an issue or not (in this specific use case), because it passes to the next day in the calendar, so if end date is 31st it passes to 1st of next month. the difference is only 1 second but what I would have done is 
```suggestion
  $t_end = $get_timestamp($date_range_item->end_value . '  23:59:59');
```
this way you don't pass into the next day, then the comment "ends at the end of the last day" would be more correct, because 23:59:59 is the last second of the day, while when setting time to 24:00 it's actually the first second of the next day (you can double check that).

templates/content/node--oe-project--oe-w-content-banner.html.twig
```diff
@@ -0,0 +1,17 @@
...
+ * Project content banner display.
+ */
+#}
+{{ pattern('content_banner', {
+  'background': 'gray',
+  'title': label,
+  'content': content.oe_teaser,
+  'image': image,
+  'attributes': create_attribute().addClass(['ps-0']),
+  'badges': content.oe_subject ? content.oe_subject|field_value|default([])|map(_item => {
```
drishu: next level chaining of filters, but a bit hard 2 read, can we make this more readable for maintainability?

tests/src/Functional/ContentProjectRenderTest.php
```diff
@@ -4,29 +4,40 @@
 
 namespace Drupal\Tests\oe_whitelabel\Functional;
```
drishu: javascript tests belong in a different folder

tests/src/Functional/ContentProjectRenderTest.php
```diff
@@ -291,4 +320,102 @@ protected function getStorage(string $entity_type_id): EntityStorageInterface {
...
+   * Updates the project date, saves the node, and refreshes the page.
+   *
+   * @param \Drupal\node\NodeInterface $node
+   *   Node to update.
+   * @param string $begin
+   *   Start date string.
+   * @param string $end
+   *   End date string.
+   */
+  protected function setProjectDateRange(NodeInterface $node, string $begin, string $end): void {
```
drishu:
```suggestion
  protected function setProjectDateRangeAndRefresh(NodeInterface $node, string $begin, string $end): void {
```

But honestly a function should do 1 thing and 1 thing only, so the AndSomething should already tell you its nok.
The test is already full of drupalGets, I dont see why it would be an issue to continue, its part of the tests stroy to see you revisit the page.

tests/src/Functional/ContentProjectRenderTest.php
```diff
@@ -291,4 +320,102 @@ protected function getStorage(string $entity_type_id): EntityStorageInterface {
...
+  /**
+   * Asserts that the d-none class has been removed from project status.
+   */
+  protected function assertProjectStatusRevealed(): void {
+    $status_section = $this->assertSession()->elementExists('css', '.bcl-project-status');
+    $this->assertFalse($status_section->hasClass('d-none'));
+  }
+
+  /**
+   * Asserts start and end timestamp in project status section.
```
drishu:
```suggestion
   * Asserts 'start' and 'end' timestamp in project status section.
```

tests/src/Functional/ContentProjectRenderTest.php
```diff
@@ -291,4 +320,102 @@ protected function getStorage(string $entity_type_id): EntityStorageInterface {
...
+      'end_value' => (new DrupalDateTime($end, 'Europe/Brussels'))->format('Y-m-d'),
+    ];
+    $node->save();
+    $this->drupalGet($node->toUrl());
+  }
+
+  /**
+   * Asserts that the d-none class has been removed from project status.
+   */
+  protected function assertProjectStatusRevealed(): void {
```
drishu:
```suggestion
  protected function assertProjectStatusVisible(): void {
```

tests/src/Functional/ContentProjectRenderTest.php
```diff
@@ -244,9 +247,35 @@ public function testProjectRendering(): void {
           'definition' => '€22,30',
         ],
       ],
-    ];
-    $field_list_html = $description_lists[4]->getHtml();
-    $field_list_assert->assertPattern($fifth_field_list_expected_values, $field_list_html);
+    ], $description_lists[3]->getHtml());
+
+    // Set a project period that is fully in the past.
+    $this->setProjectDateRange($node, '2019-03-07', '2019-03-21');
```
drishu: setProjectDateRange already calls node->save and drupalGet, making below calls a problem. But actually, the fact that setProjectDateRange contains those is wrong, at least the drupalGet, the node save you can keep in there and remove from here.

As you can already see, this is why its so important to have the method names be descriptive to what they do, because then I don't have to check what it does, and you and me both could have spotted this earlier.

drishu (changes requested):
This is it for now, these are big changes so I'll need to see them implemented before I go further.

templates/content/node--oe-project--full.html.twig
donquixote: In SRB I created a solution for this that uses utility classes and does not require an empty div.
I created an issue here, https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1500

templates/content/node--oe-project--full.html.twig
donquixote: > even if I don't fill out one field, I dont get into this else 

strange. I do.

> they should take care of this situation on their side in the template

true.

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
donquixote: A day is an interval. Perhaps we could call it a half-closed interval. So `[start...end[` or `[start...end)` in math notation, or `t >= t_start && t < t_end`. The `00:00:00` moment already belongs to the next day, but every moment nanoseconds before that is from the current day.
So the most semantically correct timestamp we can send is exactly the end of the day. It is not the last moment in the interval, but the upper limit. `sup`, not `max`.
Then it is up to the js what to do with it, to interpret it as an upper limit, not as a max. (but it doesn't make any different in practice I think)

donquixote: > it passes to the next day in the calendar, so if end date is 31st it passes to 1st of next month

this would be ok or not?
the timestamp is only used to calculate the progress bar and status label.

donquixote: btw 23:59:59 is incorrect for summer / winter time switch :)

resources/js/oe_whitelabel.project_status.js
donquixote: > We don't have requirements for this, so we should not have done it in the first place

True. My mistake.

> we should leave such work to the FRONT and UX, that is their department and such things need to be requested and validated by business first in the UX side

I would agree even without the animation, because now our stuff depends on selectors coming from front-end.
It was not my decision.

resources/js/oe_whitelabel.project_status.js
donquixote: There seems to be a limit for window.setTimeout() due to 32 bit integer. If the timeout duration is higher than that number, it will happen immediately instead. The highest duration seems to be around 10 years.

Try in browser console: `window.setTimeout(console.log, 10 * 360 * 24 * 60 * 60 * 1000, '10 years passed'); window.setTimeout(console.log, 11 * 360 * 24 * 60 * 60 * 1000, '11 years passed');`.

See https://stackoverflow.com/questions/3468607/why-does-settimeout-break-for-large-millisecond-delay-values

Either way I will remove this stuff, too much risk and it was not asked for.

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
drishu: not sure about the summer saving but agreed that in this specific use case we can get away with the +1

templates/content/node--oe-project--full.html.twig
donquixote: I think this div will look strange without a comment, and somebody will remove it eventually, not knowing what change this will cause.

templates/content/node--oe-project--oe-w-content-banner.html.twig
donquixote: would be more readable if we had some custom twig filters.

tests/src/Functional/ContentProjectRenderTest.php
donquixote: Not sure. The quotes imply that this refers to existing keys or machine names somewhere. Which is not really the case.

templates/content/node--oe-project--oe-w-content-banner.html.twig
donquixote: All the parts are still needed to deal with non-array values. And I think this stuff is still better than foreach + merge.
I can change the order, I can insert line breaks.
E.g.
```
  'badges': content.oe_subject ? (content.oe_subject
    |field_value
    |default([])
    |filter(_item => _item|drupal_escape|striptags|trim is not empty)
    |map(_item => {
      'label': _item,
      'rounded_pill': true,
    }))
```

donquixote: or
```
  'badges': content.oe_subject ? (content.oe_subject
    |field_value|default([])
    |filter(_item => _item|drupal_escape|striptags|trim is not empty)
    |map(_item => {
      'label': _item,
      'rounded_pill': true,
    }))
```

donquixote: Actually we can drop the first ternary, because `|field_value` allows null passed to it. The rest will just process an empty array which is ok.

tests/src/Functional/ContentProjectRenderTest.php
```diff
@@ -291,4 +320,102 @@ protected function getStorage(string $entity_type_id): EntityStorageInterface {
...
+
+  /**
+   * Asserts start and end timestamp in project status section.
+   *
+   * @param string $begin
+   *   Expected start date string as 'Y-m-d H:i:s'.
+   * @param string $end
+   *   Expected end date string as 'Y-m-d H:i:s'.
+   */
+  protected function assertProjectStatusTimestampsAsDateStrings(string $begin, string $end): void {
```
donquixote: Btw here I use a hard-coded timezone, but in `assertTimestampAsDateString()` I have a timezone param.
I'd say the rationale is that `assertTimestampAsDateString()` is designed as a more generic function, even if used only once.

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
donquixote: True. I will squash this into the other commit where I insert the placeholder values.
So that we get it back on revert.

tests/src/Functional/ContentProjectRenderTest.php
drishu: mm, ok

tests/src/Functional/ContentProjectRenderTest.php
drishu: I dont mind

templates/content/node--oe-project--full.html.twig
drishu: I don't agreed but not a blocker for me, we can keep it

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
```diff
@@ -144,5 +160,147 @@ function _oe_whitelabel_extra_project_preprocess_inpage_nav(array &$variables):
...
+ *
+ * @param array $variables
+ *   Variables from hook_preprocess_node().
+ */
+function _oe_whitelabel_extra_project_preprocess_status_and_progress(array &$variables): void {
+  /** @var \Drupal\node\NodeInterface $node */
+  $node = $variables['node'];
+  /** @var \Drupal\datetime_range\Plugin\Field\FieldType\DateRangeItem|null $date_range_item */
+  $date_range_item = $node->get('oe_project_dates')->first();
+  if ($date_range_item === NULL || !$date_range_item->value || !$date_range_item->end_value) {
```
drishu, line 176: I don't mind the thoroughness, but just FYI the field has validation so either both are empty or both have value

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
```diff
@@ -144,5 +160,147 @@ function _oe_whitelabel_extra_project_preprocess_inpage_nav(array &$variables):
...
+  $get_timestamp = static function (string $date_string) use ($timezone): int {
+    return (new DrupalDateTime($date_string, $timezone))->getTimestamp();
+  };
+  // Project starts at the beginning of the first day at 00:00.
+  $t_start = $get_timestamp($date_range_item->value);
+  // Project ends at the end of the last day at 24:00.
+  $t_end = $get_timestamp($date_range_item->end_value . ' +1 day');
+
+  if ($t_start >= $t_end) {
+    // Cannot show a progress bar if start and end timestamp are the same.
```
drishu: comment is inaccurate but not a blocker, we are behind, so..

resources/js/oe_whitelabel.project_status.js
```diff
@@ -0,0 +1,62 @@
...
+        });
+
+        // Process the progress bar.
+        $('.progress-bar', this).each(function () {
+          const $element = $(this);
+          $element.removeClass(colorClasses);
+          $element.addClass(colorClasses[status]);
+          // Disable css transition.
+          // It looks bad for the initial setting, and afterwards it does not help.
+          $element.css('transition', false);
```
drishu: I didn't notice any issue, dont think we should fiddle with these, rather report, but not a blocker, seems to work the same with or without. If you found issue create a ticket, and once fixed, remove this "fix"

tests/src/FunctionalJavascript/ContentProjectRenderTest.php
```diff
@@ -0,0 +1,441 @@
...
+    // Set a project period that is fully in the past.
+    $this->setProjectDateRange($node, '2019-03-07', '2019-03-21');
+    $node->save();
+    // Manipulate the server time to a date even further in the past.
+    $this->getTestTimeObject()->setTime('2017-01-01');
+    $this->drupalGet($node->toUrl());
+
+    $this->assertProjectStatusTimestampsAsDateStrings('2019-03-07 00:00:00', '2019-03-22 00:00:00');
+    $this->assertProjectStatusVisible();
+    $this->assertProjectStatus('bg-dark', 'Closed');
```
drishu, line 258: the project date range is in 2019
the current date was set to 2017
2017 < 2019 that's planned
so why does it say closed ?

I think the issue is there that the datetime_testing is not used correctly, and the current date is actually always now/today, that would explain the passing tests.
See https://github.com/openeuropa/oe_theme/blob/be9ee44e38b28f13acf138fc16934bf3a7c756df/tests/src/Functional/ContentCallForTendersRenderTest.php for usage example

tests/src/FunctionalJavascript/ContentProjectRenderTest.php
```diff
@@ -0,0 +1,441 @@
...
+    $this->assertProjectProgress(100);
+
+    // Set a project period that is ongoing.
+    $this->setProjectDateRange($node, '-5 day', '+15 days');
```
drishu, line 262: I don't like this relative value, because its was not clear for me from the beginning if this is -5 days from the prev 2019 or -5 days from today 2022. Had to go to the function and check implementation, then scroll back to where I was.

tests/src/FunctionalJavascript/ContentProjectRenderTest.php
```diff
@@ -0,0 +1,441 @@
...
+    // Set a project period that is fully in the future.
+    $this->setProjectDateRange($node, '+5 days', '+12 days');
+    $node->save();
+    // Manipulate the server time to a date further in the future.
+    $this->getTestTimeObject()->setTime('+40 days');
+    $this->drupalGet($node->toUrl());
+
+    $this->assertProjectStatusVisible();
+    $this->assertProjectStatus('bg-secondary', 'Planned');
+    $this->assertProjectProgress(0);
```
drishu, line 277: all this testing is fine, but I think it would be useful, since its a js test to also have:

- a node with a fixed date range
- visit the node assert what needs to be, ex status ongoing
- change the current date to past (but not the nodes dates as now), assert status changed to planning
- change the current date to future, assert status changed to closed

tests/src/FunctionalJavascript/ContentProjectRenderTest.php
```diff
@@ -0,0 +1,441 @@
...
+  /**
+   * Gets the overridden time object to manipulate time.
+   *
+   * This is in a custom method for better static analysis.
+   *
+   * @return \Drupal\datetime_testing\TestTimeInterface
+   *   Overridden time object.
+   */
+  protected function getTestTimeObject(): TestTimeInterface {
+    $time = \Drupal::time();
```
drishu: try with 
```
$static_time = new DrupalDateTime('2020-03-02 14:00:00', DateTimeItemInterface::STORAGE_TIMEZONE);
    $this->freezeTime($static_time);
```

drishu (changes requested):

resources/js/oe_whitelabel.project_status.js
donquixote: True, no longer necessary.

tests/src/FunctionalJavascript/ContentProjectRenderTest.php
donquixote: Selenium uses the current date, the server uses the manipulated date.
The tests would fail without javascript (and they do), but they pass with javascript.
That's the point of the ->setTime().
I should add a comment.

I also wanted to add assertions for non-js situation, but did not find how to access the unmodified html.

tests/src/FunctionalJavascript/ContentProjectRenderTest.php
donquixote: True, now it is unclear because we manipulated server time.
We do need dates relative to current time, because this is what Selenium uses.

tests/src/FunctionalJavascript/ContentProjectRenderTest.php
donquixote: not possible because we cannot manipulate Selenium date (or I don't know how).

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
drishu: up 2 u

donquixote: The notorious question about defensive code in Drupal.
I don't trust these outside world expectations enough to just ignore them, but I also don't trust them enough to log an error.
So I end up with silent suppression of unexpected cases.

donquixote: Also depends if a website will change the config.
In that case we could log a warning or just ignore and not render the thing.

resources/js/oe_whitelabel.project_status.js
donquixote: The transition would only go from cached value to current value.
But it does not happen anyway because the element is hidden on page load.
So I remove this code.

drishu (dismissed):

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
```diff
@@ -144,5 +160,147 @@ function _oe_whitelabel_extra_project_preprocess_inpage_nav(array &$variables):
...
+ *
+ * @param array $variables
+ *   Variables from hook_preprocess_node().
+ */
+function _oe_whitelabel_extra_project_preprocess_contributions(array &$variables): void {
+  $legend_items = [];
+  foreach ([
+    'oe_project_budget' => 'bg-gray-400',
+    'oe_project_budget_eu' => 'bg-primary',
+  ] as $field_name => $bg_class) {
```
brummbar: I'm really not a fan of this notation, it makes it harder to read. Extract the variable from it.

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
```diff
@@ -144,5 +160,147 @@ function _oe_whitelabel_extra_project_preprocess_inpage_nav(array &$variables):
...
+  ];
+
+  /** @var \Drupal\node\NodeInterface $node */
+  $node = $variables['node'];
+  $overall_budget = $node->get('oe_project_budget')->value;
+  $eu_budget = $node->get('oe_project_budget_eu')->value;
+
+  if ($overall_budget === NULL || $eu_budget === NULL
+    || $overall_budget <= 0 || $eu_budget < 0
+    || $overall_budget < $eu_budget
```
brummbar:
```suggestion
  if (
    $overall_budget === NULL ||
    $eu_budget === NULL ||
    $overall_budget <= 0 ||
    $overall_budget < $eu_budget
```

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
```diff
@@ -144,5 +160,147 @@ function _oe_whitelabel_extra_project_preprocess_inpage_nav(array &$variables):
...
+  ) {
+    // No pie chart can be drawn with these values.
+    return;
+  }
+
+  $ratio_01 = max(0, min(1, $eu_budget / $overall_budget));
+  $percent = (int) round($ratio_01 * 10) * 10;
```
brummbar: I don't understand the complexity here in the first line. We know that overall and eu budgets are > 0, so we will have always a number between 0 and 1 out of their fraction.

brummbar: We are also missing a comment here to understand that the ratios work only 10 by 10.

resources/js/oe_whitelabel.project_status.js
```diff
@@ -0,0 +1,59 @@
...
+   */
+  Drupal.behaviors.projectStatus = {
+    attach: function (context) {
+      $('.bcl-project-status', context).once('bcl-project-status').each(function () {
+        const $element = $(this);
+        const msBegin = $element.data('start-timestamp') * 1000;
+        const msEnd = $element.data('end-timestamp') * 1000;
+        const statusLabels = $element.data('status-labels').split('|');
+        const msNow = Date.now();
+        const status = (msNow >= msBegin) + (msNow > msEnd);
```
brummbar, line 30: Small explanation that 0 is first status, 1 is second and so on.

resources/js/oe_whitelabel.project_status.js
```diff
@@ -0,0 +1,59 @@
...
+        const progress01 = Math.max(0, Math.min(1, (msNow - msBegin) / (msEnd - msBegin)));
+        const percent = Math.round(progress01 * 200) / 2;
+
+        // Process the status label.
+        $('.badge', this).each(function () {
+          const $element = $(this);
+          $element.removeClass(colorClasses);
+          $element.addClass(colorClasses[status]);
+          if (statusLabels) {
```
brummbar: How come we wrap this with an if()?

templates/content/node--oe-project--full.html.twig
```diff
@@ -4,46 +4,50 @@
...
 {% set _project_details %}
-  {{ content.group_project_details }}
+  {% if project_status_args is not empty %}
+    {{ attach_library('oe_whitelabel/project_status') }}
+    {% include '@oe-bcl/project-status' with project_status_args only %}
+  {% endif %}
+  {% if contributions_args is defined %}
+    {% include '@oe-bcl/project-status/project-contributions' with contributions_args only %}
+  {% endif %}
+  {{ content.group_project_details|filter((subgroup, key) => key|first != '#') }}
```
brummbar, line 30: What are we doing here? It looks too much cruft.

templates/content/node--oe-project--oe-w-content-banner.html.twig
```diff
@@ -0,0 +1,20 @@
...
+  'content': content.oe_teaser,
+  'image': image,
+  'attributes': create_attribute().addClass(['ps-0']),
+  'badges': content.oe_subject
+    |field_value|default([])
+    |filter(_item => _item|drupal_escape|striptags|trim is not empty)
+    |map(_item => {
+      'label': _item,
+      'rounded_pill': true,
+    }),
```
brummbar, line 19: This also looks incredibly complex, is there a need for this?

tests/src/FunctionalJavascript/ContentProjectRenderTest.php
```diff
@@ -0,0 +1,419 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp(): void {
+    parent::setUp();
+    $this->setUpSparql();
+
+    Role::load(RoleInterface::ANONYMOUS_ID)
+      ->grantPermission('bypass node access')
```
brummbar: Uhm can't we have a more normal permission? like the others below.

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
```diff
@@ -144,5 +160,147 @@ function _oe_whitelabel_extra_project_preprocess_inpage_nav(array &$variables):
...
+  $node = $variables['node'];
+  /** @var \Drupal\datetime_range\Plugin\Field\FieldType\DateRangeItem|null $date_range_item */
+  $date_range_item = $node->get('oe_project_dates')->first();
+  if ($date_range_item === NULL || !$date_range_item->value || !$date_range_item->end_value) {
+    // One of the fields is empty.
+    return;
+  }
+
+  // Dates only store the date, not the time.
+  // Use the site-wide configured timezone, not a user-specific timezone.
```
brummbar, line 182: Did we decide on this? I remember something but not precisely. What are the drawbacks of using date_default_timezone_get() and thus a user-specific timezone, if defined?

brummbar: This is actually not aligned with the dates we are rendering below. Drupal doesn't take timezone into consideration when rendering date only fields.
This makes everything harder, since we have the dilemma of "is the date referring the day for the site timezone or my day" or "why this progress is not reflecting the other website dates which are using my user timezone"
or "why the event is over while day X is still in progress".
I'd say we remove this for now completely. Let's keep the utc dates, do something on the lines of
```
  $start = $date_range_item->start_date->setTime(0, 0, 0, 0);
  $end = $date_range_item->end_date->setTime(23, 59, 59, 0);
  $t_start = $start->getTimeStamp();
  $t_end = $end->getTimeStamp();
```
And that's it. We will follow up this to see how to tackle.

resources/js/oe_whitelabel.project_status.js
```diff
@@ -0,0 +1,59 @@
...
+    attach: function (context) {
+      $('.bcl-project-status', context).once('bcl-project-status').each(function () {
+        const $element = $(this);
+        const msBegin = $element.data('start-timestamp') * 1000;
+        const msEnd = $element.data('end-timestamp') * 1000;
+        const statusLabels = $element.data('status-labels').split('|');
+        const msNow = Date.now();
+        const status = (msNow >= msBegin) + (msNow > msEnd);
+        const progress01 = Math.max(0, Math.min(1, (msNow - msBegin) / (msEnd - msBegin)));
+        const percent = Math.round(progress01 * 200) / 2;
```
brummbar: How come this 2 factor? Can't it be * 100 without the / 2? We need documentation to understand the reasons.

resources/js/oe_whitelabel.project_status.js
```diff
@@ -0,0 +1,59 @@
...
+  Drupal.behaviors.projectStatus = {
+    attach: function (context) {
+      $('.bcl-project-status', context).once('bcl-project-status').each(function () {
+        const $element = $(this);
+        const msBegin = $element.data('start-timestamp') * 1000;
+        const msEnd = $element.data('end-timestamp') * 1000;
+        const statusLabels = $element.data('status-labels').split('|');
+        const msNow = Date.now();
+        const status = (msNow >= msBegin) + (msNow > msEnd);
+        const progress01 = Math.max(0, Math.min(1, (msNow - msBegin) / (msEnd - msBegin)));
```
brummbar, line 32: Drop a line here. UTC timestamps can be negative if before 1970, so we need to make sure that the current progress stays between 0 and 1(00)%.

brummbar (changes requested):

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
brummbar: I am very conflicted about this, also DGCOMM is planning to add timezones to this field so it can be made clear.

resources/js/oe_whitelabel.project_status.js
donquixote: This has nothing to do with 1970.
The raw progress will be < 0 if the event is in the future. It will be > 1 if the event is in the past.
This would be the same even if the current date and the event are before 1970.

The style and attribute values for the progress bar has to in the range of 0..100 so we must clamp it.

resources/js/oe_whitelabel.project_status.js
donquixote: Very made-up and artificial, perhaps silly.
Arbitrary long precision looks nasty if the aria attributes are shown somewhere or read in a screen reader.
With 200 steps we are likely to be below pixel width on most displays.
But 100 steps would also work, and raise fewer questions.

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
donquixote: The scenario is this:
- Site timezone is Brussels,BE.
- A logged-in user has a custom timezone in Sydney,Australia. Depending on site settings, the php timezone will be Sydney, not Brussels, when the request is processed.
- An anonymous user from Sydney has their own local time, but Drupal does not know about it. The php timezone will be Brussels, even though the anon user is in Sydney.

The events only store a date string without time.
We assume that the events happen in Brussels, so they begin and end at 00:00 in Brussels timezone.

With the current calculation, for both of the users from Sydney, the event would show up as "Planned" in the morning, and "Ongoing" in the evening.
If a user from California, one from Brussels, and one from Sydney, logged in or not, have a phone call while visiting the website, they will see the same project status.

If we use user timezone (which will happen if we use php timezone), then a user from California would see a different event status than a user from Sydney, and the anonymous visitor would see a different status than the logged-in user from Sydney.

> This is actually not aligned with the dates we are rendering below.

Indeed, the status would not conform with the date a user would see in their local clock.
However, this is already the case for anon visitors.

> I'd say we remove this for now completely. Let's keep the utc dates, do something on the lines of

I think your code will just use php timezone, which can be the configured timezone or the user timezone.
But let's say we use UTC - so an event starts and ends at 00:00 UTC.
I don't think this is what clients would want. I think most of them will want 00:00 Brussels time. And for those who don't, UTC is going to be just as wrong as Brussels.

I think 00:00 Brussels (or configured site default timezone) is the best we can do with the given information, every alternative is worse.

> I am very conflicted about this, also DGCOMM is planning to add timezones to this field so it can be made clear.

Once this happens, we can review this logic :)

templates/content/node--oe-project--oe-w-content-banner.html.twig
donquixote: I don't like it, but I also don't like the for loops with array merge to add new items.

So:
- `|field_value` is like `element_children()`, but unfortunately it returns NULL sometimes. Both `|filter` and `|map` below require an array as parameter, so we need `|default([])`.
- The `|filter()` cleans out field items that are empty when rendered. Perhaps we can skip this step.
- `|map` does the actual array structure conversion.

I will test if empty leftover field items are a thing. If not, I will drop `|filter()`.

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
donquixote: There is a discussion with a proposed format for this, not yet accepted though.
https://www.drupal.org/project/coding_standards/issues/1539712
The direction is to start each new line with the respective operator.
I support this general direction. I like to see lines starting with the operator.

In this case I cheated a bit to have similar conditions on the same line. But I am ok to split these.

In my own projects I sometimes did this to avoid the first line looking different from the other lines, hoping that the "compiler" will optimize it away. But it is unorthodox and I don't suggest doing it:

```php
if (false
  || $condition1
  || $condition2
) {..}
```

Do we have in-house conventions that the operator should be at the end of a line?

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
donquixote: Ok if you say so :)

templates/content/node--oe-project--full.html.twig
donquixote: This is element_children() for poor people.
We want to render each element in the group, but not the group wrapper. We want the wrapper around the entire project details including the status/progress and the contributions.

I would much prefer to have custom twig filters in oe_bootstrap_theme or elsewhere, see https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1447.
Or a custom field group formatter "bare" that does not render any wrappers.

Alternatively I could render each item in the group separately.

resources/js/oe_whitelabel.project_status.js
donquixote: Probably leftover. There is no reason now why it would be empty.

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
brummbar: It's literally point 5 of what you linked here https://www.drupal.org/project/coding_standards/issues/1539712 😄

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
brummbar: Uhm not really, thanks for pointing that issue. I usually use mine above because it aligns all the text, allows to comment out lines without breaking, and it's similar to multi-line function invocations.
Let's keep yours, but split each || in a new line.
Also note that my suggestion drops the check for $eu_budget < 0 as we are checking that for overall budget already, so if overall budget is major than 0 and eu budget is major than overall, it implies that eu is also major than 0. Am I wrong?

templates/content/node--oe-project--full.html.twig
brummbar: Not good to skip wrappers, because cache information can be there in some scenarios. Much better to use a bare template.

donquixote: I think for field groups all the relevant cache tags sit in the fields, so it would be safe like this. But I could be wrong :)

But it makes me wonder if the twig filters `|field_value` and `|field_label` from https://www.drupal.org/project/twig_field_value are safe for caching.
The module page has a cache warning about field_raw and field_target_entity but not about field_value and field_label.

I need to learn a bit more how cache tags are processed during the render process.

For the bare template: I don't like adding more and more specific templates. But something like this could work (this one is for a field, not a group):
```twig
content.body|merge({'#theme': 'field__bare'})
```

templates/content/node--oe-project--oe-w-content-banner.html.twig
brummbar: I think we need to fix in general how we do this, because again cache information attached to the "empty" values can be lost.
Let's keep it for now but revisit later.

resources/js/oe_whitelabel.project_status.js
brummbar: Uhm I still don't get it much, since we are rounding anyway... Let's remove it indeed.

resources/js/oe_whitelabel.project_status.js
brummbar: Yep, 1970 is just one of the scenarios. The important thing is that we need to keep between 0 and 100. Write that in the comment.

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
brummbar: I think what you suggested is very much in line with basically saying "The date entered is in the website timezone, so the project is in progress based on that timezone". Which I like, it was one of the "I'm confused" scenarios.
Let's see @drishu opinion, for me is a keep.

drishu: tested again, I agree its good, the main thing that gives me confidence is that the javascript Date.now() is the same no matter the users location or system settings, so it will be the same calculation

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
```diff
@@ -144,5 +160,147 @@ function _oe_whitelabel_extra_project_preprocess_inpage_nav(array &$variables):
...
+  /* @see \Drupal\system\TimeZoneResolver::getTimeZone() */
+  /** @var string $timezone */
+  $timezone = \Drupal::config('system.date')->get('timezone.default') ?? 'UTC';
```
brummbar: We forgot to add this config cache information to the render array.

brummbar (changes requested):

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
donquixote:
```php
$eu_budget = -10;
$overall_budget = 50;

if ($overall_budget === NULL  # false
    || $eu_budget === NULL  # false
    || $overall_budget <= 0  # false
    || $eu_budget < 0  # true
    || $overall_budget < $eu_budget  # false
  ) {
    // No pie chart can be drawn with these values.
    return;
  }
```
The condition for `$eu_budget < 0` is the only one that is `true`.
Without the condition, the chart is shown. With the condition, the chart is not shown.

(It is another question whether we should prevent this with validation. But I don't like to rely too much on outside constraints for my code.)

resources/js/oe_whitelabel.project_status.js
donquixote: I am now rounding to 1%.
But to explain my idea, my feeling was `37.5%` looks ok but `37.1%` already looks too fine-grained.
Now the new rationale is `37%` looks ok but `37.5%` looks too fine-grained.

These attributes are for accesibility, but I can only speculate how they may be perceived in practice.

templates/content/node--oe-project--full.html.twig
donquixote: I think we are safe here, as said. I will leave this one unchanged for now.

templates/content/node--oe-project--oe-w-content-banner.html.twig
donquixote: Actually `|drupal_escape` triggers rendering.
On the other hand, `|field_value` does not.
But ok let's keep. If we have to change this here, there are plenty of other places where we do something that will have the same effect.

tests/src/FunctionalJavascript/ContentProjectRenderTest.php
donquixote: Turns out we don't need a perm at all to see the node.

donquixote: Now I notice where this is coming from:
`Drupal\Tests\oe_whitelabel\Kernel\ContentRenderTestBase::setUp()`.
We also don't need it here.

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.module
brummbar: Yup you are correct indeed.

templates/content/node--oe-project--full.html.twig
brummbar: Can't we just change the template suggestion in the node full preprocess?

templates/content/node--oe-project--oe-w-content-banner.html.twig
brummbar: "Hey this can lead to a bug, but we did it a bit everywhere so let's ignore it". Sounds like a flawed approach to me :)
But anyway I'll make a follow-up to replace field_value with our own filter as we discussed.

donquixote: I did some digging.
Indeed some formatters add cache info directly to the `$elements` array, which is lost if we pick the children.
This will be lost with `|field_value`, but also when using a `for` loop to pick field items.

Let's indeed fix this with a new issue.

templates/content/node--oe-project--full.html.twig
donquixote: I did some debugging, and found that field group elements don't contain `'#cache'` entries.

Field group elements are populated in a preprocess on the node/entity itself, which simply moves fields around.

One thing we _could_ do is this:

```twig
{{ content.group_project_details|merge({'#theme_wrappers': []}) }}
```

This way the wrapper div would not be rendered, but the group element would still go through the render pipeline.
Perhaps in addition we also want to remove `'#attached'` data.
Obviously any processing applied by the field group formatter in the preprocess phase already happened before it hits the template.

A cleaner option would be to have a special field group formatter that does not apply any wrappers. See https://www.drupal.org/project/field_group/issues/3279137.
It seems straightforward, we already have one field group formatter in oe_whitelabel.

I am not sure if I want to do this in the scope of this issue.

brummbar (approved):

drishu (approved):
Changes are ok, approved

PR #133, opened 2022-05-05
OEL-1528: Fix whitelabel issues on template, composer and dependencies.

Maxfire:


tests/src/Functional/ContentEventRenderTest.php
```diff
@@ -24,6 +24,7 @@ class ContentEventRenderTest extends WhitelabelBrowserTestBase {
    */
   public static $modules = [
     'block',
+    'oe_corporate_blocks',
```
brummbar: This is a content type test, and we need to manually enable corporate blocks. This is exactly what we should try to avoid and fix. Please revert.

modules/oe_whitelabel_helper/oe_whitelabel_helper.info.yml
```diff
@@ -6,8 +6,8 @@ core_version_requirement: ^9.2
 
 dependencies:
   - drupal:node
+  - drupal:twig_field_value
   - oe_bootstrap_theme:oe_bootstrap_theme_helper
-  - openeuropa:oe_corporate_blocks
```
brummbar, line 10: This was correct, why did we remove it?
We need to clean the list of modules to enable manually in the runner, not fix the dependencies here.
Only our own modules, or modules that are not supposed to mandatory, should be enabled manually (which for us is the runner.yml).

brummbar (changes requested):
I don't remember anymore the issue with dependencies that originated the ticket, but some of the things done are wrong.

brummbar (dismissed):

drishu (approved):

brummbar (approved):

PR #136, opened 2022-05-06
OEL-1367: New Look & Feel for search form in search page and remove theming from block config form.

escuriola:


modules/oe_whitelabel_search/config/schema/oe_whitelabel_search.schema.yml
```diff
@@ -9,29 +9,9 @@ block.settings.whitelabel_search_block:
         action:
           type: string
           label: Action
-    input:
-      type: mapping
-      label: Input
-      mapping:
-        name:
```
drishu: I agree with removing most of this, but I'd keep the name (search_api_fulltext is horrible) and I'd keep the placeholder

modules/oe_whitelabel_search/config/schema/oe_whitelabel_search.schema.yml
```diff
@@ -9,29 +9,9 @@ block.settings.whitelabel_search_block:
...
-          type: string
-          label: Placeholder
-    button:
-      type: mapping
-      label: Button
-      mapping:
-        classes:
+        layout:
           type: string
-          label: Classes
```
drishu, line 34: now that we display a label for button I think its a good idea to add a button label to the schema and config (UCPKN also requested this)

modules/oe_whitelabel_search/oe_whitelabel_search.module
```diff
@@ -0,0 +1,41 @@
...
+ * OE Whitelabel search.
+ */
+
+declare(strict_types=1);
+
+/**
+ * Implements hook_theme_suggestions_HOOK_alter().
+ */
+function oe_whitelabel_search_theme_suggestions_form_alter(array &$suggestions, array $variables) {
+  if (!empty($variables['element']['#form_id'])) {
```
drishu: keep the format from below, where you exit early with a return in the if (then obviously you need to inverse the logic of the statement)

modules/oe_whitelabel_search/oe_whitelabel_search.module
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  $suggestions[] = 'form_element__' . $variables['element']['#form_id'];
+}
+
+/**
+ * Implements hook_theme_suggestions_HOOK_alter().
+ */
+function oe_whitelabel_search_theme_suggestions_input_alter(array &$suggestions, array $variables) {
+
```
drishu: this newline is not needed

modules/oe_whitelabel_search/oe_whitelabel_search.module
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * @file
+ * OE Whitelabel search.
+ */
+
+declare(strict_types=1);
+
+/**
+ * Implements hook_theme_suggestions_HOOK_alter().
+ */
```
drishu: would be a nice addition if we explain in the function comment, in short, what the suggestion is about, what's with this region thingy. Go on and explain the purpose also in the other 2 functions below.

modules/oe_whitelabel_search/oe_whitelabel_search.module
```diff
@@ -0,0 +1,41 @@
...
+function oe_whitelabel_search_theme_suggestions_form_alter(array &$suggestions, array $variables) {
```
drishu:
```suggestion
function oe_whitelabel_search_theme_suggestions_form_alter(array &$suggestions, array $variables): void {
```

modules/oe_whitelabel_search/oe_whitelabel_search.module
```diff
@@ -0,0 +1,41 @@
...
+  if (!empty($variables['element']['#form_id'])) {
+    $suggestions[] = 'form__' . $variables['element']['#region'];
+    $suggestions[] = 'form__' . $variables['element']['#region'] . '__' . $variables['element']['#form_id'];
+  }
+}
+
+/**
+ * Implements hook_theme_suggestions_HOOK_alter().
+ */
+function oe_whitelabel_search_theme_suggestions_form_element_alter(array &$suggestions, array $variables) {
```
drishu:
```suggestion
function oe_whitelabel_search_theme_suggestions_form_element_alter(array &$suggestions, array $variables): void {
```

modules/oe_whitelabel_search/oe_whitelabel_search.module
```diff
@@ -0,0 +1,41 @@
...
+
+  $suggestions[] = 'form_element__' . $variables['element']['#form_id'];
+}
+
+/**
+ * Implements hook_theme_suggestions_HOOK_alter().
+ */
+function oe_whitelabel_search_theme_suggestions_input_alter(array &$suggestions, array $variables) {
+
+  if (!isset($variables['element']['#form_id']) || $variables['element']['#form_id'] !== 'oe_whitelabel_search_form') {
```
drishu:
```suggestion
  if (!isset($variables['element']['#form_id']) || $variables['element']['#form_id'] !== 'oe_whitelabel_search_form'):void {
```

modules/oe_whitelabel_search/oe_whitelabel_search.module
```diff
@@ -0,0 +1,41 @@
...
+    return;
+  }
+
+  $suggestions[] = 'input__textfield__' . $variables['element']['#region'] . '__' . $variables['element']['#form_id'];
```
drishu: It's a good practice to add also the more general suggestion above the more specific one:
$suggestions[] = 'input__textfield__' . $variables['element']['#region']

modules/oe_whitelabel_search/src/Form/SearchForm.php
```diff
@@ -46,57 +46,27 @@ public static function create(ContainerInterface $container) {
...
-        'class' => [
-          $config['input']['classes'],
-          'rounded-0',
-          'rounded-start',
-        ],
-      ],
-      '#default_value' => $input_value,
+      '#form_id' => $this->getFormId(),
+      '#region' => $config['form']['layout'],
+      '#default_value' => $this->getRequest()->get('search_api_fulltext'),
```
drishu: I'd keep this one configurable, we don't know what other filters and facets the user might have on the page and 'search_api_fulltext' is quite general for search_api, if we have it configurable then site builders can use naming that avoids conflict.

modules/oe_whitelabel_search/src/Plugin/Block/SearchBlock.php
```diff
@@ -99,52 +98,17 @@ public function blockForm($form, FormStateInterface $form_state): array {
...
-      '#tree' => TRUE,
-      '#description' => $this->t('Fill in the settings of the Button field.'),
-    ];
-    $form['button']['button_classes'] = [
-      '#type' => 'textfield',
-      '#title' => $this->t('Button classes'),
-      '#description' => $this->t('Add space-separated classes that will be added to the button.'),
-      '#default_value' => $config['button']['classes'],
+
+    $form['layout'] = [
```
drishu: I like the approach of adding the region to the form so that we can target it with templates, BUT, I don't like the namings. If it's regions that are the options then lets call it region.

Also, **since we already have the region in the page**, in a subform in the parentForm -> region -> value let's try to make this a hidden field, and set its value on validation with the value of the blocks region selected

modules/oe_whitelabel_search/src/Form/SearchForm.php
```diff
@@ -46,57 +46,27 @@ public static function create(ContainerInterface $container) {
...
       '#required' => TRUE,
     ];
 
     $form['submit'] = [
+      '#type' => 'pattern',
```
drishu: let's make this a regular submit button and in the template we override with the pattern, but pass on the attributes

templates/overrides/search/form--header--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,48 @@
...
+{% set label %}
+  <span class="d-none d-lg-inline-block">{{ 'Search'|t }}</span>
+{% endset %}
+{% set submit_button = pattern('button', {
+  'variant': 'primary',
+  'icon': 'search',
+  'label': label,
+  'type': 'submit',
+  'icon_position': 'before',
+  'attributes': {
```
drishu: in the form we make it a regular drupal form submit, in the override we should merge the attributes

templates/overrides/search/form--header--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,48 @@
+{#
+/**
+ * @file
+ * Template for the header search form element .
```
drishu: reformulate this so its clear that its the header region

templates/overrides/search/form--navigation-right--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,36 @@
+{#
```
drishu, line 1: same comments as for the other template

templates/overrides/search/form-element--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,28 @@
+{#
+/**
+ * @file
+ * Theme implementation for the search form element.
+ *
```
drishu, line 5: its hard for me to tell what is different without comparing, could you add one line of description explaining what is done here pls

templates/overrides/search/form--header--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,48 @@
...
+  <div class="container">
+    <div class="row">
+      <div class="col-12 col-lg-6 offset-lg-3">
+        <form {{ attributes.addClass('bcl-search-form', 'submittable') }}>
+          <div class="bcl-search-form__group">
+            {{ element.search_input }}
+            {{ submit }}
+            {{ element.form_build_id }}
+            {{ element.form_token }}
+            {{ element.form_id }}
```
drishu: I don't like this, think you where on the right track with the suggestions for input__textfield, do the same for input__submit
Then you override the submit in its template and here you call children

config/optional/block.block.oe_whitelabel_search_form.yml
```diff
@@ -18,13 +18,7 @@ settings:
...
   form:
     action: '#'
-  input:
-    name: text
-    label: Search
-    classes: ''
-    placeholder: Search
-  button:
-    classes: ''
+    layout: navigation_right
```
drishu: if we alter this then we need a update path

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
```diff
@@ -84,7 +84,41 @@ protected function setUp(): void {
...
+    $block_entity_storage = $this->container
+      ->get('entity_type.manager')
+      ->getStorage('block');
+    $entity = $block_entity_storage->load('oe_whitelabel_search_form');
+    $builder = \Drupal::entityTypeManager()->getViewBuilder('block');
+    $build = $builder->view($entity, 'block');
+    $render = $this->container->get('renderer')->renderRoot($build);
+    $crawler = new Crawler($render->__toString());
+
+    // Assert the form rendering.
```
drishu, line 97: think its worth asserting the wrappers that where added bg-lighter -> container -> row -> col-12

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
```diff
@@ -123,32 +149,30 @@ public function testBlockRendering(): void {
...
-    $label = $wrapper->filter('label');
-    $this->assertSame('Search', $label->text());
-    $classes = 'visually-hidden js-form-required form-required form-label';
-    $this->assertSame($classes, $label->attr('class'));
     // Assert the button and icon rendering.
-    $button = $crawler->filter('.button-test-class');
+    $button = $crawler->filter('button#submit');
     $this->assertCount(1, $button);
-    $classes = 'border-start-0 rounded-0 rounded-end d-flex btn btn-light btn-md py-2 button-test-class btn btn-light';
+    $classes = 'bcl-search-form__submit px-3 btn btn-primary btn-md';
```
drishu: lets remove classes that handle spacing from the asserts (like px-3), do this evaluation in the entire test

drishu (changes requested):

modules/oe_whitelabel_search/config/schema/oe_whitelabel_search.schema.yml
escuriola: Agree with the name, but placeholder can be overriden in the input--textfield templates, for this reason I didn't add to the schema, are you sure?

modules/oe_whitelabel_search/config/schema/oe_whitelabel_search.schema.yml
escuriola: Same, the buttons can be overriden in the form templates

modules/oe_whitelabel_search/config/schema/oe_whitelabel_search.schema.yml
drishu: ok

modules/oe_whitelabel_search/oe_whitelabel_search.post_update.php
```diff
@@ -0,0 +1,20 @@
...
+use Drupal\oe_bootstrap_theme\ConfigImporter;
+
+/**
+ * Set the updated values to the search form block.
+ */
+function oe_whitelabel_search_post_update_00001(&$sandbox) {
+  $configs = [
+    'block.block.oe_whitelabel_search_form',
+  ];
+  ConfigImporter::importMultiple('module', 'oe_whitelabel_search', '/config/post_updates/00001', $configs);
```
drishu: this is not the way to do it, you are overriding stuff you should not touch (like the visibility settings for instance)

templates/overrides/search/form--header--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,23 @@
+{#
+/**
+ * @file
+ * Search form template for the header region.
+ *
+ * @see ./core/modules/system/templates/form.html.twig
+ */
+#}
+
```
drishu: remove the newlines

templates/overrides/search/form--navigation-right--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,32 @@
...
+      'px-3',
+    ],
+  },
+})%}
+
+{% set submit = element.submit|merge({
+  '#markup': submit_button|render,
+  })
+%}
+
```
drishu: since u have a separate template now all this is not needed

templates/overrides/search/form-element--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,30 @@
...
+ * Remove form element div wrapper to keep just the search form fields.
```
drishu: actually I see label, prefix, suffix and description removed

modules/oe_whitelabel_search/src/Form/SearchForm.php
```diff
@@ -48,55 +48,30 @@ public function buildForm(array $form, FormStateInterface $form_state, array $co
     $form_state->set('oe_whitelabel_search_config', $config);
     $input_value = '';
 
+    $form['#region'] = $config['form']['region'];
     if (!empty($config['input']['name'])) {
```
drishu: looking at it, the field is required, so wont be empty the way we use it, still the form could be injected so I think its a better idea to return here if empty, since the name should not be empty ever

modules/oe_whitelabel_search/src/Form/SearchForm.php
```diff
@@ -48,55 +48,30 @@ public function buildForm(array $form, FormStateInterface $form_state, array $co
...
+      '#region' => $config['form']['region'],
       '#default_value' => $input_value,
       '#required' => TRUE,
     ];
 
     $form['submit'] = [
+      '#type' => 'submit',
       '#input' => TRUE,
       '#is_button' => TRUE,
       '#executes_submit_callback' => TRUE,
```
drishu:
```suggestion
      '#value' => $this->t('Search'),
```
a simple submit button doesn't need #input => TRUE etc

modules/oe_whitelabel_search/oe_whitelabel_search.module
```diff
@@ -0,0 +1,48 @@
...
+ * Implements hook_theme_suggestions_HOOK_alter().
+ *
+ * Add suggestions based on the block region where the form is placed.
+ */
+function oe_whitelabel_search_theme_suggestions_form_alter(array &$suggestions, array $variables): void {
+  if (empty($variables['element']['#form_id'])) {
+    return;
+  }
+
+  $suggestions[] = 'form__' . $variables['element']['#region'];
```
drishu: Undefined array key "#region"

modules/oe_whitelabel_search/oe_whitelabel_search.module
```diff
@@ -0,0 +1,48 @@
...
+
+declare(strict_types=1);
+
+/**
+ * Implements hook_theme_suggestions_HOOK_alter().
+ *
+ * Add suggestions based on the block region where the form is placed.
+ */
+function oe_whitelabel_search_theme_suggestions_form_alter(array &$suggestions, array $variables): void {
+  if (empty($variables['element']['#form_id'])) {
```
drishu: This is oe_whitelabel_search so lets add suggestions only to the stuff we need, if we will need it anywhere else we can move them out of this module.
Do the condition like you did for oe_whitelabel_search_theme_suggestions_input_alter, only for the templates we need

modules/oe_whitelabel_search/src/Plugin/Block/SearchBlock.php
```diff
@@ -249,17 +197,11 @@ public function blockSubmit($form, FormStateInterface $form_state): void {
...
-      ]));
-    }
-
-    if ($values['button']['button_classes'] !== Html::cleanCssIdentifier($values['button']['button_classes'])) {
-      $form_state->setErrorByName('button][button_classes', $this->t('Field "@field_name" does not contain a valid css class.', [
-        '@field_name' => $form['button']['button_classes']['#title'],
-      ]));
-    }
+    $completeForm = $form_state->getCompleteForm();
+    $this->setConfigurationValue('form', [
```
drishu: here you should only set the value for region in the form state

modules/oe_whitelabel_search/src/Plugin/Block/SearchBlock.php
```diff
@@ -221,21 +184,6 @@ public function blockForm($form, FormStateInterface $form_state): array {
    * {@inheritdoc}
    */
   public function blockSubmit($form, FormStateInterface $form_state): void {
-    $values = $form_state->getValues();
-    $this->setConfigurationValue('form', [
```
drishu, line 225: you should keep this any only add region

modules/oe_whitelabel_search/oe_whitelabel_search.module
```diff
@@ -0,0 +1,48 @@
+<?php
+
+/**
+ * @file
+ * OE Whitelabel search.
```
drishu: Module file used for theming the search feature.

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
```diff
@@ -84,7 +84,46 @@ protected function setUp(): void {
...
+    $entity = $block_entity_storage->load('oe_whitelabel_search_form');
+    $builder = \Drupal::entityTypeManager()->getViewBuilder('block');
+    $build = $builder->view($entity, 'block');
+    $render = $this->container->get('renderer')->renderRoot($build);
+    $crawler = new Crawler($render->__toString());
+
+    // Assert the form rendering.
+    $block = $crawler->filter('div.oe-whitelabel-search-form');
+    $this->assertCount(1, $block);
+    $form = $block->filter('form#oe-whitelabel-search-form');
```
drishu: If you'r so specific then the count below wont fail, just grab 'form'

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
```diff
@@ -84,7 +84,46 @@ protected function setUp(): void {
...
+    $this->assertStringContainsString('border-start-0', $button->attr('class'));
+    $this->assertStringContainsString('btn', $button->attr('class'));
+    $this->assertStringContainsString('btn-md', $button->attr('class'));
+    $this->assertStringContainsString('btn-light', $button->attr('class'));
+    $icon = $button->filter('.bi.icon--fluid');
+    $this->assertCount(1, $icon);
+  }
+
+  /**
+   * Tests the rendering of the whitelabel search block.
```
drishu: maybe differentiate the messages a bit

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
```diff
@@ -122,33 +156,39 @@ public function testBlockRendering(): void {
...
 
+    // Assert header form wrappers.
+    $wrapper = $crawler->filter(
+      'div.bg-lighter > div.container > div.row > div.col-12.col-lg-6.offset-lg-3'
+    );
+    $this->assertCount(1, $wrapper);
+
     // Assert the form rendering.
-    $block = $crawler->filter('#block-whitelabel-search-block');
+    $block = $crawler->filter('div.oe-whitelabel-search-form');
```
drishu: assert this wrapper is within this block

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
```diff
@@ -122,33 +156,39 @@ public function testBlockRendering(): void {
...
     $this->assertCount(1, $block);
-    $form = $block->filter('#oe-whitelabel-search-form');
+    $form = $block->filter('form#oe-whitelabel-search-form');
```
drishu: and then the form should reside within the wrapper

templates/overrides/search/form--navigation-right--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,32 @@
...
+<form {{ attributes.addClass('d-flex', 'mt-3', 'mt-lg-0') }}>
```
drishu: Also, since above is not needed all you do it add these classes, you dont need to redeclare the file for that, just extend
```
{% extends "form.html.twig" %}
{% set attributes = attributes.addClass('d-flex', 'mt-3', 'mt-lg-0') %}
```

templates/overrides/search/input--textfield--header--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,17 @@
...
+ * @see ./core/modules/system/templates/input.html.twig
+ */
+#}
+{%
+  set classes = [
+  'form-control',
+  'bcl-search-form__input',
+  attributes.hasClass('error') ? 'is-invalid',
+]
+%}
```
drishu: see if u can just extend since you only work with the attributes

templates/overrides/search/input--textfield--navigation-right--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,20 @@
...
+  set classes = [
+  'form-control',
+  'border-start-0',
+  'rounded-0',
+  'rounded-start',
+  'form-control',
+  attributes.hasClass('error') ? 'is-invalid',
+]
+%}
+<input{{ attributes.addClass(classes).setAttribute('placeholder', 'Search'|t).removeClass('form-text') }} />{{ children }}
```
drishu: see if u can just extend since you only work with the attributes.
For the template where you extend, I dont think the '@see' makes any more sense

drishu (changes requested):
I have been thinking of what you said @escuriola about 'search button label' but am not convinced if the label of the form element should be considered theming or configuration, I think the latter, but will discuss it

modules/oe_whitelabel_search/src/Form/SearchForm.php
```diff
@@ -45,58 +45,38 @@ public static function create(ContainerInterface $container) {
    * {@inheritdoc}
    */
   public function buildForm(array $form, FormStateInterface $form_state, array $config = NULL): array {
-    $form_state->set('oe_whitelabel_search_config', $config);
+    if (empty($config['input']['name'])) {
+      return [];
+    }
     $input_value = '';
-
     if (!empty($config['input']['name'])) {
```
drishu: yes, but now that we check above we can drop this if wrapper

templates/overrides/search/input--textfield--header--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,16 @@
...
+ */
+#}
+{% extends "input--textfield.html.twig" %}
+{% set attributes = attributes
+  .addClass([
+    'form-control',
+    'bcl-search-form__input',
+    attributes.hasClass('error') ? 'is-invalid',
+  ])
+  .setAttribute('placeholder', 'Search'|t)
```
drishu: this should use the value set in settings[input][input_placeholder]

templates/overrides/search/input--textfield--navigation-right--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,18 @@
...
+{% extends "input--textfield.html.twig" %}
+{% set attributes = attributes
+  .addClass([
+    'form-control',
+    'border-start-0',
+    'rounded-0',
+    'rounded-start',
+    attributes.hasClass('error') ? 'is-invalid',
+  ])
+  .setAttribute('placeholder', 'Search'|t)
```
drishu: this should use the value set in settings[input][input_placeholder]

drishu (changes requested):

modules/oe_whitelabel_search/src/Form/SearchForm.php
```diff
@@ -45,58 +45,34 @@ public static function create(ContainerInterface $container) {
...
+    $form['#region'] = $config['form']['region'];
 
     $form['search_input'] = [
-      '#prefix' => '<div class="bcl-search-form__group w-100">',
-      '#suffix' => '</div>',
       '#type' => 'textfield',
       '#title' => $config['input']['label'],
       '#title_display' => 'invisible',
       '#size' => 20,
       '#margin_class' => 'mb-0',
```
drishu: can you double check if this is used, if it gets added to any of our search_input(s)? 
I remember this `margin_class` is a thing we inherited from the agency theme but I dont like it much.
If it this class is present in netlify, then we should add it in our templates not here; if it is not in netlify, then we can just remove it.

modules/oe_whitelabel_search/oe_whitelabel_search.post_update.php
```diff
@@ -0,0 +1,21 @@
...
+ * @file
+ * OpenEuropa Whitelabel Search post updates.
+ */
+
+declare(strict_types=1);
+
+use Drupal\block\Entity\Block;
+
+/**
+ * Set the updated values to the search form block.
```
drishu: Let's be a bit more specific, explaining in one short sentence what it does and why
Something like "Add region in form settings for template suggestions."
Dont know if everybody will like it but I think its better than now.

templates/overrides/search/input--textfield--header--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,15 @@
...
+#}
+{% extends "input--textfield.html.twig" %}
+{% set attributes = attributes
+  .addClass([
+    'form-control',
+    'bcl-search-form__input',
+    attributes.hasClass('error') ? 'is-invalid',
+  ])
+  .removeClass('form-text') %}
+
```
drishu: remove extra newline

templates/overrides/search/input--textfield--navigation-right--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,17 @@
...
+{% set attributes = attributes
+  .addClass([
+    'form-control',
+    'border-start-0',
+    'rounded-0',
+    'rounded-start',
+    attributes.hasClass('error') ? 'is-invalid',
+  ])
+  .removeClass('form-text') %}
+
```
drishu: remove extra newline

templates/overrides/search/input--submit--header--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,27 @@
...
+  {% endset %}
+{% endif %}
+{{ pattern('button', {
+  'variant': 'primary',
+  'icon': 'search',
+  'label': label,
+  'type': 'submit',
+  'icon_position': 'before',
+  'attributes': attributes
+    .setAttribute('id', 'submit')
```
drishu: Am just curios why we need it? we also need to be careful as other elements on the page could get the same id in subsite projects

drishu (changes requested):

modules/oe_whitelabel_search/src/Form/SearchForm.php
escuriola: Yes, is used, because if margin_class is not declared, then the template add its own margin:
image

drishu: Yes, I know of that, the question was is it being printed to the element in the page "if it gets added to any of our search_input(s)?".
I looked into it, found that no, it does not get printed, so it is not used, because we remove that wrapper in form-element--oe-whitelabel-search-form.html.twig, so lets remove it.

templates/overrides/search/input--submit--header--oe-whitelabel-search-form.html.twig
drishu: Answer was: "I didn't add, it just was in the SearchForm.php before and I moved when I created the template. I tested without it and works, so I remove them in my last commit"

modules/oe_whitelabel_search/src/Form/SearchForm.php
escuriola: You're right. Sorry

drishu (dismissed):

templates/overrides/search/form-element--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,30 @@
...
+
+{{ children }}
+
+{% if errors %}
+  <div{{ attributes.addClass(classes) }}>
+    <div class="form-item--error-message invalid-feedback d-block">
+      {{ errors }}
+    </div>
+  </div>
+{% endif %}
```
donquixote, line 17: How can we test this part with `errors`?
I tried different tricks to submit invalid values, but the errors section never shows up.

The section seems weird to me, because all the classes would only make sense for a complete form element, not just for a section with errors.

modules/oe_whitelabel_search/src/Form/SearchForm.php
```diff
@@ -45,58 +45,33 @@ public static function create(ContainerInterface $container) {
...
-      '#prefix' => '<div class="bcl-search-form__group w-100">',
-      '#suffix' => '</div>',
       '#type' => 'textfield',
       '#title' => $config['input']['label'],
       '#title_display' => 'invisible',
       '#size' => 20,
-      '#margin_class' => 'mb-0',
+      '#form_id' => $this->getFormId(),
+      '#region' => $config['form']['region'],
+      '#default_value' => $this->getRequest()->get($config['input']['name']),
```
donquixote, line 69: Interesting observation: On http://web:8080/build/en?search_api_fulltext=imputo the field is pre-filled, even though we are not on the actual search page. I guess that's ok..
(this is because we use the same logic for both variants of this form)

modules/oe_whitelabel_search/oe_whitelabel_search.module
```diff
@@ -2,7 +2,7 @@
 
 /**
  * @file
- * OE Whitelabel Search Module.
+ * Module file used for theming the search feature.
```
donquixote, line 5: For my taste we should agree on a copyable comment we can put in every *.module file. The only party interested in these `@file` docs is phpcs..
But this is not a blocker, ok to leave it as is.

modules/oe_whitelabel_search/oe_whitelabel_search.module
```diff
@@ -49,3 +49,43 @@ function oe_whitelabel_search_preprocess_facets_summary_item_list(array &$variab
...
+ *
+ * Add region suggestions for the oe_whitelabel_search_form input field.
+ */
+function oe_whitelabel_search_theme_suggestions_input_alter(array &$suggestions, array $variables): void {
+  if (!isset($variables['element']['#form_id']) || $variables['element']['#form_id'] !== 'oe_whitelabel_search_form') {
+    return;
+  }
+  $suggestions[] = $variables['theme_hook_original'] . '__' . $variables['element']['#region'];
+  $suggestions[] = $variables['theme_hook_original'] . '__' . $variables['element']['#region'] . '__' . $variables['element']['#form_id'];
+}
```
donquixote: I found a way we can avoid all these `hook_theme_suggestions_HOOK_alter()` functions.
We can put the theme hook suggestion directly into the form builder class.
See https://github.com/openeuropa/oe_whitelabel/commit/2a5695e4b6cf4c3ff132ff1cb0b0a005afec7f4e

donquixote: Besides this, I am skeptical about how we use theme hook suggestions here.
But I heard it was already agreed on with @drishu, so don't change it just for me.

I will still share my thoughts:
- I think I would put `'__' . $form_id . '__' . $region` instead of `'__' . $region . '__' . $form_id`, because this would make a more natural fallback hierarchy.
- Perhaps we should not use the region as a selector, but something else, e.g. a checkbox in the block config, or by having two different block plugins. By using the region, we make life harder for sub-themes that have their own region names. I heard that apparently I said to use the region in the refinement, in that case, my bad!
- Some changes could be done with render element altering instead of custom templates. Either in the form builder class or the block plugin class, or using `hook_block_view_BASE_BLOCK_ID_alter()`.

But as said, don't follow these suggestions unless @drishu or @brummbar agree.

modules/oe_whitelabel_search/src/Form/SearchForm.php
```diff
@@ -45,58 +45,33 @@ public static function create(ContainerInterface $container) {
...
     $form['search_input'] = [
-      '#prefix' => '<div class="bcl-search-form__group w-100">',
-      '#suffix' => '</div>',
       '#type' => 'textfield',
       '#title' => $config['input']['label'],
       '#title_display' => 'invisible',
       '#size' => 20,
-      '#margin_class' => 'mb-0',
+      '#form_id' => $this->getFormId(),
+      '#region' => $config['form']['region'],
```
donquixote: I know we are adding these values for the theme hook suggestions.
If we follow my proposal we no longer need it, see https://github.com/openeuropa/oe_whitelabel/commit/2a5695e4b6cf4c3ff132ff1cb0b0a005afec7f4e

modules/oe_whitelabel_search/src/Form/SearchForm.php
```diff
@@ -45,58 +45,33 @@ public static function create(ContainerInterface $container) {
    * {@inheritdoc}
    */
   public function buildForm(array $form, FormStateInterface $form_state, array $config = NULL): array {
+    if (empty($config['input']['name'])) {
+      return [];
+    }
```
donquixote, line 50: Why do we add a check for this one value, if all the other values are assumed as given?
Is there a reason why this one value might be empty, but the others not?

modules/oe_whitelabel_search/src/Plugin/Block/SearchBlock.php
```diff
@@ -249,17 +241,8 @@ public function blockSubmit($form, FormStateInterface $form_state): void {
...
-      ]));
-    }
-
-    if ($values['button']['button_classes'] !== Html::cleanCssIdentifier($values['button']['button_classes'])) {
-      $form_state->setErrorByName('button][button_classes', $this->t('Field "@field_name" does not contain a valid css class.', [
-        '@field_name' => $form['button']['button_classes']['#title'],
-      ]));
-    }
+    $completeForm = $form_state->getCompleteForm();
+    $form_state->setValue('region', $completeForm['region']['#value']);
```
donquixote: This can be simplified: https://github.com/openeuropa/oe_whitelabel/commit/c16126c0d12f9b7c3d6d68a9f91d8488b845adbb

templates/overrides/search/form--header--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,21 @@
...
+      <div class="col-12 col-lg-6 offset-lg-3">
+        <form {{ attributes.addClass('bcl-search-form', 'submittable') }}>
+          <div class="bcl-search-form__group">
+            {{ children }}
+          </div>
+        </form>
+      </div>
+    </div>
+  </div>
+</div>
```
donquixote, line 21: Does all of this really belong into the form template?
For me it would belong into the page template or perhaps in the block template.
Also, in the new version in netlify, we now have a page title in the same region. I know we don't want to implement the latest changes in a ticket that is already in progress, but it does validate the idea that the outer wrapper divs should not really be part of the search form.

templates/overrides/search/form-element--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,30 @@
...
+{% set classes = [
+  'js-form-item',
+  'form-item',
+  'js-form-type-' ~ type|clean_class,
+  'form-item-' ~ name|clean_class,
+  'js-form-item-' ~ name|clean_class,
+  title_display not in ['after', 'before'] ? 'form-no-label',
+  disabled == 'disabled' ? 'form-disabled',
+  errors ? 'form-item--error',
+] %}
```
donquixote: All of these classes are now only used in the `if` part below, and even then they are questionable.

templates/overrides/search/form-element--oe-whitelabel-search-form.html.twig
donquixote: If we don't care about the "errors" section, and only want to get rid of wrappers, then we could simply set `''#theme_wrappers' => []` in the render element.
We could do this in the form builder class itself, or in the block plugin class. The question is only if we want to allow sub-themes to override it, or make the block plugin and the form builder class agnostic of the whitelabel theme.
TBD.

templates/overrides/search/form-element--search-api-autocomplete.html.twig
```diff
@@ -1,98 +0,0 @@
...
-    <div class="form-item--error-message invalid-feedback d-block">
-      {{ errors }}
-    </div>
-  {% endif %}
-  {% if description_display in ['after', 'invisible'] and description.content %}
-    <small{{ description.attributes.addClass(description_classes) }}>
-      {{ description.content }}
-    </small>
-  {% endif %}
-</div>
```
donquixote, line 98: I assume we remove this because it was too broad and would also target unwanted elements?
(and now we have more specific templates)

templates/overrides/search/form-element--oe-whitelabel-search-form.html.twig
donquixote: Found it:
https://www.drupal.org/project/drupal/issues/2578561
https://git.drupalcode.org/project/drupal/-/commit/e55a7d8dc32aae6029863ea8a55e48515354053e#fee981558c2f41b50b192c470cfda2c49bb3c29a

So there is an optional module `inline_form_errors` in core, which enables this functionality.

donquixote: Personally I would rather get rid of this `error` part, which would make the template fully empty except for `{{ children }}`.
Then we would set `$form['#disable_inline_form_errors'] = TRUE;` in the form, and not have to worry about inline errors even if `inline_form_errors` module is enabled.

However, I talked with @brummbar, and for now the conclusion is to keep the errors section in the template, just without the outer wrappers.
So only this part `<div class="form-item--error-message invalid-feedback d-block">...`.
I think the current version is taken from oebt, whereas the core version has fewer classes. To me it makes more sense to use the oebt version, given that this is our base class.

Either way it will look bad - but this shall be fixed later.

modules/oe_whitelabel_search/src/Form/SearchForm.php
donquixote: Whatever, keep it.

modules/oe_whitelabel_search/src/Form/SearchForm.php
donquixote: (nothing to do here. read and resolve.)

templates/overrides/page/page.html.twig
```diff
@@ -109,7 +109,7 @@
   {% endblock %}
   <main>
     {{ page.hero }}
-    <div class="container {{ page.header ? 'mt-2': 'mt-5' }}">
+    <div class="container mt-md-4-75 mt-4">
```
donquixote, line 112: Reasons?
Perhaps it is ok, just wondering.

templates/overrides/search/form--header--oe-whitelabel-search-form.html.twig
donquixote: I don't like it but let's keep it like this, until we change our minds.

templates/overrides/page/page.html.twig
escuriola: In all pages I reviewed in netlify has this, there is no condition.
Example:
image

templates/overrides/search/form-element--oe-whitelabel-search-form.html.twig
donquixote: @escuriola I see that you changed this code, removing the outer wrappers.
The agreement when we last talked was to remove only the outer wrappers (field-*), not the inner wrapper for the error element. Now you removed all of them.

templates/overrides/page/page.html.twig
donquixote: yep, seems alright.

templates/overrides/search/form-element--oe-whitelabel-search-form.html.twig
donquixote: I tried locally with `<div class="form-item--error-message invalid-feedback d-block">` as in form-element in oebt, and with `<div class="form-item--error-message">` as in core. None of them look really good, but I'd say the core version looks a bit cleaner. In the end it does not matter much: When people want to use these inline form errors, they can open a new issue.

donquixote (changes requested):

donquixote (dismissed):

drishu (dismissed):

drishu (approved):

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
```diff
@@ -82,9 +82,46 @@ protected function setUp(): void {
...
+      ->get('entity_type.manager')
+      ->getStorage('block');
+    $entity = $block_entity_storage->load('oe_whitelabel_search_form');
+    $builder = \Drupal::entityTypeManager()->getViewBuilder('block');
+    $build = $builder->view($entity, 'block');
+    $render = $this->container->get('renderer')->renderRoot($build);
+    $crawler = new Crawler($render->__toString());
+
+    // Assert the form rendering.
+    $form = $crawler->filter('form');
```
donquixote, line 98: Shouldn't we keep the version from 1.x here?
```suggestion
    // Select the search form in the block.
    // The block template removes the block wrapper, so the form is the root
    // element.
    $form = $crawler->filter('body > form#oe-whitelabel-search-form');
```
This is from OEL-1543.

donquixote (approved):

PR #138, opened 2022-05-12
OEL-1514: Improve list pages integration in whitelabel

Maxfire:
Fixes on list pages.


Maxfire:
closed cause is created a new one here https://github.com/openeuropa/oe_whitelabel/pull/139

PR #139, opened 2022-05-17
OEL-1514: Improve list pages integration in whitelabel

Maxfire:


templates/list_pages/item-list--oe-list-pages-results.html.twig
```diff
@@ -3,19 +3,6 @@
...
- * - title: The title of the list.
- * - list_type: The tag for list element ("ul" or "ol").
- * - wrapper_attributes: HTML attributes to be applied to the list wrapper.
- * - attributes: HTML attributes to be applied to the list.
- * - empty: A message to display when there are no items. Allowed value is a
- *   string or render array.
- * - context: A list of contextual data associated with the list. May contain:
- *   - list_style: The custom list style.
- *
  * @see template_preprocess_item_list()
```
escuriola:
```suggestion
 * @see ./core/modules/system/templates/item-list.html.twig
```

escuriola (changes requested):

escuriola (dismissed):

templates/list_pages/oe-list-pages-selected-facet.html.twig
```diff
@@ -1,9 +1,13 @@
...
+    {{ pattern('badge', {
+      'background': 'light',
+      'label': item.label,
+      'url': item.url,
+      'dismissible': true,
+      'attributes': create_attribute({'class': ['me-2', 'fs-6']}),
+    }) }}
+  {%- endfor -%}
+</div>
+{%- endif %}
```
donquixote: No apply spaceless, but `{%-` syntax?

templates/list_pages/item-list--oe-list-pages-results.html.twig
```diff
@@ -3,20 +3,7 @@
  * @file
  * Item list theme override for a oe_list_pages list template.
```
donquixote: The comment is still not great, but ok to keep it.
```suggestion
 * Override for item list showing a list of entities on a list page.
```
Not sure about it, but the idea is to describe when this template is used.

modules/oe_whitelabel_list_pages/tests/src/Functional/ListPagesTest.php
```diff
@@ -72,36 +109,81 @@ public function testListPage(): void {
...
-    $assert_session->elementExists('css', 'input[data-drupal-selector="edit-submit"]');
-    $assert_session->elementExists('css', 'input[data-drupal-selector="edit-reset"]');
+  /**
+   * Asserts the title in the facets summary.
+   *
+   * @param int $expected_count
+   *   Expected number of results to be reported in the title.
+   * @param \Behat\Mink\Element\ElementInterface $container
+   *   Document to check against.
+   */
```
donquixote, line 122: Description for `$container` parameter is not correct.
The element passed here is not the "document", it is the "right column", so `'div.row > .col-lg-9'`.
I wonder if we even need this parameter or if each assert method should do its own `$this->assertSession()->elementExists('div.row > .col-lg-9')->...`, or use a helper to get that container element.
A parameter is only useful if we could also pass something else instead.
But I'm ok to keep it like this if we want to move forward.

modules/oe_whitelabel_list_pages/tests/src/Functional/ListPagesTest.php
```diff
@@ -72,36 +109,81 @@ public function testListPage(): void {
...
+  protected function assertFacetsSummaryTitle(int $expected_count, ElementInterface $container): void {
+    $facets_summary = $container->find('css', 'div.col-md-6.col-lg-8');
+    $title = $facets_summary->find('css', 'h4');
```
donquixote: Can be simplified:
```suggestion
    $title = $container->find('css', 'div.col-md-6.col-lg-8 h4');
```
Also, I wonder if we want to be so detailed with the selectors.
I think it's ok because this is our template, we are adding the classes. They are not coming from bcl. When we change them, we should change the test as well.
But then the question is how far do we want to go with this? E.g. would we also want to test for `'.text-capitalize'`?

Btw in newest netlify / bcl this has changed again. (assuming that what we see here was copied from bcl at some point)

modules/oe_whitelabel_list_pages/tests/src/Functional/ListPagesTest.php
```diff
@@ -72,36 +109,81 @@ public function testListPage(): void {
...
+   * Asserts link text on pager links.
+   *
+   * @param int $expected_count
+   *   Expected number of links in the pager.
+   * @param \Behat\Mink\Element\ElementInterface $container
+   *   Document to check against.
+   */
+  protected function assertPager(int $expected_count, ElementInterface $container): void {
+    $links = $container->findAll('css', 'ul.pagination > li.page-item');
+    $this->assertCount($expected_count, $links);
```
donquixote, line 156: We could test the `?page=` query of the last pager link instead, this would be more semantic.
But it's ok.

templates/list_pages/oe-list-pages-selected-facet.html.twig
```diff
@@ -1,9 +1,13 @@
...
-  {{ pattern('badge', {
-    'background': 'light',
-    'label': item.label,
-    'url': item.url,
-    'dismissible': true,
-    'attributes': create_attribute({'class': ['me-2', 'fs-6']}),
-  }) }}
-{% endfor %}
+{% if items -%}
+<div class="col-md-12 order-md-2 mt-3">
```
donquixote: Why are we doing this?
For me this has the effect that now each badge is on a new line. I suspect this is not intended.

templates/list_pages/node--oe-list-page--full.html.twig
```diff
@@ -24,15 +25,11 @@
...
       </div>
-      <div class="col-md-12 order-md-2 mt-3">
-        {{ content['extra_field_oe_list_page_selected_filters:node:oe_list_page'] }}
-      </div>
-      <div class="col-md-6 col-lg-4 mt-3 mt-md-0">
-      </div>
+      {{ content['extra_field_oe_list_page_selected_filters:node:oe_list_page'] }}
     </div>
     {{ content['extra_field_oe_list_page_results:node:oe_list_page'] }}
   </div>
```
donquixote, line 44: I see some discrepancy to netlify.
But this is likely due to the new bcl version.
What should we do first? Merge the bcl upgrade, or merge list pages?

templates/list_pages/node--oe-list-page--full.html.twig
```diff
@@ -24,15 +25,11 @@
     <div class="row">
       <div class="col-md-6 col-lg-8 align-self-center">
```
donquixote: I think we only need these wrappers and classes if we also have a sort element. Which currently we don't.

donquixote (changes requested):

templates/list_pages/oe-list-pages-selected-facet.html.twig
drishu: yes, as in the parent template

templates/list_pages/node--oe-list-page--full.html.twig
drishu: list pages

templates/list_pages/item-list--oe-list-pages-results.html.twig
drishu: no, the template actually wraps items in the bcl listing, I'll improve

modules/oe_whitelabel_list_pages/tests/src/Functional/ListPagesTest.php
drishu: It is agnostic of the right column, that was the point behind making it a parameter.
I don't mind what we did with the "getSearchTopRegion" but here I want to put emphasis on the structure of the page:
get left column
  assert stuff using that as context
get right column
  assert stuff using that context
  
so either I use it as a parameter to point out that this stuff is there, in the $right_column or I add it in the helper names, like assertRightColumnFacetsSummaryTitle and inside do getRightColumn(), up 2 u Mr Reviewer

modules/oe_whitelabel_list_pages/tests/src/Functional/ListPagesTest.php
drishu: my compass is to leave the ones that are not utility related but structure/layout related, so I dont add mt-4 or ps-0 but I add col-12 and such.
Its a bit loose as if I add the classes myself then I might add more classes (style related) just to assert that the code that adds the classes is not broken. 
Not 100% convinced of the best way neither..

templates/list_pages/oe-list-pages-selected-facet.html.twig
drishu: the reason is this https://github.com/openeuropa/oe_whitelabel/pull/139/commits/b497276e21a9eeec0d2e7913eb532b6de4711463#diff-d807785e35b0f83527fdc91d1d714887511d69d2efcb8db701879af84ec089c9R33
a dubious check and condition was added there when in fact that check belongs to drupal.

Good catch with the multiple lines, I reverted to the previous logic

templates/list_pages/node--oe-list-page--full.html.twig
drishu: we also have the wrapper from the selected filters `div.col-md-12 order-md-2 mt-3` and sort will come.
We dont have sort yes but want to be as close as possible to netlify.

templates/list_pages/item-list--oe-list-pages-results.html.twig
```diff
@@ -1,22 +1,9 @@
 {#
 /**
  * @file
- * Item list theme override for a oe_list_pages list template.
+ * Override for item list wrapping items in a bcl-listing.
```
donquixote, line 4: I would prefer if the comment describes which case we target, rather than what the template does. But no strong opinion. Let's keep it.

modules/oe_whitelabel_list_pages/tests/src/Functional/ListPagesTest.php
donquixote: Well at least call it "Element to check against" or "Container with the expected elements", not "Document to check against". For me, "document" means a complete html page.
Let's keep the signature, just fix the comment.

templates/list_pages/oe-list-pages-selected-facet.html.twig
donquixote: Very dubious indeed.

templates/list_pages/node--oe-list-page--full.html.twig
```diff
@@ -24,15 +25,16 @@
...
         <h4 class="mb-0">
-          <span class="text-capitalize">{{ label|render|striptags|trim }}</span>
-          ({{ content['extra_field_oe_list_page_total_count:node:oe_list_page']|render|trim }})
+          <span class="text-capitalize">{{ label }}</span>
+          ({{ content['extra_field_oe_list_page_total_count:node:oe_list_page'] }})
         </h4>
       </div>
+      {% set selected_filters = content['extra_field_oe_list_page_selected_filters:node:oe_list_page'] %}
+      {% if selected_filters[0]|length > 1 %}
```
donquixote: This check is a but sus. Let's look for a different solution.

templates/list_pages/node--oe-list-page--full.html.twig
```diff
@@ -1,39 +1,45 @@
...
+          </h4>
+        </div>
+        {% set selected_filters %}
+          {{ content['extra_field_oe_list_page_selected_filters:node:oe_list_page']|merge({'#theme': 'field__bare'}) }}
+        {% endset %}
+        {% if selected_filters|trim is not empty %}
+          <div class="col-md-12 order-md-2 mt-3">
+            {{ selected_filters }}
+          </div>
+        {% endif %}
```
donquixote, line 40: Just a note: I would call the variable something like `selected_filters_printed`, or even with underscore `_selected_filters_printed`. This would communicate that the value is captured markup, and not some kind of structured array.

But let's keep it like this, good enough.

modules/oe_whitelabel_list_pages/tests/src/Functional/ListPagesTest.php
```diff
@@ -72,36 +109,80 @@ public function testListPage(): void {
...
+   *
+   * @param string[] $expected
+   *   Expected badge labels.
+   * @param \Behat\Mink\Element\ElementInterface $container
+   *   Container with the expected elements.
+   */
+  protected function assertActiveFilterBadges(array $expected, ElementInterface $container): void {
+    $badges = $container->findAll('css', '.badge');
+    $this->assertElementsTexts($expected, $badges);
+  }
```
donquixote, line 144: Would be nice if we could verify that the wrapper div is gone if no active filters.
```suggestion
  protected function assertActiveFilterBadges(array $expected, ElementInterface $container): void {
    $active_filters_container = $container->find('css', 'div.col-md-12.order-md-2.mt-3');
    if ($expected === []) {
      $this->assertNull($active_filters_container);
      return;
    }
    $this->assertNotNull($active_filters_container);
    $badges = $active_filters_container->findAll('css', '.badge');
    $this->assertElementsTexts($expected, $badges);
  }
```
I don't really like the selectors we need here, but I can't think of anything better - unless we want to add more classes just for the test.

donquixote (commented):
Some feedback, TBD if we want to do it or ignore and move on.

donquixote (approved):

donquixote:
Other changes can be done in follow-up issues.

escuriola (approved):

PR #142, opened 2022-05-17
OEL-1585: Fix logo url link in neutral flavour and test coverage.

escuriola:


brummbar (changes requested):
The reason I suggested to use front_page is because of this:

image

It will make more clear that we are pointing to the front_page and that variable is ready already.

brummbar (approved):

drishu (approved):

PR #144, opened 2022-05-20
OEL-1618: Upgrade BCL to 0.23.0.

escuriola:


donquixote:
This seems to be ok.
However, I suspect we now have duplicate `<article>` tags in node (teaser) templates.
(I think only event teaser and news teaser)

Let's:
- Remove the outer `<article>` wrapper, if we include a card pattern that already does add an article wrapper via bcl-card.
- Pass the node attributes into the pattern.
- Add a test line to count the `article` tags when we render a news or event teaser.

Also, after changing card h* tags in oebt to h1, we need to update the tests here in whitelabel.

donquixote:
I notice we have a bunch of templates where we put custom hx elements with `fw-bold` class.
I found these by looking for "fw-bold" in templates, and then checking if it is used for hx elements.
```
templates/content/field--node--body--oe-sc-event.html.twig
templates/content/field--node--oe-documents--oe-sc-event.html.twig
templates/overrides/navigation/oe-corporate-blocks-eu-footer.html.twig
templates/overrides/navigation/oe-corporate-blocks-neutral-footer.html.twig
templates/paragraphs/paragraph--oe-description-list.html.twig
templates/paragraphs/paragraph--oe-rich-text.html.twig
```
We need to compare these to netlify. In most cases I checked, the hx tags in netlify don't have the fw-bold class.

We could also consider, instead of having custom title elements, to use the new option of sending the title as a parameter to the pattern or bcl template. Unfortunately the patterns currently do not support this option.

TBD if we do this here or in a follow-up issue.

donquixote:
Conclusion: Fix these in the same issue.

modules/oe_whitelabel_paragraphs/tests/src/Kernel/Paragraphs/ContentRowTest.php
```diff
@@ -159,17 +159,17 @@ public function testWithInpageNavigationRendering(): void {
...
+    $links_block_title = $content->filter('h2.bcl-heading')->eq(1);
     $this->assertSame('Links block test', $links_block_title->text());
     $facts_figures = $content->filter('div.bcl-fact-figures--default');
     $this->assertStringContainsString('Facts and Figures test', $facts_figures->filter('h2.bcl-heading')->text());
     $blockquote_blockquote = $content->filter('blockquote.blockquote');
     $this->assertStringContainsString('Maecenas id urna eleifend', $blockquote_blockquote->text());
     $blockquote_footer = $content->filter('figcaption.blockquote-footer');
     $this->assertSame('Quote 1', trim($blockquote_footer->text()));
-    $social_media_title = $content->filter('h2.bcl-heading')->eq(2);
+    $social_media_title = $content->filter('h2.bcl-heading')->eq(3);
```
donquixote, line 172: Seems a bid odd that we have to rely on `->eq(*)` here, instead of having a specific selector.
But I'd say it's good enough: We check the text later, it is unlikely that the title would appear on the wrong block.

donquixote:
I listed 6 templates above. I see only 2 of them were changed.

templates/content/field--node--body--oe-sc-event.html.twig
```diff
@@ -32,9 +24,20 @@
...
+      {%- endif -%}
+    {%- endset -%}
+    {% include '@oe-bcl/bcl-heading/heading.html.twig' with {
+      title: _label,
+      title_tag: 'h3',
+      attributes: create_attribute().addClass([
+        'field__label',
+        label_display == 'visually_hidden' ? 'visually-hidden',
+      ])
+    } only %}
```
donquixote, line 40: Can't we keep the title_classes variable at the top where it was? Just remove the fw-bold.

donquixote: Also do we need to check for different values of `label_display`, given that we control the settings? And do we need the `field__label` class? The only reason would be that we already did this before and we don't want to change too much in this pull request.

The other problem with this template is that it targets all view modes, not just 'full'.

So I see two options:
1. Only do minimal changes to the template. In that case, put the title_classes at the top where it was.
2. Fix the template: Target "--full" view mode, remove unnecessary checks and classes.

I think in this issue we should go with option 1.

templates/content/field--node--body--oe-sc-event.html.twig
```diff
@@ -32,9 +24,20 @@
...
 {% else %}
   <div{{ attributes.addClass(classes) }}>
-    <h3{{ title_attributes.addClass(title_classes) }}>
-      {{ label }}{% if label_display == 'inline' %}<span class="me-1">:</span>{% endif %}
-    </h3>
+    {%- set _label -%}
+      {{- label -}}
+      {%- if label_display == 'inline' -%}
+        <span class="me-1">:</span>
+      {%- endif -%}
```
donquixote: I think the extra line breaks add more clutter. Keep it as it was.
```suggestion
      {{- label -}}{%- if label_display == 'inline' -%}<span class="me-1">:</span>{%- endif -%}
```
Btw, I don't see these ':' in core field template, only in oebt.
Also, ':' + inline does not make sense in combination with the h3.
The only reason to keep it is to reduce the scope of changes in this PR.

donquixote: Let's keep it as is for now, up to second reviewer.

templates/content/field--node--oe-documents--oe-sc-event.html.twig
```diff
@@ -32,9 +24,20 @@
...
+      {%- endif -%}
+    {%- endset -%}
+    {% include '@oe-bcl/bcl-heading/heading.html.twig' with {
+      title: _label,
+      title_tag: 'h3',
+      attributes: create_attribute().addClass([
+        'field__label',
+        label_display == 'visually_hidden' ? 'visually-hidden',
+      ])
+    } only %}
```
donquixote, line 40: Same feedback as for the other field above.

templates/content/node--oe-sc-event--teaser.html.twig
```diff
@@ -8,15 +8,14 @@
...
+{{ pattern('card', {
+  variant: 'search',
+  title: _title,
+  text: content.oe_summary,
+  image: image,
+  meta: [
+    content.oe_sc_event_dates|field_value,
+  ],
+  attributes: attributes,
+}) }}
```
donquixote, line 20: Note to 2nd reviewer:
- The article tag is now included in card pattern.
- Passing 'attributes' to the card pattern means that all the entity attributes will now be on the card.

templates/list_pages/oe-list-pages-selected-facet.html.twig
```diff
@@ -1,8 +1,9 @@
+{{ dump() }}
```
donquixote: @escuriola Please remove!

donquixote:
@escuriola And now project has the wrong heading fonts!
`templates/content/node--oe-project--full.html.twig`
Sorry!

templates/content/field--node--body--oe-sc-event.html.twig
donquixote: Oh, I see h2 in netlify, not h3.

templates/content/field--node--oe-documents--oe-sc-event.html.twig
donquixote: As above, in netlify this is h2, not h3.

donquixote (changes requested):

donquixote:
> @escuriola And now project has the wrong heading fonts! `templates/content/node--oe-project--full.html.twig` Sorry!

And list page too - netlify has h2 for heading on top of the results.
`templates/list_pages/node--oe-list-page--full.html.twig`

also search page - netlify has h2 for heading on top of the results.
`templates/overrides/search/block--facets-summary-block.html.twig`

Perhaps there is more.

templates/content/field--node--body--oe-sc-event.html.twig
```diff
@@ -32,9 +31,14 @@
...
   <div{{ attributes.addClass(classes) }}>
-    <h3{{ title_attributes.addClass(title_classes) }}>
-      {{ label }}{% if label_display == 'inline' %}<span class="me-1">:</span>{% endif %}
-    </h3>
+    {%- set _label -%}
+      {{- label -}}{%- if label_display == 'inline' -%}<span class="me-1">:</span>{%- endif -%}
+    {%- endset -%}
+    {% include '@oe-bcl/bcl-heading/heading.html.twig' with {
+      title: _label,
+      title_tag: 'h2',
```
donquixote: We don't need to specify h2, it's the default.
Not a blocker though.

templates/content/field--node--oe-documents--oe-sc-event.html.twig
```diff
@@ -32,9 +31,14 @@
...
   <div{{ attributes.addClass(classes) }}>
-    <h3{{ title_attributes.addClass(title_classes) }}>
-      {{ label }}{% if label_display == 'inline' %}<span class="me-1">:</span>{% endif %}
-    </h3>
+    {%- set _label -%}
+      {{- label -}}{%- if label_display == 'inline' -%}<span class="me-1">:</span>{%- endif -%}
+    {%- endset -%}
+    {% include '@oe-bcl/bcl-heading/heading.html.twig' with {
+      title: _label,
+      title_tag: 'h2',
```
donquixote: h2 is the default, we can omit it.
Not a blocker though.

templates/content/field--node--body--oe-sc-event.html.twig
donquixote: Actually now I see this all over the place. Let's fix it.

templates/overrides/search/block--facets-summary-block.html.twig
```diff
@@ -14,7 +14,14 @@
     {{ title_prefix }}
     {% if label %}
       <div class="col-md-6 col-lg-8 align-self-center">
-        <h4 class="mb-0"><span class="text-capitalize">{{ label }} </span> {{ results_count }}</h4>
+        {% set _title %}
+          {{ label|render|striptags }} ({{ results_count }})
+        {% endset %}
+        {% include '@oe-bcl/bcl-heading/heading.html.twig' with {
+          title: _title,
+          title_tag: 'h2',
```
donquixote: h2 is default.

templates/content/node--oe-project--full.html.twig
```diff
@@ -6,15 +6,22 @@
...
 
 {% macro field_with_header(field) %}
   {% set _content %}{{ field|field_value }}{% endset %}
   {% if _content|trim %}
-    {{ _self.header(field|field_label, field['#inpage_nav_id'] ?? null) }}
+    {% set _attributes = create_attribute().addClass(['mb-4', 'pt-3']) %}
+    {% if field['#inpage_nav_id'] is not empty %}
+      {% set _attributes = _attributes.setAttribute('id', field['#inpage_nav_id']) %}
+    {% endif %}
+    {{ _self.header(field|field_label, _attributes) }}
```
donquixote: Not convinced of these changes.
In the macro I intentionally did _not_ add 'pt-3' because we should fully rely on margins, not paddings, for vertical spacing.
The first element (usually "Project details", but could be another one if details is empty) does not need special treatment, because we absorb the top margin of the first element with `<div class="h-0 invisible mt-n5 mb-5"></div>`.

Let's remove the commit from this branch and create a follow-up instead, as discussed the other day - we don't want to fix every hx title in this issue, only the "components".

donquixote: (yes this contradicts my review comment where I asked for this change, but our conversation about scope happened after that)

modules/oe_whitelabel_list_pages/tests/src/Functional/ListPagesTest.php
```diff
@@ -121,13 +121,8 @@ protected function createListPage(): NodeInterface {
...
   protected function assertFacetsSummaryTitle(int $expected_count, ElementInterface $container): void {
-    $title = $container->find('css', 'div.col-md-6.col-lg-8 h4');
-    $text = $title->find('css', 'span.text-capitalize');
-
-    $this->assertSame(
-      sprintf('%s (%s)', $text->getText(), $expected_count),
-      $title->getText()
-    );
+    $title = $container->find('css', 'div.col-md-6.col-lg-8 h2.text-capitalize');
+    $this->assertSame('News list page ' . sprintf('(%s)', $expected_count), $title->getText());
```
donquixote: I like this simplification. The list page title is always the same, so no reason to make it dynamic.

Although, the usage of sprintf is a bit strange here. Either use string concat all the way or use sprintf() for the complete string.
```suggestion
    $this->assertSame(sprintf('News list page (%s)', $expected_count), $title->getText());
```
or
```suggestion
    $this->assertSame('News list page (' . $expected_count . ')'), $title->getText());
```

templates/list_pages/node--oe-list-page--full.html.twig
```diff
@@ -25,10 +25,14 @@
     <div class="col-12 col-lg-9 col-xxl-8">
       <div class="row">
         <div class="col-md-6 col-lg-8 align-self-center">
-          <h4 class="mb-0">
-            <span class="text-capitalize">{{ label }}</span>
-            ({{ content['extra_field_oe_list_page_total_count:node:oe_list_page'] }})
-          </h4>
+          {% set _title %}
+            {{ label|render|striptags }}({{ content['extra_field_oe_list_page_total_count:node:oe_list_page'] }})
+          {% endset %}
```
donquixote: We miss a space before the '('.
On the other hand, we get free space from the field template. The `|striptags` removes tags, but not spaces.
The `apply spaceless` that is applied to the entire template only removes space between tags, not within tags, so we should add the `-` here.
Let's do this in a cleaner way so we get exactly one space before `(` but no extra space elsewhere.
```suggestion
          {% set _title -%}
            {{ label|field_value }} ({{ content['extra_field_oe_list_page_total_count:node:oe_list_page'] }})
          {%- endset %}
```
I did a `{{ dump(label['#cache']) }}` to confirm there is no cache data we would lose by calling `|field_value`. And we already plan to fix cache handling for `|field_value` anyway.

templates/list_pages/node--oe-list-page--full.html.twig
```diff
@@ -25,10 +25,14 @@
...
+          {% include '@oe-bcl/bcl-heading/heading.html.twig' with {
+            title: _title,
+            title_tag: 'h2',
+            attributes: create_attribute().addClass(['mb-0', 'text-capitalize']),
+          } only %}
```
donquixote: Btw this is one of the candidates to handle in a follow-up issue.
It just feels wrong to me to split it out now that we already touched it.

templates/overrides/search/block--facets-summary-block.html.twig
```diff
@@ -14,7 +14,14 @@
     {{ title_prefix }}
     {% if label %}
       <div class="col-md-6 col-lg-8 align-self-center">
-        <h4 class="mb-0"><span class="text-capitalize">{{ label }} </span> {{ results_count }}</h4>
+        {% set _title %}
+          {{ label|render|striptags }} ({{ results_count }})
+        {% endset %}
```
donquixote: For me this results in duplicate '(': "Search Results ((19))".
This is because the '(' is already in `facets-summary-count.html.twig`.

Also the `|render|striptags` is pointless here because unlike for list page, `label` is a string, not a field render element.

Let's move this change to a follow-up issue, following our conversation last week.

templates/list_pages/node--oe-list-page--full.html.twig
donquixote: Following our discussion from last week, we could move this change to a follow-up and leave the template unchanged in this issue.

donquixote (changes requested):

donquixote (dismissed):

tests/src/PatternAssertions/CardAssert.php
```diff
@@ -33,7 +33,7 @@ protected function getAssertions($variant): array {
       ],
```
drishu, line 33: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1771

drishu (approved):

brummbar (approved):

PR #150, opened 2022-05-24
OEL-1571: Display country instead skos entity reference country URL.

escuriola:


modules/oe_whitelabel_contact_forms/oe_whitelabel_contact_forms.module
```diff
@@ -74,13 +74,11 @@ function oe_whitelabel_contact_forms_preprocess_status_messages(&$variables) {
...
-      $value = 'value';
-      if ($field->getFieldDefinition()->getType() == 'skos_concept_entity_reference') {
-        $value = 'target_id';
-      }
       $items[] = [
         'term' => $field->getFieldDefinition()->getLabel(),
-        'definition' => $field->first()->getValue()[$value],
+        'definition' => $field->getFieldDefinition()->getType() === 'skos_concept_entity_reference'
+        ? $field->referencedEntities()
+        : $field->first()->getValue()['value'],
```
donquixote: Wrong indentation. I know it is PhpStorm's fault but we should fix manually.
```suggestion
        'definition' => $field->getFieldDefinition()->getType() === 'skos_concept_entity_reference'
          ? $field->referencedEntities()
          : $field->first()->getValue()['value'],
```

donquixote: Also, by passing a Concept entity to the template, the template will implicitly call `->label()` on that entity, through this line in bcl-descritpion-list.html.twig:
```twig
                {{- _definition.label -}}
```
This is how it works: If the value is an object, then it will try `$obj->$key` and then `$obj->$key()`.
See `twig_get_attribute()`. Crazy, eh?

Let's not rely on this magic.
Instead, let's do something like this:
(please don't copy this 1:1, it is only for the rough direction)

```php
/** @var \Drupal\rdf_skos\Entity\Concept[] $concept_entities */
$concept_entities = $field->referencedEntities();
if (!isset($concept_entities[0])) {
  continue;
}
$items[] = [
  'term' => $field->getFieldDefinition()->getLabel(),
  'definition' => [['label' => $concept_entities[0]->label()]],
];
```

donquixote: I think it is complex enough that we should use if/else instead of a ternary.

modules/oe_whitelabel_contact_forms/oe_whitelabel_contact_forms.module
```diff
@@ -74,13 +74,22 @@ function oe_whitelabel_contact_forms_preprocess_status_messages(&$variables) {
...
-      if ($field->getFieldDefinition()->getType() == 'skos_concept_entity_reference') {
-        $value = 'target_id';
+
+      $definition = '';
+      if ($field->getFieldDefinition()->getType() === 'skos_concept_entity_reference') {
+        /** @var \Drupal\rdf_skos\Entity\Concept[] $concept_entities */
+        $concept_entities = $field->referencedEntities();
+        if (isset($concept_entities[0])) {
+          $definition = [['label' => $concept_entities[0]->label()]];
+        }
```
donquixote: If `$concept_entities[0]` is not set, we would rather skip (= `continue;`) this field instead of showing empty.
On the other hand, this case cannot really happen because we checked `->isEmpty()` above.

I imagine @brummbar or @drishu would say we don't even need to check `isset()` because we fully rely on the `->isEmpty()` check. I am not sure about it, there could be a case where the field is not empty, but the referenced entity does not exist or cannot be loaded.

Btw In oe_theme, we call `$field->view(['label' => 'hidden', ...])` instead of what we do here.
The result is a render element.
But if we use this, we would want to get rid of field wrappers, ideally.
But for now let's use `->label()`, we can rethink later.

donquixote: Btw: As long as the value is a simple string, we can skip the `[[..]]` array levels here, and also the 'label'.
This is due to how bcl-item-list works:
- If `$item['definition']` is "iterable" (always true for arrays), then the format has to be `$item['definition'][]['label'] = 'hello'`.
- If `$item['definition']` is a string or non iterable object, then the format can be `$item['definition'] = 'hello'`, so we save multiple array levels. This is what we already do for other values right now. But this only works because these are simple strings instead of render elements.

drishu: yes, for me the empty check is sufficient

donquixote: Ok, so let's

- Drop the `[[`.
- Drop the `isset` check, and assume that the skos entity exists.

modules/oe_whitelabel_contact_forms/oe_whitelabel_contact_forms.module
```diff
@@ -74,13 +74,20 @@ function oe_whitelabel_contact_forms_preprocess_status_messages(&$variables) {
       if ($field->isEmpty() || !$field->access()) {
         continue;
       }
-      $value = 'value';
-      if ($field->getFieldDefinition()->getType() == 'skos_concept_entity_reference') {
-        $value = 'target_id';
+
+      $definition = '';
```
donquixote: This line is not needed now.

donquixote (approved):

drishu (approved):

PR #151, opened 2022-05-25
OEL-1418: Update oe_whitelabel after remove paragraphs required modules on oe_bootstrap_theme.

escuriola:
…

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         "drupal/core": "^9.2",
         "drupal/twig_field_value": "^2.0",
         "openeuropa/composer-artifacts": "^1.0.0-alpha1",
-        "openeuropa/oe_bootstrap_theme": "1.0.0-beta1"
+        "openeuropa/oe_bootstrap_theme": "0.1418.202205311215"
     },
     "require-dev": {
         "composer/installers": "^1.11",


composer.json
```diff
@@ -38,7 +38,7 @@
         "openeuropa/oe_contact_forms": "~1.1",
         "openeuropa/oe_content": "^2.8.0",
         "openeuropa/oe_corporate_blocks": "^4.4",
-        "openeuropa/oe_media": "^1.14",
+        "openeuropa/oe_media": "^1.16",
```
drishu: why do you need to bump this up? what fails ?

escuriola: Because is the oe_bootstrap_theme 1.x required version: https://github.com/openeuropa/oe_bootstrap_theme/blob/1.x/composer.json

So as I removed there, I updated in the other components.

composer.json
```diff
@@ -78,6 +78,11 @@
                 }
             }
         },
+        "patches": {
+            "openeuropa/oe_bootstrap_theme": {
+                "latest": "https://github.com/openeuropa/oe_bootstrap_theme/compare/1.0.0-beta1..OEL-1418.diff"
```
drishu: we make major composer changes, so you need a dev tag rather

composer.json
drishu: No, that was a mistake on my side, so I'm glad we removed :)
But please drop the commit that bumbs this

drishu (changes requested):

drishu (approved):

escuriola:
PR Only for testing purposes.

PR #152, opened 2022-05-25
OEL-1540: Fix Scrollspy js file requirement when there is no items to display.

escuriola:


templates/content/node--oe-sc-event--full.html.twig
```diff
@@ -45,10 +45,14 @@
 {% endset %}
 
 <article{{attributes}}>
-  {{ pattern('inpage_navigation', {
-    title: 'Page content',
-    links: inpage_navigation_links,
-    content: inpage_navigation_fields,
-    full_layout: true,
-  }) }}
+  {% if inpage_navigation_links is null %}
```
donquixote: Please use is empty, so we also skip if empty array.

To clarify: NULL currently works here, but only because we early return in the preprocess before we initialize the variable.
We should not rely on this, a simple refactoring could lead to having an empty array here instead of NULL.

donquixote: (resolve once read, ignore)
Btw the variable should really be named `inpage_navigation_content` instead of `inpage_navigation_links`, but out of scope here.

tests/src/Functional/ContentEventRenderTest.php
```diff
@@ -153,6 +153,17 @@ public function testEventPage(): void {
       'Event body',
       $crawler->filter('#oe-content-body p')->text()
     );
+
+    // Assert inpage_navigation not loaded if there is no body and documents.
+    /** @var \Drupal\node\Entity\Node $node */
+    $node = \Drupal::entityTypeManager()->getStorage('node')->load($node->id());
```
donquixote: Do we need to load a fresh instance of `$node`? I don't think it is necessary.
Also, if we really need, we could just 
```suggestion
    $node = Node::load($node->id());
```
and skip the `@var` comment above.

donquixote (dismissed):

templates/content/node--oe-sc-event--full.html.twig
```diff
@@ -45,10 +45,14 @@
 {% endset %}
```
drishu, line 45: right now we produce an empty 'description_list' which adds unwanted white space if 'description_items' is empty, this was no problem before, but since we made this change, we need to check description_items is not empty now before populating inpage_navigation_fields

drishu (changes requested):

templates/content/node--oe-sc-event--full.html.twig
```diff
@@ -45,10 +45,14 @@
...
 
 <article{{attributes}}>
-  {{ pattern('inpage_navigation', {
-    title: 'Page content',
-    links: inpage_navigation_links,
-    content: inpage_navigation_fields,
-    full_layout: true,
-  }) }}
+  {% if inpage_navigation_links is empty and description_items is not empty %}
```
drishu, line 48: this is ugly and complex, would be better to wrap an if around the `set inpage_navigation_links`

templates/content/node--oe-sc-event--full.html.twig
drishu: this was a mistake actrualy on my side, inpage_navigation_fields should not depend on description_items

templates/content/node--oe-sc-event--full.html.twig
drishu: actually drop the commit that does this description_items check, was my mistake

drishu (dismissed):

donquixote (dismissed):

drishu (approved):

donquixote (approved):

PR #153, opened 2022-06-02
Fix displaying HR when pager is not displayed

piotrsmykaj:
Drupal returns the pager element in the template, even if it's configured to be enabled but it's not displayed because there are no results. This change is to display the HR line only when the pager is displayed. 

Below screenshot:
MicrosoftTeams-image (3)

### DIFF
--- a/templates/overrides/navigation/pager.html.twig
+++ b/templates/overrides/navigation/pager.html.twig
@@ -6,6 +6,8 @@
 #}
 {% extends "@oe_bootstrap_theme/overrides/navigation/pager.html.twig" %}
 {% block pager %}
-  <hr class="d-none d-md-block mt-4-5">
-  {{ parent()}}
+  {% if (parent()) %}
+    <hr class="d-none d-md-block mt-4-5">
+    {{ parent()}}
+  {% endif %}
 {% endblock %}


kp77:
+1 for this, we have the same issue. There are cases when the pager is in the output build, regardless if there are pages or not (e.g. `\Drupal\comment\Plugin\Field\FieldFormatter\CommentDefaultFormatter::viewElements()`)

The pager template should always check if there are items before adding any output, as in the case of the parent templates.

templates/overrides/navigation/pager.html.twig
```diff
@@ -6,6 +6,8 @@
 #}
 {% extends "@oe_bootstrap_theme/overrides/navigation/pager.html.twig" %}
 {% block pager %}
-  <hr class="d-none d-md-block mt-4-5">
-  {{ parent()}}
+  {% if (parent()) %}
```
drishu, line 9: I think the condition should be kept as in the system template

drishu (changes requested):
I would have loved to use this PR but the branch naming and commit message are not as per convention (https://github.com/openeuropa/documentation/blob/master/docs/initiative/how-to-contribute.md), so I have to use a new one https://github.com/openeuropa/oe_whitelabel/pull/204

PR #158, opened 2022-06-14
OEL-1653: Show time in event dates

escuriola:


modules/oe_whitelabel_starter_event/src/Plugin/Field/FieldFormatter/EventDateRangeFormatter.php
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function viewElements(FieldItemListInterface $items, $langcode) {
+    $elements = [];
+
+    foreach ($items as $delta => $item) {
+      if (empty($item->start_date) && empty($item->end_date)) {
```
donquixote: You need to swap `&&` to `||` when you negate condition parts.
```suggestion
      if (empty($item->start_date) || empty($item->end_date)) {
```
The code below does not work well if only one of the dates is set.
(I know this is a negated form of the condition found in formatters in datetime_range module, so you can use early continue and reduce indentation levels.)

modules/oe_whitelabel_starter_event/src/Plugin/Field/FieldFormatter/EventDateRangeFormatter.php
```diff
@@ -0,0 +1,114 @@
...
+
+      /** @var \Drupal\Core\Datetime\DrupalDateTime $start_date */
+      $start_date = $item->start_date;
+      /** @var \Drupal\Core\Datetime\DrupalDateTime $end_date */
+      $end_date = $item->end_date;
+
+      // Event on a single day.
+      if ($start_date->format('Y-m-d') === $end_date->format('Y-m-d')) {
+        $elements[$delta] = [
+          'start_date' => $this->buildCustomDate($start_date, 'l d F Y, H.i'),
```
donquixote: If we want to implement the pdf specs for date formats, we need translatable date formats.
These hard-coded date format strings cannot accomplish this.

modules/oe_whitelabel_starter_event/src/Plugin/Field/FieldFormatter/EventDateRangeFormatter.php
```diff
@@ -0,0 +1,114 @@
...
+        continue;
+      }
+
+      /** @var \Drupal\Core\Datetime\DrupalDateTime $start_date */
+      $start_date = $item->start_date;
+      /** @var \Drupal\Core\Datetime\DrupalDateTime $end_date */
+      $end_date = $item->end_date;
+
+      // Event on a single day.
+      if ($start_date->format('Y-m-d') === $end_date->format('Y-m-d')) {
```
donquixote: I like to have these comments inside the `if ()` block, if they describe the state if the condition is met.
```suggestion
      if ($start_date->format('Y-m-d') === $end_date->format('Y-m-d')) {
        // The event is on a single day.
```
or
```suggestion
      if ($start_date->format('Y-m-d') === $end_date->format('Y-m-d')) {
        // Start and end date are on the same day.
```

donquixote: The condition uses a different timezone than the actual date format.
This means that a date range can appear to be on separate days for the condition, but then the printed dates are the same.
E.g. I got this: "Wednesday 08 June 2022, 00.04 - Wednesday 08 June 2022, 19.14 (CEST)".

modules/oe_whitelabel_starter_event/src/Plugin/Field/FieldFormatter/EventDateRangeFormatter.php
```diff
@@ -0,0 +1,114 @@
...
+        $elements[$delta] = [
+          'start_date' => $this->buildCustomDate($start_date, 'l d F Y, H.i'),
+          'separator' => ['#plain_text' => '-'],
+          'end_date' => $this->buildCustomDate($end_date, 'H.i (T)'),
+        ];
+      }
+      // Event on multiple days.
+      else {
```
donquixote: Comment can be inside the `else` block, to describe the known state after the check.
```suggestion
      else {
        // Start day and end day are different in the displayed time zone.
```

modules/oe_whitelabel_starter_event/src/Plugin/Field/FieldFormatter/EventDateRangeFormatter.php
donquixote: See how comments outside and inside the if block use different verb forms. Outside you would say "Check if ...", inside you can make simple "is" statements.
```php
// Check if the sky is blue.
if ($sky->getColor() === 'blue') {
  // The sky is blue.
}
else {
  // The sky is not blue.
}
```

modules/oe_whitelabel_starter_event/src/Plugin/Field/FieldFormatter/EventDateRangeFormatter.php
```diff
@@ -0,0 +1,114 @@
...
+        $elements[$delta] = [
+          'start_date' => $this->buildCustomDate($start_date, 'l d F Y, H.i'),
+          'separator' => ['#plain_text' => '-'],
```
donquixote: This separator, unlike the one for single day, must have spaces, according to the pdf.
```suggestion
          'separator' => ['#plain_text' => ' - '],
```

modules/oe_whitelabel_starter_event/src/Plugin/Field/FieldFormatter/EventDateRangeFormatter.php
donquixote: Actually all we would need is to call `$this->setTimeZone($start_date)` and `$this->setTimeZone($end_date)` before we go into this condition.

modules/oe_whitelabel_starter_event/src/Plugin/Field/FieldFormatter/EventDateRangeFormatter.php
```diff
@@ -0,0 +1,114 @@
...
+
+    foreach ($items as $delta => $item) {
+      if (empty($item->start_date) && empty($item->end_date)) {
+        continue;
+      }
+
+      /** @var \Drupal\Core\Datetime\DrupalDateTime $start_date */
+      $start_date = $item->start_date;
+      /** @var \Drupal\Core\Datetime\DrupalDateTime $end_date */
+      $end_date = $item->end_date;
```
donquixote: Btw,
I know this is coming from core, but:
By getting the original datetime objects from the field and then calling `$this->setTimeZone()`, we are polluting the field data in the loaded entity object, changing the time zone. This won't generally be a problem in practice, but still kinda bad. See https://www.drupal.org/project/drupal/issues/3288055.
We could fix it here by cloning the date objects, but let's ignore it for now.

modules/oe_whitelabel_starter_event/src/Plugin/Field/FieldFormatter/EventDateRangeFormatter.php
donquixote: We would need 3 configurable date formats:
- Date only.
- Time only.
- Date + time. This could be a combination of the above, but not all languages use a comma to separate the two parts, so it would be difficult.

We can have 3 configurable options in settingsForm(). This way we don't have to hard-code anything in the formatter.

tests/src/Functional/ContentEventRenderTest.php
```diff
@@ -130,6 +130,24 @@ public function testEventPage(): void {
...
+      'value' => '2022-02-15T08:00:00',
+      'end_value' => '2022-02-22T18:00:00',
+    ]);
+    $node->save();
+
+    $this->drupalGet('node/' . $node->id());
+
+    $this->assertSession()->pageTextMatches(
+      '/[a-zA-Z]+\\s[0-9]+\\s[a-zA-Z]+\\s2022,\\s[0-9]*\\.[0-9]+-[a-zA-Z]+\\s[0-9]+\\s[a-zA-Z]+\\s2022,\\s[0-9]*\\.[0-9]+\\s\\([a-zA-Z]+\\)/i'
+    );
```
donquixote: Why regex?
I think there are no moving parts here, so the output should be predictable.
The only moving part _might be_ the visitor timezone. Not sure which timezone is used in tests. If this is not predictable, we should set a timezone that is different from UTC. See e.g. `Drupal\Tests\datetime\Functional\DateTestBase::setSiteTimezone()`.

donquixote (changes requested):

modules/oe_whitelabel_starter_event/src/Plugin/Field/FieldFormatter/EventDateRangeFormatter.php
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * @FieldFormatter(
+ *   id = "event_date_range_format",
+ *   label = @Translation("Event date range format"),
+ *   field_types = {
+ *     "daterange"
+ *   }
+ * )
+ */
+class EventDateRangeFormatter extends DateTimeFormatterBase {
```
donquixote: I forgot to mention: I don't think this formatter is specific to events.
We should change the plugin id and the class name, and also move it into oe_whitelabel_helper. Possibly even oe_bootstrap_theme_helper - TBD.
Basically the formatter does this:
- If same day, it shows `$start_date->format($day_format) . $separator0 . $start_time->format($time_format) . $time_separator . $end_time->format($time_format)`.
- If different day, it shows `$start_date->format($datetime_format) . $day_separator . $end_date->format($datetime_format)`.

The formats and the separators should be configurable and translatable.

This sounds quite generic, and is similar to what e.g. https://www.drupal.org/project/daterange_compact promises to do. (I am not saying we should use that contrib module, I just want to show this use case is quite generic.)

modules/oe_whitelabel_starter_event/config/install/core.entity_view_display.node.oe_sc_event.teaser.yml
```diff
@@ -54,3 +54,4 @@ hidden:
   oe_content_short_title: true
   oe_documents: true
   oe_featured_media: true
+  search_api_excerpt: true
```
donquixote, line 55: And another thing I forgot to mention.
We are changing the appearance of date range teaser, but we are not changing the formatter or showing the time.
Do we actually have a spec for this?

escuriola: The Event teaser uses short dates in the PDF, but instead of "to" the separator is "-". For this reason I didn't change anything.

modules/oe_whitelabel_starter_event/src/Plugin/Field/FieldFormatter/EventDateRangeFormatter.php
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    return $elements;
+  }
+
+  /**
+   * Prepare render for a date with custom format.
+   */
+  protected function buildCustomDate(DrupalDateTime $date, $format) {
+    $this->setTimeZone($date);
```
donquixote: Now we no longer need to call `->setTimeZone()` here.

modules/oe_whitelabel_starter_event/src/Plugin/Field/FieldFormatter/EventDateRangeFormatter.php
```diff
@@ -0,0 +1,148 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function settingsForm(array $form, FormStateInterface $form_state) {
+    $form['date_format'] = [
+      '#type' => 'textfield',
+      '#title' => $this->t('Date format'),
+      '#description' => $this->t('See <a href="https://www.php.net/manual/datetime.format.php#refsect1-datetime.format-parameters" target="_blank">the documentation for PHP date formats</a>.'),
+      '#default_value' => $this->getSetting('date_format'),
+    ];
```
donquixote: As you have shown, this one is not really needed :)
The combo is just start|datetime + '-' + end|time, and start|datetime + ' - ' + end|datetime.
So we only need 2 options.

donquixote: Ofc the idea was to use translatable formats. TBD.

modules/oe_whitelabel_starter_event/src/Plugin/Field/FieldFormatter/EventDateRangeFormatter.php
donquixote: Actually we need just 2, as Carlos has already shown :)

tests/src/Functional/ContentEventRenderTest.php
```diff
@@ -130,6 +136,24 @@ public function testEventPage(): void {
       trim($content_banner->filter('.oe-sc-event__oe-summary')->text())
     );
 
+    $date = $crawler->filter('dd');
```
donquixote: The selector appears bizarre to me.
It seems to work, but don't we have something better?

modules/oe_whitelabel_starter_event/config/install/core.entity_view_display.node.oe_sc_event.teaser.yml
donquixote: I do see the "Event teaser" part in the pdf now. It does use hyphen with spaces.
I see that we don't need to add the spaces with the separator here, the core formatter already adds the spaces.
So it seems ok.

Can we add a line in ContentEventRenderTest::testEventRenderingTeaser() to test the date output? I think we currently don't have it.

modules/oe_whitelabel_starter_event/src/Plugin/Field/FieldFormatter/EventDateRangeFormatter.php
donquixote: For translatable formats, just replicate what you find in the core date formatters.

donquixote (changes requested):

modules/oe_whitelabel_helper/config/install/daterange_compact.format.oe_whitelabel_date_time_long.yml
```diff
@@ -0,0 +1,16 @@
+langcode: en
+status: true
+dependencies: {  }
+id: oe_whitelabel_date_time_long
+label: 'Date and time, long names'
+default_pattern: 'l d F Y, H.i (T)'
+default_separator: ' - '
+same_day_start_pattern: 'l d F Y, H.i'
+same_day_end_pattern: 'H.i (T)'
+same_day_separator: '-'
```
donquixote, line 10: This is different from the value in the update hook config.

modules/oe_whitelabel_helper/config/post_updates/00002_date_range_formats/daterange_compact.format.oe_whitelabel_date_time_long.yml
```diff
@@ -0,0 +1,16 @@
+langcode: en
+status: true
+dependencies: {  }
+id: oe_whitelabel_date_time_long
+label: 'Date and time, long names'
+default_pattern: 'l d F Y, H.i (T)'
+default_separator: ' - '
+same_day_start_pattern: 'l d F Y, H.i'
+same_day_end_pattern: 'H.i (T)'
+same_day_separator: ' - '
```
donquixote:
```suggestion
same_day_separator: '-'
```
Align with config/install.
According to the pdf there are no spaces on same day.

composer.json
```diff
@@ -9,6 +9,7 @@
         "php": ">=7.4",
         "cweagans/composer-patches": "^1.7",
         "drupal/core": "^9.2",
+        "drupal/daterange_compact": "^2.0",
```
donquixote, line 12: The module seems like a good idea, but: Are these formats translatable?

I found this issue: https://www.drupal.org/project/daterange_compact/issues/3023808

> Thanks for the contributions. Formats are translatable in version 2.0, using a combination of #3 and #7.

But how do we translate them?

modules/oe_whitelabel_helper/config/install/daterange_compact.format.oe_whitelabel_date_only_short_month.yml
```diff
@@ -0,0 +1,16 @@
+langcode: en
+status: true
+dependencies: {  }
+id: oe_whitelabel_date_only_short_month
+label: 'Date only, short month'
+default_pattern: 'd M Y'
+default_separator: ' - '
+same_day_start_pattern: 'd M Y'
```
donquixote: I don't see a leading zero in the pdf for other date formats. So I assume we also don't want it here.
```suggestion
same_day_start_pattern: 'j M Y'
```

modules/oe_whitelabel_helper/config/install/daterange_compact.format.oe_whitelabel_date_only_short_month.yml
```diff
@@ -0,0 +1,16 @@
+langcode: en
+status: true
+dependencies: {  }
+id: oe_whitelabel_date_only_short_month
+label: 'Date only, short month'
+default_pattern: 'd M Y'
```
donquixote:
```suggestion
default_pattern: 'j M Y'
```

modules/oe_whitelabel_helper/config/install/daterange_compact.format.oe_whitelabel_date_time_long.yml
```diff
@@ -0,0 +1,16 @@
+langcode: en
+status: true
+dependencies: {  }
+id: oe_whitelabel_date_time_long
+label: 'Date and time, long names'
+default_pattern: 'l d F Y, H.i (T)'
+default_separator: ' - '
+same_day_start_pattern: 'l d F Y, H.i'
```
donquixote:
```suggestion
same_day_start_pattern: 'l j F Y, H.i'
```

modules/oe_whitelabel_helper/config/install/daterange_compact.format.oe_whitelabel_date_time_long.yml
```diff
@@ -0,0 +1,16 @@
+langcode: en
+status: true
+dependencies: {  }
+id: oe_whitelabel_date_time_long
+label: 'Date and time, long names'
+default_pattern: 'l d F Y, H.i (T)'
```
donquixote: I see no leading zero in the pdf.
> Tuesday 8 February 2022, 14.30 - Wednesday 9 February 2022, 16.30
```suggestion
default_pattern: 'l j F Y, H.i (T)'
```

composer.json
donquixote: If it is currently not fully possible, perhaps we can contribute in the future.

modules/oe_whitelabel_starter_event/src/Plugin/Field/FieldFormatter/EventDateRangeFormatter.php
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function formatDate($date) {
+    return $this->dateFormatter->format($date->getTimestamp(), 'custom', 'Y-m-d');
+  }
+
+}
```
donquixote: We can remove this formatter.

modules/oe_whitelabel_starter_event/config/schema/oe_whitelabel_starter_event.schema.yml
```diff
@@ -0,0 +1,10 @@
+field.formatter.settings.event_date_range_format:
+  type: mapping
+  label: 'Date Time range format'
+  mapping:
+    time_format:
+      type: string
+      label: 'Time format'
+    datetime_format:
+      type: string
+      label: 'DateTime format'
```
donquixote: We can remove the schema along with the formatter.

tests/src/Functional/ContentEventRenderTest.php
```diff
@@ -130,6 +80,24 @@ public function testEventPage(): void {
       trim($content_banner->filter('.oe-sc-event__oe-summary')->text())
     );
 
+    $date = $crawler->filter('dl > dd');
+
+    // Assert event dates starting and ending same day.
+    $this->assertEquals('Wednesday 09 February 2022, 21.00-23.00 (CET)', trim($date->text()));
+
+    // Assert event dates starting and ending at different days.
```
donquixote, line 88: These "Assert" comments feel wrong to me, if they are not on top of an actual assertion.
Here we are just setting things up, the assert happens later. ("arrange, act, assert")

tests/src/Functional/ContentEventRenderTest.php
```diff
@@ -130,6 +80,24 @@ public function testEventPage(): void {
...
+      'value' => '2022-02-15T08:00:00',
+      'end_value' => '2022-02-22T18:00:00',
+    ]);
+    $node->save();
+
+    $this->drupalGet($node->toUrl());
+    $crawler = $client->getCrawler();
+
+    $date = $crawler->filter('dl > dd');
+    $this->assertEquals('Tuesday 15 February 2022, 09.00 (CET) - Tuesday 22 February 2022, 19.00 (CET)', trim($date->text()));
```
donquixote: We would not need the trim if we would use spaceless or `-%}` syntax in the template.
The output currently has `<dd> date </dd>` instead of `<dd>date</dd>`.

donquixote: This said: The spaces are not visible in a `<dd>` element, so it's not a real problem.

donquixote (changes requested):
A bit of feedback.

And one question: Should we have tests for the date range formats in isolation, or is it enough to test them as part of a content type?
If we test in isolation we could also cover translatability.

But not a blocker, I would be ok to merge this without these additional tests. The event test will reveal regressions in the date range formats.

composer.json
brummbar: It's available in 2.x afaik.

tests/src/Functional/ContentEventRenderTest.php
brummbar: That's to explain what is this block of code doing.

tests/src/Functional/ContentEventRenderTest.php
brummbar: I don't understand the remark, if there's any. Anyway the trim() is just to be safe, not really needed.

composer.json
donquixote: That's what they say, but can we verify it somehow?
It would be a deal breaker otherwise.

tests/src/Functional/ContentEventRenderTest.php
donquixote: Well currently without the trim the test would fail, because of leftover space from the template.
if you inspect the output in the browser you will see leftover space within the `<dd>`.

We can ignore it and keep the `trim()` here for suppression, or we can use the space eater syntax (`-%}` instead of just `%}` in the template.

```twig
  {% set _event_dates -%}
    {{ content.oe_sc_event_dates|field_value|safe_join(', ') }}
  {%- endset %}
```

brummbar: Ticket is approaching 200% time spent, let's focus on important stuff.

composer.json
brummbar: I did, and it works.

donquixote (approved):

drishu (approved):

PR #159, opened 2022-06-20
OEL-1662: Add registration URL field to the full content view

escuriola:


templates/content/node--oe-sc-event--oe-w-content-banner.html.twig
```diff
@@ -11,4 +11,5 @@
     content: content.oe_summary,
     image: image,
   }) }}
+    {{  content.oe_sc_event_registration_url }}
```
drishu: it should be inside the gray area
image

tests/src/Functional/ContentEventRenderTest.php
```diff
@@ -130,6 +131,12 @@ public function testEventPage(): void {
       trim($content_banner->filter('.oe-sc-event__oe-summary')->text())
     );
 
+    // Assert registration button.
+    $link = $crawler->filter('div.container > div.row > div.col-8 > a[target="_blank"]');
+    $this->assertCount(1, $link);
```
drishu, line 85: below also assert that there is a button inside the link, not just the button text

modules/oe_whitelabel_starter_event/config/install/core.entity_view_display.node.oe_sc_event.oe_w_content_banner.yml
```diff
@@ -27,6 +29,18 @@ content:
     third_party_settings: {  }
     weight: 1
     region: content
+  oe_sc_event_registration_url:
+    type: event_link_external
```
drishu: the formatter should be defined in oe_whitelabel_helper and then its machine name will be different

drishu (changes requested):

templates/content/field--node--oe-sc-event-registration-url.html.twig
```diff
@@ -0,0 +1,20 @@
...
+    'label': 'Register'|t,
+    'icon': 'calendar-check',
+    'icon_position': 'before',
+  }) }}
+{% endset %}
+{% set _attributes = element[0]['#options']['attributes'] %}
+{{ pattern('link', {
+  'label': button,
+  'path': element[0]['#url'].toString(),
+  'attributes': _attributes,
```
drishu: add margin top mt-3

drishu (changes requested):
I wont ask for a dedicated test for the formatter since we are out of time

templates/content/field--node--oe-sc-event-registration-url.html.twig
escuriola: Adding the mt-3 there was any result. I added a "mt-4" class

drishu (dismissed):

modules/oe_whitelabel_helper/src/Plugin/Field/FieldFormatter/ExternalLinkFormatter.php
```diff
@@ -0,0 +1,40 @@
...
+namespace Drupal\oe_whitelabel_helper\Plugin\Field\FieldFormatter;
+
+use Drupal\Core\Field\FieldItemListInterface;
+use Drupal\link\Plugin\Field\FieldFormatter\LinkFormatter;
+
+/**
+ * Plugin implementation of the 'external_link' formatter.
+ *
+ * @FieldFormatter(
+ *   id = "external_link",
```
donquixote: Please prefix this with 'oe_whitelabel_'.

donquixote:
```suggestion
 *   id = "oe_whitelabel_external_link",
```

templates/content/field--node--oe-sc-event-registration-url.html.twig
```diff
@@ -0,0 +1,21 @@
...
+    {{ pattern('button', {
+      'label': 'Register'|t,
+      'icon': 'calendar-check',
+      'icon_position': 'before',
+    }) }}
+  {% endset %}
+  {% set _attributes = element[0]['#options']['attributes'] %}
+  {{ pattern('link', {
+    'label': button,
+    'path': element[0]['#url'].toString(),
```
donquixote: Note: We lose any cache info this way, _but_ it is ok, there is no cache info in this field.

donquixote: Note: Also we would normally go through a for loop to process field items, but I think it is fine: The template won't be called if the field is empty, and the first item is always at `[0]`. So ok for me.

templates/content/field--node--oe-sc-event-registration-url.html.twig
```diff
@@ -0,0 +1,21 @@
...
+    'attributes': _attributes,
+  }) }}
+</div>
```
donquixote: Note:
I notice that in netlify the button classes are directly on the link tag.
I find this in `src/compositions/bcl-event/data/data-page.js`:
```
    links: [
      {
        label: "Register",
        icon_position: "before",
        icon: {
          name: "calendar-check",
          path: "/icons.svg",
        },
        attributes: new drupalAttribute().addClass([
          "btn",
          "btn-primary",
          "d-block",
          "d-md-inline-block",
        ]),
      },
    ],
```
So their solution would require to pass explicit classes to the content banner pattern or bcl component, which we want to avoid. So ok for this solution.

donquixote: Even better would be a bcl component or pattern for "button link". But not today.

templates/content/field--node--oe-sc-event-registration-url.html.twig
```diff
@@ -0,0 +1,21 @@
...
+#}
+<div class="mt-4">
+  {% set button %}
+    {{ pattern('button', {
+      'label': 'Register'|t,
+      'icon': 'calendar-check',
+      'icon_position': 'before',
+    }) }}
+  {% endset %}
+  {% set _attributes = element[0]['#options']['attributes'] %}
```
donquixote: Note:
We could inline both of these variables, but we don't have to.

```twig
  {{ pattern('link', {
    'label': pattern('button', {
      'label': 'Register'|t,
      'icon': 'calendar-check',
      'icon_position': 'before',
    }),
    'path': element[0]['#url'].toString(),
    'attributes': element[0]['#options']['attributes'],
  }) }}
```

templates/content/node--oe-sc-event--oe-w-content-banner.html.twig
```diff
@@ -4,11 +4,15 @@
...
+{% set _content %}
+  {{ content.oe_summary }}
+  {{ content.oe_sc_event_registration_url }}
+{% endset %}
 <div{{attributes}}>
   {{ pattern('content_banner', {
     background: 'gray',
     title: label,
-    content: content.oe_summary,
+    content: _content,
```
donquixote: Note:
We _could_ inline this using render element composition:
```twig
  {{ pattern('content_banner', {
    background: 'gray',
    title: label,
    content: [
      content.oe_summary,
      content.oe_sc_event_registration_url,
    ],
    image: image,
  }) }}
```
But let's keep it as it is for now.

tests/src/Functional/ContentEventRenderTest.php
```diff
@@ -130,6 +131,13 @@ public function testEventPage(): void {
       trim($content_banner->filter('.oe-sc-event__oe-summary')->text())
     );
 
+    // Assert registration button.
+    $link = $crawler->filter('.bcl-content-banner a[target="_blank"]');
+    $this->assertCount(1, $link);
+    $button = $link->filter('button');
+    $this->assertStringContainsString('Register', $button->text());
+    $this->assertStringContainsString('calendar-check', $button->html());
+
```
donquixote, line 104: Given that we already have `$link` as a separate variable, we should also assert the href attribute.

donquixote: And we should test one version with an internal link.
The test in oe_starter_content does not count for this because we don't have the formatter there.

donquixote (changes requested):
Some notes, and some changes to be done.

modules/oe_whitelabel_helper/src/Plugin/Field/FieldFormatter/ExternalLinkFormatter.php
```diff
@@ -0,0 +1,40 @@
...
+
+      if ($url->isExternal()) {
+        $element[$delta]['#options']['attributes']['target'] = '_blank';
+      }
+    }
+
+    return $element;
+  }
+
+}
```
donquixote: @escuriola 
In the other PR (https://github.com/openeuropa/oe_starter_content/pull/11#discussion_r903721591), @drishu mentioned we should have a dedicated test for the formatter.
The ContentEventRenderTest also indirectly covers the formatter, but the result is tampered by our field template.
A dedicated test (kernel ideally) is the right thing to do. This is for projects that use the formatter in other places, without overriding the field template.

donquixote: Btw, I wonder if this formatter belongs into oe_bootstrap_theme. Or perhaps it could live in a separate module.
But I think it was decided prior that we want to have it here, so I am ok with that.

modules/oe_whitelabel_helper/src/Plugin/Field/FieldFormatter/ExternalLinkFormatter.php
```diff
@@ -0,0 +1,40 @@
...
+   * {@inheritdoc}
+   */
+  public function viewElements(FieldItemListInterface $items, $langcode) {
+    $element = parent::viewElements($items, $langcode);
+
+    foreach ($items as $delta => $item) {
+      $url = $this->buildUrl($item);
+
+      if ($url->isExternal()) {
+        $element[$delta]['#options']['attributes']['target'] = '_blank';
```
donquixote: I just found these two modules.
https://www.drupal.org/project/external_links_new_tab
https://www.drupal.org/project/sel

One interesting thing from the module pages:

> If you simply just have a target="_blank" on outbound links, you put your site visitors at risk of tabnapping.

> Additionally, it also adds rel="noopener" to external links to improve performance and prevent security vulnerabilities.

On the other hand, here I find this https://mathiasbynens.github.io/rel-noopener/

> 2021 update: Browsers now implicitly set rel=noopener for any target=_blank link, following a spec change. If the demo on this page no longer seems scary, congratulations — you’re using a modern browser!

donquixote (changes requested):

drishu (dismissed):

modules/oe_whitelabel_helper/src/Plugin/Field/FieldFormatter/ExternalLinkFormatter.php
donquixote: Let's keep `_blank` for now, and do a follow-up if we really need to add `noopener`.

modules/oe_whitelabel_helper/src/Plugin/Field/FieldFormatter/ExternalLinkFormatter.php
donquixote: Looks like we no longer have a custom formatter for this.
Resolving this.

templates/content/field--node--oe-sc-event-registration-url.html.twig
donquixote: I think I am wrong about cache info :) We are in the field template, so any cache info on the field was already processed.
As for the loop, this was done - good!

donquixote (approved):

drishu (dismissed):

modules/oe_whitelabel_starter_event/config/post_updates/00003_add_registration_url_field/core.entity_view_display.node.oe_sc_event.full.yml
```diff
@@ -0,0 +1,72 @@
...
+    - field.field.node.oe_sc_event.body
+    - field.field.node.oe_sc_event.oe_documents
+    - field.field.node.oe_sc_event.oe_featured_media
+    - field.field.node.oe_sc_event.oe_sc_event_dates
+    - field.field.node.oe_sc_event.oe_sc_event_location
+    - field.field.node.oe_sc_event.oe_sc_event_registration_url
+    - field.field.node.oe_sc_event.oe_summary
+    - node.type.oe_sc_event
+  module:
+    - datetime_range
```
donquixote: This is no longer in sync with the configs in config/install.

modules/oe_whitelabel_starter_event/config/post_updates/00003_add_registration_url_field/core.entity_view_display.node.oe_sc_event.full.yml
```diff
@@ -0,0 +1,72 @@
...
+    type: entity_reference_entity_view
+    label: above
+    settings:
+      view_mode: default
+      link: true
+    third_party_settings: {  }
+    weight: 3
+    region: content
+  oe_sc_event_dates:
+    type: daterange_default
```
donquixote: This is no longer in sync with the configs in config/install.

modules/oe_whitelabel_starter_event/config/post_updates/00003_add_registration_url_field/core.entity_view_display.node.oe_sc_event.full.yml
```diff
@@ -0,0 +1,72 @@
...
+    label: inline
+    settings:
+      timezone_override: ''
+      format_type: oe_whitelabel_starter_event_date
+      separator: to
```
donquixote: This is no longer in sync with the configs in config/install.

modules/oe_whitelabel_starter_event/config/post_updates/00003_add_registration_url_field/core.entity_view_display.node.oe_sc_event.teaser.yml
```diff
@@ -0,0 +1,59 @@
...
+    - datetime_range
+    - text
+    - user
+id: node.oe_sc_event.teaser
+targetEntityType: node
+bundle: oe_sc_event
+mode: teaser
+content:
+  oe_sc_event_dates:
+    type: daterange_default
```
donquixote: This is no longer in sync with the configs in config/install.

donquixote (changes requested):
Please sync the config files in update hook with those in config/install.

donquixote (approved):

drishu (approved):

PR #161, opened 2022-06-29
OEL-1638: Check variable items has facets before trying to modify them to avoid warning.

escuriola:


brummbar (dismissed):

modules/oe_whitelabel_search/oe_whitelabel_search.module
```diff
@@ -42,6 +42,9 @@ function oe_whitelabel_search_form_facets_form_alter(&$form, FormStateInterface
 function oe_whitelabel_search_preprocess_facets_summary_item_list(array &$variables): void {
   $items = [];
   foreach ($variables['items'] as $item) {
+    if (empty($item['value']['#type'])) {
```
drishu: I know this works but what we want to check if it is set, not if it is empty, empty checks for more things than we want to
```suggestion
    if (!isset($item['value']['#type'])) {
```

drishu (changes requested):

drishu (approved):

brummbar (approved):

PR #162, opened 2022-06-30
OEL-1552: Theme the person content type

escuriola:


templates/content/node--oe-sc-person--full.html.twig
```diff
@@ -0,0 +1,20 @@
+{#
+/**
+ * @file
+ * Person full display.
+ */
+#}
+<article{{attributes}}>
+  <div class="row">
+    <div class="col-12 col-lg-10 col-xl-9 col-xxl-8">
+      <h2 class="mb-4">{{ 'Additional information'|t }}</h2>
```
drishu: if oe_sc_person_additional_info is empty this title still shows

templates/content/node--oe-sc-person--full.html.twig
```diff
@@ -0,0 +1,20 @@
...
+      <div class="mb-4-5">
+        {{ content.oe_sc_person_additional_info|field_value }}
+      </div>
+      <h2 class="mb-4">{{ 'Related documents'|t }}</h2>
```
drishu: if oe_sc_person_additional_info is empty this title still shows

templates/content/node--oe-sc-person--oe-w-content-banner.html.twig
```diff
@@ -0,0 +1,41 @@
+{#
+/**
+ * @file
+ * Person content banner display.
+ */
+#}
+
+{% set _meta = [] %}
+{% if not node.oe_sc_person_position.isEmpty() %}
+  {% set _meta = _meta|merge([content.oe_sc_person_position|field_value]) %}
```
drishu: Let's put occupation before position like in netlify

templates/content/node--oe-sc-person--teaser.html.twig
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * @file
+ * Theme override to display a node of bundle person in the teaser view mode.
+ */
+#}
+{% set _title %}
+  <a class="standalone" href="{{ url }}">{{ label }}</a>
+{% endset %}
+{% set _meta = [] %}
+{% if not node.oe_sc_person_position.isEmpty() %}
```
drishu: again in netlify position is the second item

templates/content/node--oe-sc-person--full.html.twig
```diff
@@ -0,0 +1,20 @@
...
+      <div class="mb-4-5">
+        {{ content.oe_sc_person_documents|field_value }}
+      </div>
+    </div>
+  </div>
```
drishu, line 13: do here a content|without(..) in case subsites add other fields and for cache bubbling

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
```diff
@@ -0,0 +1,84 @@
...
+    $cacheability->applyTo($variables);
+    return;
+  }
+
+  // Get the media source.
+  $source = $media->getSource();
+
+  $is_image = $source instanceof MediaAvPortalPhotoSource || $source instanceof Image;
+
+  // If it's not an image and not a video, bail out.
```
drishu:
```suggestion
  // If it's not an image bail out.
```

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
```diff
@@ -0,0 +1,84 @@
...
+  if (!$is_image) {
+    $cacheability->applyTo($variables);
+    return;
+  }
+
+  $thumbnail = $media->get('thumbnail')->first();
+  $variables['image'] = ImageValueObject::fromImageItem($thumbnail);
+
+  $social_links = $node->get('oe_social_media_links')->getValue();
```
drishu, line 95: you have a series of bail outs for image, that means this code is never reached in case it bails out for the image, these two things are not related and social links should not depend on the image.

templates/content/node--oe-sc-person--oe-w-content-banner.html.twig
```diff
@@ -0,0 +1,41 @@
...
+        }
+      ) %}
+      <div class="mb-3 mb-md-0 d-block d-md-inline-block me-md-4-5">
+        {{ pattern('link', _link) }}
+      </div>
+    {% endfor %}
+  </div>
+  </div>
+{% endset %}
+<div{{ attributes }}>
```
drishu, line 23: this div is not closed

templates/content/node--oe-sc-person--oe-w-content-banner.html.twig
```diff
@@ -0,0 +1,41 @@
...
+        {
+          'attributes': create_attribute().addClass(['d-block', 'd-md-inline-block', 'mb-3', 'mb-md-0', 'standalone'])
+        }
+      ) %}
+      <div class="mb-3 mb-md-0 d-block d-md-inline-block me-md-4-5">
+        {{ pattern('link', _link) }}
+      </div>
+    {% endfor %}
+  </div>
+  </div>
```
drishu: what is this closing ?

tests/src/Functional/ContentPersonRenderTest.php
```diff
@@ -0,0 +1,222 @@
...
+  protected static $modules = [
+    'block',
+    'oe_whitelabel_starter_person',
+  ];
+
+  /**
+   * Tests the person page.
+   */
+  public function testPersonPage(): void {
+    $node = $this->createExamplePerson();
```
drishu: lets do first a node with just the required fields, assert that labels (and any other elements?) that should not show with empty values don't.

tests/src/Functional/ContentPersonRenderTest.php
```diff
@@ -0,0 +1,222 @@
...
+      trim($image->attr('src'))
+    );
+    $this->assertEquals('Starter Image test alt',
+      $image->attr('alt')
+    );
+
+    // Assert content banner content elements.
+    $this->assertEquals(
+      'Stefan Mayer',
+      trim($content_banner->filter('.card-title')->text())
```
drishu: I tried without all these trims and it passed

drishu: I tried without all these trims and it passed

templates/content/field--oe-document-reference--oe-document--oe-document.html.twig
```diff
@@ -0,0 +1,11 @@
+{#
```
drishu, line 1: why not field--oe-document-reference--oe-document.html.twig ?

templates/content/field--oe-document-reference--oe-document-group.html.twig
```diff
@@ -0,0 +1,17 @@
+{#
+/**
+ * @file
+ * Field template of the document group reference entity.
+ *
+ * @see ./core/themes/stable/templates/field/field.html.twig
+ */
+#}
+{% for item in items %}
+  {% if item.content['#media'] is empty %}
```
drishu: don't like this but dont have a better idea right now, next reviewer ?

drishu (changes requested):

composer.json
```diff
@@ -93,7 +93,14 @@
         "patches": {
```
drishu, line 95: One more thing i noticed just now, when I build composer I get a new folder
image
pls look into it

templates/content/node--oe-sc-person--full.html.twig
escuriola: I'm not sure which argument I must add at the without. Could you help me with this @drishu?

composer.json
escuriola: image

templates/content/node--oe-sc-person--full.html.twig
```diff
@@ -0,0 +1,24 @@
+{#
+/**
+ * @file
+ * Person full display.
+ */
+#}
+<article{{attributes}}>
+  <div class="row">
+    <div class="col-12 col-lg-10 col-xl-9 col-xxl-8">
+      {% if content.oe_sc_person_additional_info|field_value is not empty %}
```
drishu:
```
 content.oe_sc_person_additional_info.isEmpty() 
```
does not work here ?

templates/content/node--oe-sc-person--full.html.twig
drishu:
```
|without('oe_sc_person_additional_info', 'oe_sc_person_documents', 'oe_sc_person_occupation', 'oe_sc_person_position', 'oe_sc_person_country', 'oe_summary', ..
```
any field you print in the templates should listed.

escuriola: Ok! Understood!

templates/content/node--oe-sc-person--full.html.twig
escuriola: not working

drishu: yes its node.oe_sc_person_additional_info.isEmpty()

templates/content/node--oe-sc-person--full.html.twig
drishu: My indication was wrong, only the ones printed in that template.
But anyway, I checked, I found no cache info attached to content, so I will remove this commit.

drishu (dismissed):

composer.json
```diff
@@ -94,7 +94,14 @@
...
             "drupal/entity_reference_revisions": {
                 "https://www.drupal.org/project/entity_reference_revisions/issues/2937835": "https://www.drupal.org/files/issues/2021-03-26/entity_reference_revisions-field_formatter_label-2937835-36.patch"
+            },
+            "openeuropa/oe_starter_content": {
+                "OEL-1551-Document_fields_for_person": "https://github.com/openeuropa/oe_starter_content/compare/1.0.0-beta1...EPIC-OEL-1548-Person-v1.diff"
+            },
+            "openeuropa/oe_content": {
+                "OEL-1551-Document_fields_for_person": "https://github.com/openeuropa/oe_content/compare/3.x...OEL-1551.diff"
```
drishu: patch is against 3.x but we use 2.8, so, we should use 3.0.0-beta3, and patch https://github.com/openeuropa/oe_content/compare/3.0.0-alpha3...OEL-1551.diff
But this patch contains composer changes, so best to use the branch instead of patch.

composer.json
drishu: we will probably gitignore

drishu (dismissed):

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
```diff
@@ -0,0 +1,86 @@
...
+ * Implements template_preprocess_node() for the Person node type.
+ */
+function oe_whitelabel_starter_person_preprocess_node__oe_sc_person(&$variables) {
+  if ($variables['view_mode'] !== 'oe_w_content_banner' && $variables['view_mode'] !== 'teaser') {
+    return;
+  }
+
+  // Caches are handled by the formatter usually. Since we are not rendering
+  // the original render arrays, we need to propagate our caches to the
+  // paragraph template.
```
donquixote: I think you mean the node template, not the paragraph template.
And, do we have to copy this everywhere?

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
```diff
@@ -0,0 +1,86 @@
...
+
+use Drupal\Core\Cache\CacheableMetadata;
+use Drupal\Core\Url;
+use Drupal\media\MediaInterface;
+use Drupal\media\Plugin\media\Source\Image;
+use Drupal\media_avportal\Plugin\media\Source\MediaAvPortalPhotoSource;
+use Drupal\oe_bootstrap_theme\ValueObject\ImageValueObject;
+
+/**
+ * Implements template_preprocess_node() for the Person node type.
```
donquixote:
```suggestion
 * Implements hook_preprocess_HOOK() for the Person content type.
```

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
```diff
@@ -0,0 +1,86 @@
...
+  $variables['social_links'] = [];
+  foreach ($social_links as $key => $link) {
+    $variables['social_links'][$key]['icon_position'] = 'before';
+    $variables['social_links'][$key]['icon']['path'] = $variables['bcl_icon_path'];
+    $variables['social_links'][$key]['icon']['name'] = $link['link_type'];
+    $variables['social_links'][$key]['label'] = $link['title'];
+    $variables['social_links'][$key]['path'] = Url::fromUri($link['uri'])->toString();
+  }
+
+  $cacheability->applyTo($variables);
```
donquixote: This seems useless here. We have not added anything to the `$cacheability` yet.

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
```diff
@@ -0,0 +1,86 @@
...
+
+  $social_links = $node->get('oe_social_media_links')->getValue();
+
+  $variables['social_links'] = [];
+  foreach ($social_links as $key => $link) {
+    $variables['social_links'][$key]['icon_position'] = 'before';
+    $variables['social_links'][$key]['icon']['path'] = $variables['bcl_icon_path'];
+    $variables['social_links'][$key]['icon']['name'] = $link['link_type'];
+    $variables['social_links'][$key]['label'] = $link['title'];
+    $variables['social_links'][$key]['path'] = Url::fromUri($link['uri'])->toString();
```
donquixote:
```suggestion
    $variables['social_links'][$key] = [
      'icon_position' => 'before',
      'icon' => [
        'path' => $variables['bcl_icon_path'],
        'name' => $link['link_type'],
      ],
      'label' => $link['title'],
      'path' => Url::fromUri($link['uri'])->toString(),
    ];
```

donquixote: (please review if I did everything correctly)

donquixote: And rename `$key` to `$delta`, as this is the common variable name for field item indices.

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
```diff
@@ -0,0 +1,86 @@
...
+use Drupal\media\Plugin\media\Source\Image;
+use Drupal\media_avportal\Plugin\media\Source\MediaAvPortalPhotoSource;
+use Drupal\oe_bootstrap_theme\ValueObject\ImageValueObject;
+
+/**
+ * Implements template_preprocess_node() for the Person node type.
+ */
+function oe_whitelabel_starter_person_preprocess_node__oe_sc_person(&$variables) {
+  if ($variables['view_mode'] !== 'oe_w_content_banner' && $variables['view_mode'] !== 'teaser') {
+    return;
```
donquixote: The teaser needs only the image, whereas the content banner needs image + social links.
in oe_whitelabel_starter_event.module we see how a preprocess can be split up to only

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  // Get the media source.
+  $source = $media->getSource();
+
+  $is_image = $source instanceof MediaAvPortalPhotoSource || $source instanceof Image;
+
+  // If it's not an image bail out.
+  if (!$is_image) {
+    $cacheability->applyTo($variables);
```
donquixote, line 75: Do we need to add `$source` to the `$cacheability`?
Maybe not, because a file does not simply change by itself.
But if we don't, then why do we have to repeat this line 3 times instead of just calling it once outside of the `if`?
(I know it is copied from event or news, where we do the same)

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
donquixote: In fact we don't need `$key` or `$delta` at all, we can just:
```php
  foreach ($social_links as $link) {
    $variables['social_links'][] = [
```

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
```diff
@@ -0,0 +1,86 @@
...
+ */
+function oe_whitelabel_starter_person_preprocess_node__oe_sc_person(&$variables) {
+  if ($variables['view_mode'] !== 'oe_w_content_banner' && $variables['view_mode'] !== 'teaser') {
+    return;
+  }
+
+  // Caches are handled by the formatter usually. Since we are not rendering
+  // the original render arrays, we need to propagate our caches to the
+  // paragraph template.
+  $cacheability = CacheableMetadata::createFromRenderArray($variables);
```
donquixote: We don't need `$cacheability` up here. Move it down to where we actually use it.

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
donquixote: Actually the comment is not really useful or relevant, let's drop it.

templates/content/field--oe-document-reference--oe-document-group.html.twig
```diff
@@ -0,0 +1,17 @@
+{#
+/**
+ * @file
+ * Field template of the document group reference entity.
```
donquixote: I would like to see the different parts of the template name mentioned here more explicitly:
- Entity type = 'oe_document_reference'
- Bundle = 'oe_document_group'?
- Field name = (any field in the bundle)
  - Currently this template applies to all fields of oe_document_group bundle. What about splitting it up? Or is there a limit on template name length? `field--oe-document-reference--oe-documents--oe-document-group.html.twig` and `field--oe-document-reference--oe-title--oe-document-group.html.twig`.

templates/content/field--oe-document-reference--oe-document-group.html.twig
donquixote: Make two templates out of it, then you don't need this confusing distinction.

donquixote: Alternatively I would have said make a single template for the document reference entity bundle. But it seems there is no entity template for document reference. So we have to stick with the field templates.

templates/content/field--oe-document-reference--oe-document.html.twig
```diff
@@ -0,0 +1,11 @@
+{#
+/**
+ * @file
+ * Field template of the document reference entity.
```
donquixote: Again, we can be more specific here. And perhaps a more targeted template name?
Yes we currently have only one field in this bundle, but by adding entity type + bundle + field name in the template name, it is more easy to see what we are targeting.

templates/content/node--oe-sc-person--full.html.twig
```diff
@@ -0,0 +1,24 @@
+{#
+/**
+ * @file
+ * Person full display.
+ */
+#}
+<article{{attributes}}>
+  <div class="row">
+    <div class="col-12 col-lg-10 col-xl-9 col-xxl-8">
+      {% if not node.oe_sc_person_additional_info.isEmpty() %}
```
donquixote: I don't like this field emptiness check.
What if a field has content, but the visitor has no permission to see it?
This applies even more for the other field where document media might have its own access control.

For other content types we have used different ways to conditionally show the field label:
- For starter event we have created field templates for each field. Here we can be relatively sure that Drupal will determine beforehand if the field is empty.
  - To do this here, we would have to change the field title of "Related documents".
- For project we have created a macro for field + title.

Not fully happy with either option. What can we do? Let's talk.

templates/content/node--oe-sc-person--oe-w-content-banner.html.twig
```diff
@@ -0,0 +1,41 @@
+{#
+/**
+ * @file
+ * Person content banner display.
+ */
+#}
+
+{% set _meta = [] %}
+{% if not node.oe_sc_person_occupation.isEmpty() %}
```
donquixote: Again not a friend of these emptiness checks that don't take into account if the user has permission to see, or if the formatter turns out empty html.
We can drop this entire section for `_meta`, see below.

templates/content/node--oe-sc-person--oe-w-content-banner.html.twig
```diff
@@ -0,0 +1,41 @@
...
+    {% endfor %}
+  </div>
+{% endset %}
+<div{{ attributes }}>
+  {{ pattern('content_banner', {
+    background: 'gray',
+    title: label,
+    content: _content,
+    image: image,
+    meta: _meta,
```
donquixote:
```suggestion
    meta: [
      content.oe_sc_person_occupation|field_value,
      content.oe_sc_person_position|field_value,
      content.oe_sc_person_country|field_value,
    ]|filter(element => element is not empty),
```

donquixote: And drop the `_meta` section above.

templates/content/node--oe-sc-person--oe-w-content-banner.html.twig
```diff
@@ -0,0 +1,41 @@
...
+  <div class="mt-2-5 mb-2">
+    {% for _link in social_links %}
+      {% set _link = _link|merge(
+        {
+          'attributes': create_attribute().addClass(['d-block', 'd-md-inline-block', 'mb-3', 'mb-md-0', 'standalone'])
+        }
+      ) %}
+      <div class="mb-3 mb-md-0 d-block d-md-inline-block me-md-4-5">
+        {{ pattern('link', _link) }}
+      </div>
```
donquixote: Wow, the markup from netlify for this section is super crazy for what is does.
I would do this differently, but on the other hand, we implement what the front-end team gives us?
You may bring it up in the team channel.

donquixote: Either way the `_link|merge(...)` can be inlined here. No need for the `{% set _link = ... %}` above.
`{{ pattern('link', _link|merge(...)) }}`.

templates/content/node--oe-sc-person--teaser.html.twig
```diff
@@ -0,0 +1,25 @@
...
+{% set _title %}
+  <a class="standalone" href="{{ url }}">{{ label }}</a>
+{% endset %}
+{% set _meta = [] %}
+{% if not node.oe_sc_person_occupation.isEmpty() %}
+  {% set _meta = _meta|merge([content.oe_sc_person_occupation|field_value]) %}
+{% endif %}
+{% if not node.oe_sc_person_position.isEmpty() %}
+  {% set _meta = _meta|merge([content.oe_sc_person_position|field_value]) %}
+{% endif %}
```
donquixote: Again, I suggest a solution similar to content banner.

modules/oe_whitelabel_starter_person/config/install/core.entity_view_display.node.oe_sc_person.full.yml
```diff
@@ -0,0 +1,56 @@
...
+    - text
+    - user
+id: node.oe_sc_person.full
+targetEntityType: node
+bundle: oe_sc_person
+mode: full
+content:
+  oe_sc_person_additional_info:
+    type: text_default
+    label: above
```
donquixote: We have 1x label above and 1x label hidden, for no reason.
I'd say hide both?

donquixote: We are hard-coding the labels in the template, so..

modules/oe_whitelabel_starter_person/config/install/core.entity_view_display.node.oe_sc_person.oe_w_content_banner.yml
```diff
@@ -0,0 +1,103 @@
...
+    weight: 4
+    region: content
+  oe_sc_person_first_name:
+    type: string
+    label: above
+    settings:
+      link_to_entity: false
+    third_party_settings: {  }
+    weight: 0
+    region: content
```
donquixote: We don't use this field explicitly in the template, so we can remove it from the config.
Also check other fields here and in teaser view mode, that we don't need.

tests/src/Functional/ContentPersonRenderTest.php
```diff
@@ -0,0 +1,280 @@
...
+    $this->assertCount(1, $content_banner);
+
+    $this->assertEquals(
+      'Stefan Mayer',
+      $content_banner->filter('.card-title')->text()
+    );
+    $this->assertEquals(
+      'DG Test',
+      $content_banner->filter('.card-body > div.my-4 > span:nth-child(1)')->text()
+    );
```
donquixote: Note:
The selector is not ideal, but I don't know a much better solution. We could target the .text-muted, but even that might not be optimal. Let's keep as is.

tests/src/Functional/ContentPersonRenderTest.php
```diff
@@ -0,0 +1,280 @@
...
+    $this->assertStringNotContainsString(
+      'Additional information',
+      $content->text()
+    );
+
+    // Assert documents are not present.
+    $this->assertStringNotContainsString(
+      'Related documents',
+      $content->text()
+    );
```
donquixote: I think we should avoid these super-specific selectors for _negative_ assertions.
Otherwise, if any of the classes above is removed, the assertions will no longer fail if the text is present.
Instead, I think we can just assert that the labels are nowhere to be found in the page.

```suggestion
    // Assert additional information is not present.
    $this->assertStringNotContainsString(
      'Additional information',
      $crawler->text()
    );

    // Assert documents are not present.
    $this->assertStringNotContainsString(
      'Related documents',
      $crawler->text()
    );
```

tests/src/Functional/ContentPersonRenderTest.php
```diff
@@ -0,0 +1,280 @@
...
+    $this->assertEquals(
+      'Germany',
+      $content_banner->filter('.card-body > div.my-4 > span:nth-child(3)')->text()
+    );
+
+    // Assert content banner image.
+    $image = $content_banner->filter('img');
+    $this->assertCount(1, $image);
+    $this->assertCount(1, $image->filter('.card-img-top'));
```
donquixote: For positive assertions we can start directly with the most specific selectors.
_Unless_ we want to prove that no additional images exist - but then we'd rather prove this for the entire page, not just the content banner.
```suggestion
    $image = $content_banner->filter('img.card-img-top');
    $this->assertCount(1, $image);
```

tests/src/Functional/ContentPersonRenderTest.php
```diff
@@ -0,0 +1,280 @@
...
+      'This field is used to add a short biography of the person.',
+      $content_banner->filter('p')->text()
+    );
+
+    $this->assertEquals(
+      'Twitter profile',
+      $content_banner->filter('div.mb-2 > div.mb-3 > a.standalone')->text()
+    );
+
+    $content = $crawler->filter('div.col-12.col-lg-10.col-xl-9.col-xxl-8');
```
donquixote: Note: Here the very specific selector is ok because we use it for _positive_ assertions.
Well actually these classes are questionable, but they are added in our template, and coming from netlify, so it is legitimate that we want to test them here.

tests/src/Functional/ContentPersonRenderTest.php
```diff
@@ -0,0 +1,280 @@
...
+    );
+
+    $this->assertEquals(
+      'Related documents',
+      $content->filter('h2')->eq(1)->text()
+    );
+    $this->assertEquals(
+      'Event document test',
+      $content->filter('.bcl-file p')->text()
+    );
```
donquixote: We could add more test coverage for document groups.
Should we wait until we have more spec from front-end or UX?

tests/src/Functional/ContentPersonRenderTest.php
```diff
@@ -0,0 +1,280 @@
...
+
+  /**
+   * Tests the person rendered in 'Teaser' view mode.
+   */
+  public function testPersonRenderingTeaser(): void {
+    $node = $this->createExamplePersonWithAllFields();
+    // Build node teaser view.
+    $builder = \Drupal::entityTypeManager()->getViewBuilder('node');
+    $build = $builder->view($node, 'teaser');
+    $render = $this->container->get('renderer')->renderRoot($build);
```
donquixote: I know this is copied from elsewhere, but can we rename `$render` -> `$markup`?
Because this is what it is, a Drupal Markup object representing rendered html.

tests/src/Functional/ContentPersonRenderTest.php
```diff
@@ -0,0 +1,280 @@
...
+    $article = $crawler->filter('article');
+    $this->assertCount(1, $article);
+
+    $this->assertEquals(
+      'Stefan Mayer',
+      $article->filter('h1.card-title')->text()
+    );
+    $image = $article->filter('img');
+    $this->assertCount(1, $image);
+    $this->assertCount(1, $image->filter('.card-img-top'));
```
donquixote:
```suggestion
    $image = $article->filter('img.card-img-top');
    $this->assertCount(1, $image);
```
Unless you want to verify that we have no unexpected images beyond the 1st - but then we'd do it for the entire `$crawler` content.

donquixote (changes requested):
A little bit of feedback.

templates/content/node--oe-sc-person--oe-w-content-banner.html.twig
donquixote: @escuriola now the `_link|merge(..)` is duplicated.

modules/oe_whitelabel_starter_person/README.md
```diff
@@ -0,0 +1 @@
+# OpenEuropa Whitelabel Starter Person
```
brummbar: This is empty, so it's not needed.

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
brummbar: Source is not needed, media is already added above.

modules/oe_whitelabel_starter_person/config/install/core.entity_view_display.node.oe_sc_person.oe_w_content_banner.yml
brummbar: Was this remark implemented?

modules/oe_whitelabel_starter_person/config/install/core.entity_view_display.node.oe_sc_person.oe_w_content_banner.yml
```diff
@@ -0,0 +1,89 @@
...
+content:
+  oe_sc_person_country:
+    type: address_country_default
+    label: hidden
+    settings: {  }
+    third_party_settings: {  }
+    weight: 2
+    region: content
+  oe_sc_person_image:
+    type: entity_reference_entity_view
```
brummbar: We are preprocessing later on this field by ourselves, so there's no need to have Drupal load and prepare the view mode for this entity, correct? So we can either hide it or use a simpler formatter. For me, I would hide it.

templates/content/node--oe-sc-person--full.html.twig
brummbar: Can't we use the field template here, instead of inlining inside the view mode? This way if the field is empty, it should not render by default as far as I know.

templates/content/node--oe-sc-person--full.html.twig
```diff
@@ -0,0 +1,24 @@
...
+        <h2 class="mb-4">{{ 'Additional information'|t }}</h2>
+        <div class="mb-4-5">
+          {{ content.oe_sc_person_additional_info|field_value }}
+        </div>
+      {% endif %}
+      {% if not node.oe_sc_person_documents.isEmpty() %}
```
brummbar: Same here as above.

templates/content/field--oe-document-reference--oe-documents--oe-document-group.html.twig
```diff
@@ -0,0 +1,13 @@
+{#
+/**
+ * @file
+ * Field template of the document reference document group entity.
+ *
+ * @see ./core/themes/stable/templates/field/field.html.twig
+ */
+#}
+{% for item in items %}
+  <div class="ps-4{{ not loop.last ? ' mb-3' }}">
```
brummbar: Let's drop the ps-4
image

tests/src/Functional/ContentPersonRenderTest.php
```diff
@@ -0,0 +1,287 @@
...
+use Drupal\media\Entity\Media;
+use Drupal\node\NodeInterface;
+use Drupal\Tests\media\Traits\MediaTypeCreationTrait;
+use Drupal\Tests\TestFileCreationTrait;
+use Symfony\Component\DomCrawler\Crawler;
+
+/**
+ * Tests that the Person content type renders correctly.
+ */
+class ContentPersonRenderTest extends WhitelabelBrowserTestBase {
```
brummbar: PersonContentRenderTest.

brummbar (changes requested):
Let's make sure we followup on why the email icon for the socials is not showing. I think we need the BCL update for that, so it's fine for now.

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
donquixote: This was not done.

modules/oe_whitelabel_starter_person/config/install/core.entity_view_display.node.oe_sc_person.oe_w_content_banner.yml
escuriola: Yes, this remarks referred to to the First name and Last name fields, are not used in any of the views because we use the title.

templates/content/field--oe-document-reference--oe-document--oe-document.html.twig
donquixote: @drishu Why?
The original name targets entity type + field name + bundle. Reducing it would only target entity type + field name or entity type + bundle. Or worse, in this case I think it will target both. So it will target other fields in oe_document view mode, _and_ it will target other instances of the oe_document field.

I think we should usually target the exact field instance. And for nodes, we should target the field instance in a specific view mode.

@escuriola ^

templates/content/field--oe-document-reference--oe-documents--oe-document-group.html.twig
donquixote: @brummbar Without the ps-4, there is no clear distinction where the group ends before single documents.
I don't like it either. Let's see what the UX team comes up with.

donquixote (changes requested):

tests/src/Functional/PersonContentRenderTest.php
```diff
@@ -0,0 +1,287 @@
...
+
+  /**
+   * Creates an example person node with required and optional fields.
+   *
+   * @return \Drupal\node\NodeInterface
+   *   Person node.
+   */
+  protected function createExamplePersonWithAllFields(): NodeInterface {
+    /** @var \Drupal\node\Entity\Node $node */
+    $node = \Drupal::entityTypeManager()
```
brummbar: Don't change things between reviews without explaining why (maybe with separate commit).
I see "improve node test creation", but for me you duplicated the code, there's no improvement. Restore the call to createExamplePersonWithRequiredFieldsOnly().

brummbar (changes requested):
We have some work to do in a followup to prevent showing field labels when access to media entities is removed, but for now it looks good, just one comment in the tests.

templates/content/field--oe-document-reference--oe-documents--oe-document-group.html.twig
brummbar: For now we remove.

templates/content/field--oe-document-reference--oe-documents--oe-document-group.html.twig
```diff
@@ -0,0 +1,13 @@
+{#
+/**
+ * @file
+ * Field template of the document reference document group entity.
+ *
+ * @see ./core/themes/stable/templates/field/field.html.twig
+ */
+#}
+{% for item in items %}
+  <div{{ not loop.last ? ' class="mb-3"' }}">
```
donquixote: Left-over double quote.
```suggestion
  <div{{ not loop.last ? ' class="mb-3"' }}>
```

templates/content/field--oe-document-reference--oe-document--oe-document.html.twig
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * @file
+ * Field template of the document reference entity.
+ *
+ * @see ./core/themes/stable/templates/field/field.html.twig
+ */
+#}
+{% for item in items %}
+  <div{{ not loop.last ? ' class="mb-3"' }}">
```
donquixote:
```suggestion
  <div{{ not loop.last ? ' class="mb-3"' }}>
```

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
donquixote: Still not done.

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Implements template_preprocess_node() for the Person node type.
+ */
+function oe_whitelabel_starter_person_preprocess_node__oe_sc_person__oe_w_content_banner(array &$variables) {
+  _oe_whitelabel_starter_person_preprocess_image($variables);
+  _oe_whitelabel_starter_person_social_media_links($variables);
+}
+
+/**
+ * Implements template_preprocess_node() for the Person node type.
```
donquixote:
```suggestion
 * Implements hook_preprocess_HOOK() for Person ct in teaser view mode.
```

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
donquixote: Or:
```suggestion
 * Implements hook_preprocess_HOOK() for Person ct in content banner view mode.
```
If you have better idea tell us in chat.

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
```diff
@@ -0,0 +1,110 @@
...
+ */
+function oe_whitelabel_starter_person_preprocess_node__oe_sc_person__teaser(array &$variables) {
+  _oe_whitelabel_starter_person_preprocess_image($variables);
+}
+
+/**
+ * Helper function to preprocess the media.
+ *
+ * @param array $variables
+ *   Render array variables.
```
donquixote: These are not render array variables but theme hook variables.
```suggestion
 *   Variables from the theme preprocess function.
```

modules/oe_whitelabel_starter_person/config/install/core.entity_view_display.node.oe_sc_person.oe_w_content_banner.yml
```diff
@@ -0,0 +1,81 @@
...
+    label: hidden
+    settings:
+      trim_length: 80
+      url_only: false
+      url_plain: false
+      rel: ''
+      target: ''
+    third_party_settings: {  }
+    weight: 4
+    region: content
```
donquixote:
```suggestion
```
Seems they are not used in the template, and pulled from the node instead.

modules/oe_whitelabel_starter_person/config/install/core.entity_view_display.node.oe_sc_person.oe_w_content_banner.yml
donquixote: @escuriola Please be a bit more proactive: We had multiple fields here that are not used in the template. Do we really need multiple iterations of review to get rid of them?
If you are not sure about it, you can always ask in the chat.

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
```diff
@@ -0,0 +1,110 @@
...
+  $variables['image'] = ImageValueObject::fromImageItem($thumbnail);
+
+  $cacheability->applyTo($variables);
+}
+
+/**
+ * Helper function to preprocess the social media links.
+ *
+ * @param array $variables
+ *   Render array variables.
```
donquixote: This is not a render element.
```suggestion
 *   Variables from the preprocess function.
```

tests/src/Functional/ContentPersonRenderTest.php
donquixote: I see we did not add such tests.
Maybe you have a good argument as to why, but in that case you should either comment here, or bring it up in chat and then mention the conclusion here, e.g. "We decided in chat that we should add these tests later, not now.".

templates/content/field--oe-document-reference--oe-document.html.twig
donquixote: This comment was not addressed. The comment is still there.

templates/content/field--oe-document-reference--oe-document--oe-document.html.twig
```diff
@@ -0,0 +1,13 @@
+{#
+/**
+ * @file
+ * Field template of the document reference entity.
```
donquixote: I already commented on this before but it got lost when the file was renamed.
Well, not _lost_ really, it was still visible in the "Conversation" tab - please always check that!
Not sure how exactly, discussion in chat ongoing.

templates/content/field--oe-document-reference--oe-title--oe-document-group.html.twig
```diff
@@ -0,0 +1,11 @@
+{#
+/**
+ * @file
+ * Field template of the document reference document group title entity.
```
donquixote: I see the comment was changed, but now it is very weird.
There is no "title entity".
Discussion in chat ongoing.

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
```diff
@@ -0,0 +1,110 @@
...
+ */
+function _oe_whitelabel_starter_person_social_media_links(array &$variables): void {
+  /** @var \Drupal\node\NodeInterface $node */
+  $node = $variables['node'];
+
+  $social_links = $node->get('oe_social_media_links')->getValue();
+
+  $variables['social_links'] = [];
+  foreach ($social_links as $key => $link) {
```
donquixote: Now we still have unused variable `$key` here.
```suggestion
  foreach ($social_links as $link) {
```

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
donquixote: This was only half implemented. Unused variable `$key` still there.

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
```diff
@@ -0,0 +1,110 @@
...
+  /** @var \Drupal\media\Entity\Media $media */
+  $media = \Drupal::service('entity.repository')->getTranslationFromContext($media, $node->language()->getId());
+  $cacheability = CacheableMetadata::createFromRenderArray($variables);
+  $cacheability->addCacheableDependency($media);
+
+  // Run access checks on the media entity.
+  $access = $media->access('view', $variables['user'], TRUE);
+  $cacheability->addCacheableDependency($access);
+  if (!$access->isAllowed()) {
+    $cacheability->applyTo($variables);
```
donquixote, line 64: (tbd, don't implement directly)

Now we call this 3x, before each return.
We are not adding anything else to the `$cacheability`, so we could do it just once and avoid code repetition.
```suggestion
  $cacheability->applyTo($variables);
  if (!$access->isAllowed()) {
```
and then remove the other calls below.

Of course this only works if we don't add anything else to the `$cacheability` later.

donquixote: We have to change this back once we introduce image styles.
So maybe keep it for now? @drishu

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
```diff
@@ -0,0 +1,110 @@
...
+ * @param array $variables
+ *   Render array variables.
+ */
+function _oe_whitelabel_starter_person_social_media_links(array &$variables): void {
+  /** @var \Drupal\node\NodeInterface $node */
+  $node = $variables['node'];
+
+  $social_links = $node->get('oe_social_media_links')->getValue();
+
+  $variables['social_links'] = [];
```
donquixote: We can drop this line.
The array key will be auto-created, and if not we don't need it.
```suggestion
```

templates/content/node--oe-sc-person--oe-w-content-banner.html.twig
```diff
@@ -0,0 +1,33 @@
+{#
+/**
+ * @file
+ * Person content banner display.
+ */
+#}
+{% set _content %}
+  {{ content.oe_summary|field_value }}
+  <div class="mt-2-5 mb-2">
```
donquixote: Please check if `social_links` is empty. Otherwise we get an empty div here.

templates/content/node--oe-sc-person--oe-w-content-banner.html.twig
```diff
@@ -0,0 +1,33 @@
+{#
+/**
+ * @file
+ * Person content banner display.
+ */
+#}
+{% set _content %}
+  {{ content.oe_summary|field_value }}
```
donquixote: Do we need `|field_value` here?
We are already hiding the label, and the usual field div does not hurt.

templates/content/field--oe-document-reference--oe-document--oe-document.html.twig
donquixote: And why do we need this div with mb-3 at all?
The documents already have `mt-4`.
On the other hand: Imo this mt-4 from `pattern-file.html.twig` is questionable, so we could discuss if we really need it.

donquixote: Actually for a single-value field this is really pointless!
Keep in mind `oe_document` is single-value, whereas `oe_documents` in `oe_document_group` is multiple-value.

templates/content/field--oe-document-reference--oe-documents--oe-document-group.html.twig
donquixote: As above:
Why do we need this div with mb-3 at all?
The documents already have mt-4.
On the other hand: Imo this mt-4 from pattern-file.html.twig is questionable, so we could discuss if we really need it.

templates/content/field--oe-document-reference--oe-title--oe-document-group.html.twig
```diff
@@ -0,0 +1,11 @@
...
+ *
+ * @see ./core/themes/stable/templates/field/field.html.twig
+ */
+#}
+{% for item in items %}
+  <h3 class="fs-4 mt-3 mb-2">{{ item.content }}</h3>
```
donquixote, line 10: Currently the title is closer to the document or group above than the one it acts as a label for.
What should we do here? TBD.

donquixote: Let's keep it until we get an updated spec for these document groups.

donquixote (changes requested):

templates/content/field--oe-document-reference--oe-documents--oe-document-group.html.twig
escuriola: I add because it was in the ewcms template and francesco ask to do the same.

templates/content/field--oe-document-reference--oe-document--oe-document.html.twig
escuriola: Templates are needed to remove the "Document" subtitle in each Document or Document group

tests/src/Functional/ContentPersonRenderTest.php
escuriola: I was waiting for the design changes, I thought we would have news soon. I'm going to start with them

templates/content/field--oe-document-reference--oe-document.html.twig
escuriola: I changed the comments for the templates, take a look. I tried to compose as you said before:
Entity type + Bundle + field in the comment

--> Field template of the oe_document_reference for the oe_document title.

tests/src/Functional/ContentPersonRenderTest.php
escuriola: I added a document group and I tested it.

templates/content/field--oe-document-reference--oe-documents--oe-document-group.html.twig
escuriola: I remove the left-over double quote.
For the class you mention, I follow what Francesco said about to take the same ewcms has. I think that as long as we don't have the final designs we can leave it like this

templates/content/field--oe-document-reference--oe-document--oe-document.html.twig
escuriola: I'm not sure what do you mean, but all the templates are needed because if not a "Document" subtitle is displayed in each document.

modules/oe_whitelabel_starter_person/config/install/core.entity_view_display.node.oe_sc_person.oe_w_content_banner.yml
escuriola: You're right. Done

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
escuriola: Any news here @drishu @donquixote

templates/content/node--oe-sc-person--oe-w-content-banner.html.twig
escuriola: I've removed it. For me it's ok. Maybe we can discuss and standarise it everywhere.

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * Implements hook_preprocess_HOOK() for Person ct in content banner view mode.
+ */
+function oe_whitelabel_starter_person_preprocess_node__oe_sc_person__oe_w_content_banner(array &$variables) {
+  _oe_whitelabel_starter_person_preprocess_image($variables);
+  _oe_whitelabel_starter_person_social_media_links($variables);
+}
+
+/**
+ * IImplements hook_preprocess_HOOK() for Person ct in teaser view mode.
```
donquixote:
```suggestion
 * Implements hook_preprocess_HOOK() for Person ct in teaser view mode.
```

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
```diff
@@ -0,0 +1,109 @@
...
+ */
+function oe_whitelabel_starter_person_preprocess_node__oe_sc_person__teaser(array &$variables) {
+  _oe_whitelabel_starter_person_preprocess_image($variables);
+}
+
+/**
+ * Helper function to preprocess the media.
```
donquixote: I think "preprocess the media" is slightly misleading. Maybe this is better?
```suggestion
 * Helper function to prepare the image for Person ct.
```

templates/content/field--oe-document-reference--oe-document--oe-document.html.twig
```diff
@@ -0,0 +1,13 @@
+{#
+/**
+ * @file
+ * Field template of the oe_document_reference for the oe_document document.
```
donquixote, line 4: Still awkward, but I don't have a good counter proposal.
Let's keep it, unless @drishu has a better idea during second review.
(see also the other field templates)

tests/src/Functional/PersonContentRenderTest.php
```diff
@@ -0,0 +1,296 @@
...
+    $media_document->save();
+
+    $document_reference = \Drupal::entityTypeManager()
+      ->getStorage('oe_document_reference')
+      ->create([
+        'type' => 'oe_document',
+        'oe_document' => $media_document,
+        'status' => 1,
+      ],
+      );
```
donquixote:
```suggestion
      ]);
```
The opening `([` are on the same line, so the closing `])` should be too.

tests/src/Functional/PersonContentRenderTest.php
```diff
@@ -0,0 +1,296 @@
...
+    $document_reference->save();
+    $document_group_reference = \Drupal::entityTypeManager()
+      ->getStorage('oe_document_reference')
+      ->create([
+        'type' => 'oe_document_group',
+        'oe_title' => 'Curriculum Vitae',
+        'oe_documents' => [$media_document, $media_document],
+        'status' => 1,
+      ],
+      );
```
donquixote:
```suggestion
      ]);
```
same as above.

tests/src/Functional/PersonContentRenderTest.php
```diff
@@ -0,0 +1,296 @@
...
+    $node->set('oe_summary', 'This field is used to add a short biography of the person.');
+    $node->set('oe_sc_person_additional_info', 'Additional information example field.');
+    $node->set('oe_social_media_links', [
+      'uri' => 'https://twitter.com',
+      'title' => 'Twitter profile',
+      'link_type' => 'twitter',
+    ]);
+    $node->set(
+      'oe_sc_person_documents',
+      [$document_reference, $document_group_reference]
```
donquixote, line 287: Note: I would put a trailing comma here, _but_ I see plenty of examples in core where they don't do this - probably to support PHP < 7.3.
Also phpcs does not complain with or without the trailing comma.
So let's keep it as is :)

tests/src/Functional/PersonContentRenderTest.php
```diff
@@ -0,0 +1,296 @@
...
+
+    $document_group_title = $content->filter('h3.fs-4');
+    $this->assertEquals('Curriculum Vitae', $document_group_title->text());
+    $files = $content->filter('.bcl-file');
+    $this->assertEquals(
+      'Person document test',
+      $files->filter('p')->text()
+    );
+
+    $this->assertCount(3, $files);
```
donquixote, line 152: The assertions for document group don't go very deep. But maybe it's enough.
Let's keep it unless @drishu complains in next review.

templates/content/node--oe-sc-person--oe-w-content-banner.html.twig
donquixote: The condition needs to be outside the wrapper div.

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
donquixote: Keep it for next reviewer.

templates/content/field--oe-document-reference--oe-document--oe-document.html.twig
donquixote: Yes the templates are needed.

But the wrapper div does not really do much for us:
- For single-value field, the `loop.last` is always true, so the class will never be added.
- For multiple-value field, the `mb-3` class will be added for some items, but the `mb-3` margin vanishes in the bigger `mt-4` from the document itself.

If you want, leave it for second reviewer @drishu :)

templates/content/field--oe-document-reference--oe-documents--oe-document-group.html.twig
donquixote: I would need to look at ewcms, no time for this unless you provide a link :)
But ok to keep until we get updated design.

donquixote (changes requested):
github is weird.
I just submitted the review only by refreshing the browser window, without using the control at the top right.
Anyway, the comments do appear so let's go.

templates/content/field--node--oe-sc-person-additional-info.html.twig
```diff
@@ -0,0 +1,18 @@
+{#
+/**
+ * @file
+ * Field template for person additional information.
+ *
+ * @see ./core/themes/stable/templates/field/field.html.twig
+ */
+#}
```
donquixote, line 8: Note: (nothing wrong)
I first thought that we target node type + field name. And then wondered why we don't have `--` between them, only `-`.
But no, the machine name of the field is `oe_sc_person_additional_info`, and it is only used for the `oe_sc_person` ct.
So we don't target this ct explicitly, but it is still ok to just use the field name.

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
donquixote: Actually, up to you, if you want you can ignore.

templates/content/field--oe-document-reference--oe-document--oe-document.html.twig
drishu: think we should keep and address the bigger issue of .mt-4 appearing in the file pattern and establish a course of action

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
drishu: agreed @donquixote, we need to code for what we have, and lets also remove the empty line 70

templates/content/field--oe-document-reference--oe-document--oe-document.html.twig
drishu: My proposal is less scientific but more humane and understandable
"Field template of the Document reference entity."
We use the label not machine name this way we can understand.

But @donquixote has to agree first, if not we leave as is

tests/src/Functional/PersonContentRenderTest.php
drishu: we should be using https://github.com/openeuropa/oe_bootstrap_theme/blob/1.x/tests/src/PatternAssertion/FilePatternAssert.php

modules/oe_whitelabel_starter_person/oe_whitelabel_starter_person.module
escuriola: I'm not sure but IMHO, this is not a blocker for this task, we have more or less the same in all the CT, so we should create a new ticket and do the fix everywhere and not block this ticket anymore, it's a ticket with a high priority to do this improvement now.

templates/content/field--oe-document-reference--oe-document--oe-document.html.twig
donquixote: Actually this was Carlos first version, just in lowercase.
My initial request here was to include the field name, bundle and entity type, so we describe what we target.
But it is hard to put that into a single line.

One option would be to put these explanations in a separate block below:
```suggestion
 * Template for the 'Document' field in the 'Document reference' entity type.
 *
 * Entity type: 'oe_document_reference'.
 * Field: 'oe_document'.
 * Bundle: 'oe_document'.
```

Another option is to keep the comment simple and not have it include all the info, and instead we expect the reader to check the template name.
Tbh, I would not trust the comment anyway.

But, here, we actually can squeeze it into one line!
```suggestion
 * Template for the 'Document' field in the 'Document' document reference type.
```

templates/content/field--oe-document-reference--oe-documents--oe-document-group.html.twig
```diff
@@ -0,0 +1,13 @@
+{#
+/**
+ * @file
+ * Field template of the oe_document_reference for the oe_document_group document.
```
donquixote, line 4:
```suggestion
 * Template for 'Documents' field in 'Document group' document reference type.
```
(omitting the 'the' to keep the line character limit)

templates/content/field--oe-document-reference--oe-title--oe-document-group.html.twig
```diff
@@ -0,0 +1,11 @@
+{#
+/**
+ * @file
+ * Field template of the oe_document_reference for the oe_document_group title.
```
donquixote, line 4:
```suggestion
 * Template for the 'Title' field in 'Document group' document reference type.
```

drishu (dismissed):
Let's follow up on the remaining issues

templates/content/field--oe-document-reference--oe-document--oe-document.html.twig
drishu: ok

templates/content/field--oe-document-reference--oe-document--oe-document.html.twig
drishu: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1866

donquixote (dismissed):

drishu (dismissed):
Approved, we convened on follow up tickets

donquixote (dismissed):

drishu (approved):

brummbar (approved):

PR #164, opened 2022-07-04
OEL-1577: Javascript fallback message.

escuriola:


templates/overrides/page/page.html.twig
```diff
@@ -47,6 +47,13 @@
  * @ingroup themeable
  */
 #}
+<noscript>
+  {{ pattern('alert', {
+    variant: 'warning',
+    message: 'Please enable JavaScript to continue using this application'|t,
+    icon: 'exclamation-triangle-fill',
+  }) }}
+</noscript>
```
donquixote: Seem ok to me.
A few changes that we _could_ make to make this easier to override in a custom theme _without_ copying the complete page template:
- Put the text into a variable which can be customized when extending the template.
- Put the `<noscript>` part into a twig block.

Also not 100% sure about the text, e.g. does a visitor see the website as an "application"?

But let's not change any of this for now, instead I let the second reviewer decide.

donquixote: Actually, one more thing I notice: The element has an "x" button which obviously does not work.
I don't know if we can fix that.

templates/overrides/page/page.html.twig
```diff
@@ -47,6 +47,13 @@
  * @ingroup themeable
  */
 #}
+<noscript>
+  {{ pattern('alert', {
+    variant: 'warning',
+    message: 'Please enable JavaScript to continue using this application'|t,
+    icon: 'exclamation-triangle-fill',
```
donquixote:
```suggestion
    icon: 'exclamation-triangle-fill',
    dismissible: false,
```
This will remove the defunct close button.

templates/overrides/page/page.html.twig
```diff
@@ -47,6 +47,13 @@
  * @ingroup themeable
  */
 #}
+<noscript>
+  {{ pattern('alert', {
+    variant: 'warning',
+    message: 'Please enable JavaScript to continue using this application'|t,
```
donquixote:
```suggestion
    message: 'Please enable JavaScript and reload the page.'|t,
```
Or propose something better.

templates/overrides/page/page.html.twig
donquixote: After talking with Francesco: Make the block, but no variable. E.g.
```twig
{% block noscript_message %}
  [...]
```

brummbar: We could also call the block "noscript" directly.

brummbar (changes requested):
We should place the warning after the header:

image

Just after these lines
```
    {{ page.hero }}
    <div class="container mt-md-4-75 mt-4">
      <div class="row">
```

donquixote (dismissed):

drishu (approved):
Approved on behalf of @brummbar

donquixote (approved):

escuriola:
Replaced by https://github.com/openeuropa/oe_whitelabel/pull/168/

PR #176, opened 2022-08-10
OEL-1919: Fix broken styling for facetapi.

drishu:


resources/sass/default.style.scss
```diff
@@ -7,6 +7,7 @@
 // Component
 // -----------------------------------------------------------------------------
 @import "components/my.component";
```
drishu, line 9: I could do some cleaning up of files we dont use, what do you think ?

templates/facets/facets-item-list.html.twig
```diff
@@ -0,0 +1,36 @@
...
+  {%- endif %}
+  {% if items or empty %}
+    {%- if title is not empty -%}
+      <legend class="col-form-label"><span class="fieldset-legend">{{ title }}</span></legend>
+    {%- endif -%}
+
+    {%- if items -%}
+      <{{ list_type }}{{ attributes }}>
+        {%- for item in items -%}
+          <li{{ item.attributes.addClass('mb-2') }}>{{ item.value }}</li>
```
drishu: I lowered the margin here from mb-3 like in netlify to mb-2, also created a ticket to follow up

templates/facets/block--facets.html.twig
```diff
@@ -0,0 +1,10 @@
+<div{{ attributes }}>
```
drishu: not related to this template but was thinking of moving all templates from under /overrides into /templates

tests/src/Functional/FacetsRenderTest.php
```diff
@@ -0,0 +1,155 @@
...
+
+    $assert = $this->assertSession();
+    $assert->elementTextContains('css', 'main h4', $block->label());
+    $assert->elementTextContains('css', 'span.source-summary-count', '(5)');
+  }
+
+  /**
+   * Tests facets block rendering.
+   */
+  public function testFacetBlock(): void {
```
drishu, line 80: I renamed the file so it all looks new, this method is what is actually added.

resources/sass/components/_facets.scss
```diff
@@ -0,0 +1,4 @@
+.oel-facets-checkbox-list {
```
brummbar: I think we can determine this by using the parent? `.facets-widget-checkbox` or due to line 9 in the twig, directly `.item-list__checkbox`.

resources/sass/default.style.scss
brummbar: We should do it in a separate issue, weird to see here "my.component".

templates/facets/facets-item-list.html.twig
```diff
@@ -0,0 +1,36 @@
...
+ * Override facets item lists, alter title, inject classes.
+ */
+#}
+<div class="facets-widget- {{- facet.widget.type -}} ">
+  {% if facet.widget.type -%}
+    {%- set attributes = attributes.addClass('item-list__' ~ facet.widget.type) %}
+  {%- endif %}
+  {% if facet.widget.type == 'dropdown' -%}
+    {%- set attributes = attributes.addClass('form-select') %}
+  {%- endif %}
```
brummbar, line 13: Not sure if we should do this here or in a preprocess. Probably better in a preprocess. In any case, let's add some blocks around to ease overrides from users.

brummbar: Since we already have a check inherited from the parent regarding dropdown, it's ok to keep it here. But let's add blocks.

brummbar (changes requested):
image
I see there is some spacing issue still. Ignore the date range, we don't support it yet (not in this ticket anyway)

resources/sass/components/_facets.scss
```diff
@@ -0,0 +1,6 @@
+.facets-widget-checkbox .item-list__checkbox,
+.facets-widget-links .item-list__links,
+.facets-widget-date_range .item-list__date_range {
```
drishu, line 3: this is the only thing I had to touch to style this date range widget, so as long as we dont have visual regression there is no point in adding the module to composer or tests

brummbar (approved):

PR #177, opened 2022-08-12
OEL-1900: Upgrade BCL to 0.25.

drishu:


modules/oe_whitelabel_contact_forms/tests/src/Kernel/ContactFormRenderTest.php
```diff
@@ -112,20 +113,63 @@ public function testContactForm(): void {
...
-    $this->assertEquals("Phone", $dt->eq(5)->text());
-    $this->assertEquals("sender_name", $dd->eq(0)->text());
-    $this->assertEquals("test@example.com", $dd->eq(1)->text());
-    $this->assertEquals("subject", $dd->eq(2)->text());
-    $this->assertEquals("welcome_message", $dd->eq(3)->text());
-    $this->assertEquals("Topic name", $dd->eq(4)->text());
-    $this->assertEquals("0123456", $dd->eq(5)->text());
+
+    $description_lists = $crawler->filter('dl');
+    $description_list_assert = new DescriptionListAssert();
```
drishu, line 118: I want to move all pattern assertion classes that I find into oebt, should I do it in this ticket or prep a new one?

modules/oe_whitelabel_paragraphs/tests/src/Kernel/Paragraphs/ListingParagraphsTest.php
```diff
@@ -4,14 +4,14 @@
 
 namespace Drupal\Tests\oe_whitelabel_paragraphs\Kernel\Paragraphs;
 
+use Drupal\file\Entity\File;
```
drishu, line 7: The failing test was due to the ListingAssertion but since I was here I sorted

modules/oe_whitelabel_contact_forms/tests/src/Kernel/ContactFormRenderTest.php
brummbar: They don't need to be moved, they need to be redone XD The ones in whitelabel I think they are mostly wrong.

brummbar:
Included in #186 .

PR #179, opened 2022-08-25
OEL-1913: Theme oe_carousel paragraph.

drishu:


modules/oe_whitelabel_paragraphs/tests/src/Functional/ParagraphsTest.php
```diff
@@ -18,6 +18,9 @@ class ParagraphsTest extends BrowserTestBase {
    */
   protected static $modules = [
     'node',
+    'oe_media',
+    'composite_reference',
```
drishu, line 22: had to enable this, I think its a mistake in oe para
https://github.com/openeuropa/oe_paragraphs/blob/67f1ebb33c15acb1a2f01778cd71cf93d2882d04/modules/oe_paragraphs_carousel/config/install/field.field.paragraph.oe_carousel.field_oe_carousel_items.yml#L9
https://github.com/openeuropa/oe_paragraphs/blob/67f1ebb33c15acb1a2f01778cd71cf93d2882d04/modules/oe_paragraphs_carousel/config/install/field.field.paragraph.oe_carousel_item.field_oe_media.yml#L10

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.module
```diff
@@ -460,6 +462,79 @@ function oe_whitelabel_preprocess_paragraph__oe_facts_figures(array &$variables)
...
+      $link_item = $sub_paragraph->get('field_oe_link')->first();
+      $slide['link'] = [
+        'path' => $link_item->getUrl()->toString(),
+        'label' => $link_item->get('title')->getValue(),
+      ];
+    }
+
+    $variables['slides'][] = $slide;
+
+    $cacheability->applyTo($variables);
```
brummbar: Shouldn't this be outside of the loop?

modules/oe_whitelabel_paragraphs/tests/src/Functional/ParagraphsTest.php
```diff
@@ -289,6 +291,96 @@ public function testListingParagraph(): void {
...
+
+    $file_1 = file_save_data(file_get_contents($fixtures_path . 'example_1.jpeg'), 'public://example_1.jpeg');
+    $file_1->setPermanent();
+    $file_1->save();
+
+    $file_2 = file_save_data(file_get_contents($fixtures_path . 'example_1.jpeg'), 'public://example_2.jpeg');
+    $file_2->setPermanent();
+    $file_2->save();
+
+    $media_storage = $this->container->get('entity_type.manager')
```
brummbar: In functional tests we should always use \Drupal::service() and not $this->container.

modules/oe_whitelabel_paragraphs/tests/src/Functional/ParagraphsTest.php
```diff
@@ -289,6 +291,96 @@ public function testListingParagraph(): void {
...
+    $image = $items[0]->find('css', 'img');
+    $this->assertStringContainsString('example_1.jpeg', $image->getAttribute('src'));
+
+    $this->assertStringContainsString('Carousel item 2', $items[1]->getHtml());
+    $this->assertStringContainsString('Caption 2', $items[1]->getHtml());
+    $link = $items[1]->find('css', 'a');
+    $this->assertSame('Link 2', $link->getText());
+    $this->assertSame('https://www.example2.com', $link->getAttribute('href'));
+    $image = $items[1]->find('css', 'img');
+    $this->assertStringContainsString('example_2.jpeg', $image->getAttribute('src'));
```
brummbar: Why not use the carousel pattern assertion here?

modules/oe_whitelabel_paragraphs/tests/src/Kernel/Paragraphs/CarouselTest.php
```diff
@@ -0,0 +1,205 @@
...
+    $expected_values['items'][1]['image']['alt'] = 'Second image bg';
+    $expected_values['items'][2]['caption_title'] = 'BG Item 3';
+    $expected_values['items'][2]['image']['src'] = file_create_url($bg_file_1->getFileUri());
+    $expected_values['items'][2]['image']['alt'] = 'First image bg';
+    $expected_values['items'][3]['caption_title'] = 'BG Item 4';
+    $expected_values['items'][3]['caption'] = 'BG Item description 4';
+    $expected_values['items'][3]['link']['path'] = 'http://www.example.com/';
+    $expected_values['items'][3]['link']['label'] = 'BG CTA 4';
+    $expected_values['items'][3]['image']['src'] = file_create_url($bg_file_2->getFileUri());
+    $expected_values['items'][3]['image']['alt'] = 'Second image bg';
```
brummbar: We are basically overwriting all the values, let's just redeclare the array as whole.

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.module
```diff
@@ -460,6 +462,79 @@ function oe_whitelabel_preprocess_paragraph__oe_facts_figures(array &$variables)
...
+    $sub_paragraph = \Drupal::service('entity.repository')
+      ->getTranslationFromContext($sub_paragraph, $paragraph->language()->getId());
+    /** @var \Drupal\media\Entity\Media $media */
+    $media = $sub_paragraph->get('field_oe_media')->entity;
+    if (!$media instanceof MediaInterface) {
+      // The media entity is not available anymore, skip the item.
+      continue;
+    }
+    // Retrieve the correct media translation.
+    $media = \Drupal::service('entity.repository')->getTranslationFromContext($media, $paragraph->language()->getId());
```
brummbar: `\Drupal::service('entity.repository')` could/should be moved out of the loop in this case.

modules/oe_whitelabel_paragraphs/tests/src/Kernel/Paragraphs/CarouselTest.php
```diff
@@ -0,0 +1,205 @@
...
+      ]);
+      $paragraph->save();
+      $paragraph->addTranslation('bg', [
+        'field_oe_title' => 'BG Item ' . $i,
+        'field_oe_text' => $i % 2 === 0 ? 'BG Item description ' . $i : '',
+        'field_oe_link' => $i % 2 === 0 ? [
+          'uri' => 'http://www.example.com/',
+          'title' => 'BG CTA ' . $i,
+        ] : [],
+      ])->save();
```
brummbar: I guess it's ok, but I'm never a fan of these complex creations, I would have just had explicit data added. It's ok anyway.

modules/oe_whitelabel_paragraphs/tests/src/Kernel/Paragraphs/ParagraphsTestBase.php
```diff
@@ -35,8 +37,10 @@ abstract class ParagraphsTestBase extends AbstractKernelTestBase {
     'oe_media',
     'oe_media_avportal',
     'oe_media_iframe',
+    'oe_multilingual',
```
brummbar: Not a fan of this, and not a fan of the base paragraph test at all as we are adding all the modules all the time and we cannot catch the paragraphs in isolation. At this point the kernel test is useless, and we should just use a browser test... Let's avoid to add these modules here. Let's keep it like the document paragraph test, without requiring oe_multilingual.

modules/oe_whitelabel_paragraphs/tests/src/Kernel/Paragraphs/ParagraphsTestBase.php
```diff
@@ -80,16 +85,18 @@ protected function setUp(): void {
...
+   * @param string $langcode
+   *   Rendering language code, defaults to 'en'.
    *
    * @return string
    *   Rendered output.
    *
    * @throws \Exception
    */
-  protected function renderParagraph(ParagraphInterface $paragraph): string {
+  protected function renderParagraph(ParagraphInterface $paragraph, string $langcode = 'en'): string {
```
brummbar:
```suggestion
  protected function renderParagraph(ParagraphInterface $paragraph, string $langcode = NULL): string {
```
This would match the view() method default parameter.

brummbar (changes requested):

modules/oe_whitelabel_paragraphs/tests/src/Functional/ParagraphsTest.php
drishu: I keep forgetting to correct my copy pasta

modules/oe_whitelabel_paragraphs/tests/src/Functional/ParagraphsTest.php
drishu: was in the spirit of the other tests, seems redundant now that we have 2 tests asserting the same, I'd drop one of them, what do you think ?

modules/oe_whitelabel_paragraphs/tests/src/Kernel/Paragraphs/CarouselTest.php
drishu: y, was copy pasta from oe_theme

modules/oe_whitelabel_paragraphs/tests/src/Kernel/Paragraphs/CarouselTest.php
drishu: y not my style either, but my tests where slow so I had to make up time with a lot of copy paste from oe_theme

brummbar (approved):

PR #183, opened 2022-09-19
OEL-1910: Drop drupal 9.2 support.

drishu:


drishu (commented):
I think we can also go with an upgrade of facets and oe_media.
If, I'll have time left I'll try to find what is up with `"easyrdf/easyrdf": "1.0.0 as 0.9.1",`

brummbar (approved):

PR #189, opened 2022-10-26
OEL-1572: Add oebt and oewt style to ckeditor.

drishu:


oe_whitelabel.info.yml
```diff
@@ -15,7 +15,7 @@ libraries-override:
         css/system.admin.css: false
 
 ckeditor_stylesheets:
-  - assets/css/oe-bcl-default.min.css
+  - assets/css/oe_whitelabel.style.min.css
```
brummbar, line 18: We don't define any css in whitelabel that could impact the wysiwyg, we can drop this line.

brummbar (changes requested):

drishu:
was 2 quick to close, we still need this PR. Could not reopen so I had to create https://github.com/openeuropa/oe_whitelabel/pull/192

PR #193, opened 2022-11-23
OEL-2109: Style the publication content type.

brummbar:


composer.json
```diff
@@ -27,6 +27,7 @@
         "drupal/entity_reference_revisions": "^1.9",
         "drupal/extra_field": "^2.1",
         "drupal/facets_form": "1.0.0-alpha3",
+        "drupal/field_display_label": "^1.3",
```
drishu: we need to check if this module is QA approved, so that sites can use it

modules/oe_whitelabel_helper/src/Plugin/field_group/FieldGroupFormatter/InpageNavigationPattern.php
```diff
@@ -0,0 +1,74 @@
...
+ * @FieldGroupFormatter(
+ *   id = "oe_whitelabel_helper_inpage_navigation_pattern",
+ *   label = @Translation("Inpage navigation pattern"),
+ *   description = @Translation("Format a field group using the inpage navigation pattern."),
+ *   supported_contexts = {
+ *     "view"
+ *   }
+ * )
+ */
+class InpageNavigationPattern extends PatternFormatterBase {
```
drishu, line 22: an isolated test could be written for this, perhaps in PatternFormatterTest, but the rendering of it is already covered by the publication test, so I dont feel its mandatory, I leave it up to you

modules/oe_whitelabel_helper/src/Plugin/field_group/FieldGroupFormatter/InpageNavigationPattern.php
```diff
@@ -0,0 +1,74 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_whitelabel_helper\Plugin\field_group\FieldGroupFormatter;
```
drishu, line 5: I'm thinking we could add field_group as a dependency in the .info

modules/oe_whitelabel_helper/src/Plugin/field_group/FieldGroupFormatter/InpageNavigationPattern.php
```diff
@@ -0,0 +1,74 @@
...
+
+use Drupal\Component\Utility\Html;
+use Drupal\Core\Render\Element;
+
+/**
+ * Format a field group using the description list pattern.
```
drishu:
```suggestion
 * Format a field group using the inpage navigation pattern.
```

modules/oe_whitelabel_helper/config/schema/oe_whitelabel_helper.schema.yml
```diff
@@ -14,16 +14,17 @@ field_group.field_group_formatter_plugin.oe_whitelabel_helper_pattern_base:
   type: field_group.field_group_formatter_plugin.base
   label: 'Mapping for the base pattern formatter settings'
   mapping:
-    label:
```
drishu, line 17: I think we still need it, see PatternFormatterBase::settingsForm()L79

drishu (dismissed):

modules/oe_whitelabel_helper/config/schema/oe_whitelabel_helper.schema.yml
brummbar: No it's already defined by the base field group plugin, the pattern formatter base doesn't call the parent as it has too much stuff so they copy the label field only.

modules/oe_whitelabel_helper/src/Plugin/field_group/FieldGroupFormatter/InpageNavigationPattern.php
brummbar: I was lazy for this, as setting this up requires quite a bit. We need to take a test entity type, create various fields, configure the view mode...

composer.json
brummbar: Yes, I didn't ask yet because if we get rid of the label differences, we can drop the module altogether.
This module doesn't have dedicated tests either.

modules/oe_whitelabel_helper/src/Plugin/field_group/FieldGroupFormatter/InpageNavigationPattern.php
brummbar: Actually, like that we force people to install field_group. But by leaving it out, Drupal won't use the plugin at all and only force field_group to be enabled when installing one of our submodules that require it.

modules/oe_whitelabel_helper/src/Plugin/field_group/FieldGroupFormatter/InpageNavigationPattern.php
brummbar: If you agree we can test this but I would split it up so we can cover all the points of the plugin (inline vs hidden labels).
I can't believe I'm saying to split a test out of the implementing ticket but... I did 😄

drishu (dismissed):

composer.json
```diff
@@ -45,8 +45,8 @@
         "openeuropa/oe_corporate_blocks": "^4.8",
         "openeuropa/oe_list_pages": "^0.20",
         "openeuropa/oe_multilingual": "^1.11",
-        "openeuropa/oe_paragraphs": "^1.16",
-        "openeuropa/oe_starter_content": "^1.0.0-beta3",
+        "openeuropa/oe_paragraphs": "^1.17",
+        "openeuropa/oe_starter_content": "dev-OEL-1724",
```
drishu: now that its merged lets switch back to EPIC-OEL-1723-Publication-v1

drishu (dismissed):

drishu (approved):

PR #194, opened 2022-11-25
OEL-1676: Add address to Event teaser.

drishu:


modules/oe_whitelabel_helper/config/schema/oe_whitelabel_helper.schema.yml
```diff
@@ -5,6 +5,11 @@ field.formatter.settings.oe_whitelabel_helper_address_inline:
     delimiter:
       type: string
       label: 'Delimiter for address items.'
+    properties:
+      type: sequence
+      sequence:
+        type: string
```
brummbar: If we put boolean here, it will save it as boolean value instead of key: key or key: "0" which is better.
We could also only save the selected ones, as anyway they are the only ones we need.

modules/oe_whitelabel_helper/src/Plugin/Field/FieldFormatter/AddressInlineFormatter.php
```diff
@@ -154,4 +169,44 @@ protected function extractAddressItems(string $string, array $replacements): arr
...
 
+  /**
+   * Provides the options for the properties display setting.
+   *
+   * @return array
+   *   The properties display options.
+   */
+  protected function getPropertiesDisplayOptions(): array {
+    return [
+      'country' => $this->t('The country'),
```
brummbar:
```suggestion
      'country' => $this->t('Country'),
```

modules/oe_whitelabel_helper/src/Plugin/Field/FieldFormatter/AddressInlineFormatter.php
```diff
@@ -154,4 +169,44 @@ protected function extractAddressItems(string $string, array $replacements): arr
...
+  /**
+   * Alters the format string depending on fields options selected.
+   *
+   * @param string $format_string
+   *   The format string.
+   *
+   * @return string
+   *   The altered format string.
+   */
+  protected function alterFormatString(string $format_string): string {
```
brummbar: Instead of altering the string, we could also drop directly all the values in the extractAddressItems() method.

Something along these lines:
```
$properties = array_map(function (string $property): string {
      return '%' . $property;
    }, array_keys(array_filter($this->getSetting('properties'))));

    if (!empty($properties)) {
      foreach ($replacements as $key => &$value) {
        if (!in_array($key, $properties)) {
          $value = '';
        }
      }
    }
```

It's the same result but basically we let the code that handles the replacements in the strings in a single place.
What do you think?

templates/content/node--oe-sc-event--teaser.html.twig
```diff
@@ -15,6 +15,7 @@
   image: image,
   meta: [
     content.oe_sc_event_dates|field_value,
+    content.oe_sc_event_location|field_value
```
brummbar, line 18: I had an empty space before the location because I left the date empty.

brummbar (changes requested):

brummbar (approved):

PR #205, opened 2023-01-13
OEL-1574: Enable oebt version of slim_select.

drishu:
- Add slim_select key to all multiple selects
- Enforce oebt slim-select js for compatibility with our themes

### DIFF
--- a/oe_whitelabel.theme
+++ b/oe_whitelabel.theme
@@ -239,4 +239,6 @@ function oe_whitelabel_library_info_alter(array &$libraries, string $extension):
       ],
     ],
   ];
+  // The slim.select css is already present in the parent theme.
+  unset($libraries['slim.select']['css']);
 }
--- a/oe_whitelabel.theme
+++ b/oe_whitelabel.theme
@@ -184,3 +184,59 @@ function oe_whitelabel_preprocess_media__document__default(&$variables) {
     $variables['translations'][] = $wrapper->toFileValueObject();
   }
 }
+
+/**
+ * Implements hook_element_info_alter().
+ */
+function oe_whitelabel_element_info_alter(array &$info): void {
+  $module_handler = \Drupal::service('module_handler');
+
+  if (isset($info['select']) && $module_handler->moduleExists('slim_select')) {
+    $info['select']['#process'][] = '_oe_whitelabel_process_element_select';
+  }
+}
+
+/**
+ * Process callback for applying slim_select on elements.
+ *
+ * @param array $element
+ *   The element to alter.
+ *
+ * @return array
+ *   The altered element.
+ *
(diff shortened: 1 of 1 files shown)


oe_whitelabel.theme
```diff
@@ -184,3 +184,53 @@ function oe_whitelabel_preprocess_media__document__default(&$variables) {
     $variables['translations'][] = $wrapper->toFileValueObject();
   }
 }
+
+/**
+ * Implements hook_element_info_alter().
+ */
+function oe_whitelabel_element_info_alter(array &$info): void {
+  $module_handler = \Drupal::service('module_handler');
```
drishu, line 192: We could also create a submodule that depends on slim_select make the thing thus optional and drop this if. The risk of this proposal would be that hook_library_info_alter here would be executed before slim_select's

brummbar (approved):

oe_whitelabel.theme
```diff
@@ -184,3 +184,61 @@ function oe_whitelabel_preprocess_media__document__default(&$variables) {
...
+ *   The altered element.
+ *
+ * @see oe_whitelabel_element_info_alter
+ */
+function _oe_whitelabel_process_element_select(array &$element): array {
+  if (!$element['#multiple']) {
+    return $element;
+  }
+
+  $element['#attributes']['class'][] = 'multi-select';
```
donquixote, line 215: It's a bit weird: Now this class is only added if slim_select is enabled. Is this by intention?

oe_whitelabel.theme
```diff
@@ -184,3 +184,61 @@ function oe_whitelabel_preprocess_media__document__default(&$variables) {
...
 }
+
+/**
+ * Implements hook_element_info_alter().
+ */
+function oe_whitelabel_element_info_alter(array &$info): void {
+  $module_handler = \Drupal::service('module_handler');
+
+  if (isset($info['select']) && $module_handler->moduleExists('slim_select')) {
+    $info['select']['#process'][] = '_oe_whitelabel_process_element_select';
```
donquixote, line 195: Instead of a `'#process'` callback to add `['#slim_select']` info, we could add the `'#slim_select'` directly here:

```php
$info['select']['#slim_select'] = [];
```

This has the additional advantage of not overwriting existing '#slim_select' settings that were added by other code, e.g. by a `hook_field_widget_complete_options_select_form_alter()`.

brummbar: This is invalid as per https://github.com/openeuropa/oe_whitelabel/issues/235#issuecomment-1802047091 .

oe_whitelabel.theme
brummbar: There's no point to comment old PRs that are already merged :)

PR #209, opened 2023-02-23
OEL-2115: Update to PHP 8.1

brummbar:


composer.json
```diff
@@ -6,50 +6,44 @@
...
         "drupal/better_exposed_filters": "^5.0",
         "drupal/composite_reference": "^2.1",
         "drupal/config_devel": "~1.2",
-        "drupal/core-composer-scaffold": "^9.3",
-        "drupal/core-dev": "^9.3",
-        "drupal/ctools": "^3.7",
+        "drupal/core-composer-scaffold": "^9.4",
         "drupal/description_list_field": "^1.0@alpha",
-        "drupal/drupal-extension": "~4.1",
+        "drupal/drupal-extension": "^5.0.0alpha1",
```
brummbar: We actually don't need Behat, so we can drop this and keep core-dev.

tests/src/Functional/ContentEventRenderTest.php
```diff
@@ -96,7 +96,7 @@ public function testEventPage(): void {
 
     $link = $crawler->filter('.bcl-content-banner a[href="/build/node/' . $node->id() . '"]');
     $this->assertCount(1, $link);
-    $this->assertObjectNotHasAttribute('target', $link);
+    $this->assertFalse(property_exists($link, 'target'));
```
brummbar: Can't we just use ->getAttribute('href') ?

tests/src/Kernel/MultilingualBlockTest.php
```diff
@@ -92,76 +92,76 @@ public function testBlockRendering(): void {
     $button_header = $crawler->filter('button.btn-close');
     $this->assertSame('modal', $button_header->attr('data-bs-dismiss'));
     $link_language = $crawler->filter('a#link_bg');
-    $this->assertSame('http://localhost/bg/%3Cnone%3E', $link_language->attr('href'));
+    $this->assertEmpty($link_language->attr('href'));
```
brummbar, line 95: Quite some weird old assertions here, was it a big bug?

drishu: It's that change in drupal core rout matching, current_route_match for this Kernel test used to be `<none>` and after the code change its just empty

brummbar:
Looks good.

drishu (approved):

PR #211, opened 2023-03-10
OEL-2257: Updated readme file for whitelabel helper module.

### LINKED ISSUE #208: Update oe_whitelabel_helper readme requirements section
Some hard dependencies were removed in 1c6da23, but the helper module's [readme](https://github.com/openeuropa/oe_whitelabel/blob/1.1.0/modules/oe_whitelabel_helper/README.md#requirements) still says the following about them:

> To be able to enable this module you will have to provide the dependent modules in your projects composer.json

I think the text should be updated to reflect that these packages doesn't necessarily have to be required for the module to work.

tibi2303:


modules/oe_whitelabel_helper/README.md
```diff
@@ -41,6 +32,4 @@ To be able to enable this module you will have to provide the dependent modules
 
 ```
 composer require openeuropa/oe_corporate_blocks
```
drishu: you can remove also references to oe_corporate_blocks, even if the composer contains this the info file of this module does not depend on it

drishu (changes requested):

drishu (approved):

PR #215, opened 2023-03-24
OEL-2267: Swap deprecated budget for new decimal fields.

drishu:


tests/src/FunctionalJavascript/ContentProjectRenderTest.php
```diff
@@ -277,6 +277,26 @@ public function testProjectRendering(): void {
     $this->assertProjectStatusVisible();
     $this->assertProjectStatus('bg-secondary', 'Planned');
     $this->assertProjectProgress(0);
+
+    // Assert budget with new fields.
+    $node->set('oe_project_eu_budget', '300.00');
```
brummbar: Here we need to use values that trigger the bug, so we are sure that we are using the new fields.
Example "104479592" and "7812356"

modules/oe_whitelabel_extra_project/src/Plugin/ExtraField/Display/EuContributionExtraField.php
```diff
@@ -0,0 +1,104 @@
...
+ * @ExtraFieldDisplay(
+ *   id = "oe_whitelabel_extra_project_eu_contrib",
+ *   label = @Translation("EU contribution"),
+ *   bundles = {
+ *     "node.oe_project",
+ *   },
+ *   visible = true
+ * )
+ */
+class EuContributionExtraField extends ExtraFieldDisplayFormattedBase implements ContainerFactoryPluginInterface {
```
brummbar: Let's improve over oe_theme :D
These extra fields are exactly the same except for two strings, let's make a base class with two abstract methods for the field names.

modules/oe_whitelabel_extra_project/src/Plugin/ExtraField/Display/EuContributionExtraField.php
```diff
@@ -0,0 +1,104 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function viewElements(ContentEntityInterface $entity) {
+    if ($entity->get('oe_project_eu_contrib')->isEmpty() && $entity->get('oe_project_budget_eu')->isEmpty()) {
+      return [];
+    }
+    $build = [];
+
+    $display_options = [
```
brummbar: Instead of making these settings unchangeable unless we update here, what about just using the configuration of the display?
Is there a way to read the current field (e.g. "oe_project_eu_contrib") display settings and use those to render either the new one, or the old one?

brummbar: Something along these lines:
```
// @todo can we get the current view mode?
$display = EntityViewDisplay::collectRenderDisplay($entity, $view_mode);
$display_settings = $display->getComponent('oe_project_eu_contrib');
$field_name = $entity->get('oe_project_eu_contrib')->isEmpty() ? 'oe_project_budget_eu' : 'oe_project_eu_contrib');
$build[] = $this->viewBuilder->viewField($entity->get($field_name), $display_options);

return $build;
```

brummbar (changes requested):
I tested and it works correctly.
Let's add some lines in the README to notify people about the drush command to execute to migrate the data.

modules/oe_whitelabel_extra_project/src/Plugin/ExtraField/Display/EuContributionExtraField.php
drishu: we have $view_mode = $this->getViewMode(); from the extra field base class but we don't have any configuration to load when we do getComponent(), because the old fields are gone, and the extra fields are not configurable

modules/oe_whitelabel_extra_project/src/Plugin/ExtraField/Display/BudgetExtraFieldBase.php
```diff
@@ -0,0 +1,102 @@
...
+    return $build;
+  }
+
+  /**
+   * Returns the old budget field name.
+   *
+   * @return string
+   *   The budget field.
+   */
+  abstract protected function getFieldBudget(): string;
```
brummbar:
```suggestion
  abstract protected function getLegacyBudgetFieldName(): string;
```

modules/oe_whitelabel_extra_project/src/Plugin/ExtraField/Display/BudgetExtraFieldBase.php
```diff
@@ -0,0 +1,102 @@
...
+
+  /**
+   * Returns the new budget field name.
+   *
+   * @return string
+   *   The budget field.
+   */
+  abstract protected function getFieldEuBudget(): string;
```
brummbar:
```suggestion
  abstract protected function getBudgetFieldName(): string;
```

brummbar (changes requested):

brummbar (approved):

PR #217, opened 2023-03-29
OEL-2278: Added check for slim_select_library path

### LINKED ISSUE #214: Incorrect slimselect warning on Status report page
After #205, OEWT overrides slim_select module's library paths and uses the required assets from OEBT. The module [doesn't detect ](https://git.drupalcode.org/project/slim_select/-/blob/1.0.0-alpha2/slim_select.install#L21-31)a slimselect library in the project, so it displays a warning on `/admin/reports/status`

```
SLIM SELECT LIBRARY: The library will be loaded from the CDN. Consider using NPM in order to get a local copy of the library. Please, refer to the [module's README](https://git.drupalcode.org/project/slim_select) file for instructions.
```

This can be confusing, so it would be nice to either change the status of the requirement to OK with some custom message, or simply remove it:

```
/**
 * Implements hook_requirements_alter().
 */
function oe_whitelabel_helper_requirements_alter(array &$requirements): void {
  unset($requirements['slim_select_library']);
}
```

tibi2303:


modules/oe_whitelabel_helper/oe_whitelabel_helper.module
```diff
@@ -84,3 +84,15 @@ function oe_whitelabel_helper_field_formatter_settings_summary_alter(array &$sum
...
 }
+
+/**
+ * Implements hook_requirements_alter().
+ */
+function oe_whitelabel_helper_requirements_alter(array &$requirements): void {
+  $theme_handler = \Drupal::service('theme_handler');
+  $theme_path = $theme_handler->getTheme('oe_bootstrap_theme')->getPath();
+  $path = '/' . $theme_path . '/assets/js/slimselect.min.js';
+  if (file_exists($path)) {
```
drishu: We know the file is there, we put it there, I'd rather check if this message is set, if the module is not enabled, then the message is missing
`isset($requirements['slim_select_library'])`

modules/oe_whitelabel_helper/oe_whitelabel_helper.module
```diff
@@ -84,3 +84,15 @@ function oe_whitelabel_helper_field_formatter_settings_summary_alter(array &$sum
...
+    unset($requirements['slim_select_library']);
```
drishu: rather then doing this we should mimic the modules behaviour and set a success message
```
  $requirements['slim_select_library'] = [
    'title' => t('Slim Select library'),
    'severity' => REQUIREMENT_OK,
    'value' => t('Library available at :path.', [
      ':path' => $path,
    ]),
  ];
```

drishu (changes requested):

drishu (approved):

PR #222, opened 2023-07-04
EPHEH-48: Add support for AV Media paragraphs

drishu:


modules/oe_whitelabel_paragraphs/tests/src/Kernel/Paragraphs/ParagraphsTestBase.php
```diff
@@ -36,6 +37,7 @@ abstract class ParagraphsTestBase extends AbstractKernelTestBase {
     'oe_media_avportal',
     'oe_media_iframe',
     'oe_paragraphs',
+    'oe_paragraphs_av_media',
```
brummbar: Let's move these to the new test, instead of the base class.

modules/oe_whitelabel_paragraphs/tests/src/Kernel/Paragraphs/AvMediaParagraphsTest.php
```diff
@@ -0,0 +1,174 @@
...
+    $media_av_video->save();
+
+    // Create an "AV Portal Photo" media.
+    $media_av_photo = $media_storage->create([
+      'bundle' => 'av_portal_photo',
+      'name' => 'AV Portal Photo',
+      'oe_media_avportal_photo' => 'P-038924/00-15',
+    ]);
+    $media_av_photo->save();
+    $remote_thumbnail_url_encrypted = Crypt::hashBase64('store2/4/P038924-35966.jpg');
```
brummbar: We have methods to create most of these medias in \Drupal\Tests\oe_whitelabel\Traits\MediaCreationTrait.

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.module
```diff
@@ -208,7 +208,69 @@ function oe_whitelabel_preprocess_paragraph__oe_text_feature_media(array &$varia
...
+function oe_whitelabel_preprocess_paragraph__oe_av_media(array &$variables): void {
+  /** @var \Drupal\paragraphs\Entity\Paragraph $paragraph */
+  $paragraph = $variables['paragraph'];
+
+  // Caches are handled by the formatter usually. Since we are not rendering
+  // the original render arrays, we need to propagate our caches to the
+  // paragraph template.
+  $cacheability = CacheableMetadata::createFromRenderArray($variables);
+
+  $cacheability->applyTo($variables);
```
brummbar: This is actually wrong, we just created the object.

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.module
```diff
@@ -208,7 +208,69 @@ function oe_whitelabel_preprocess_paragraph__oe_text_feature_media(array &$varia
...
+
+  if ($is_image) {
+    $thumbnail = $media->get('thumbnail')->first();
+    $variables['image'] = ImageValueObject::fromStyledImageItem($thumbnail, 'oe_bootstrap_theme_medium_no_crop');
+  }
+  elseif ($is_video) {
+    _oe_whitelabel_featured_media_set_embedded_media($variables, $media, $cacheability, $source);
+  }
+
+  $cacheability->applyTo($variables);
```
brummbar: Most of this is duplicated, shall we have a separate method?

brummbar (changes requested):

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.module
drishu: I've added a separate method, did notice big parts of this code also in other submodules, was thinking, often projects copy this code, would be even more reusable for us and them if we create a class where we collect most of this code, what do you think?

brummbar (changes requested):
It looks all good but the rendering of AV Portal video is strange:

image

it's caused by an extra wrapper around the field.

brummbar (approved):

PR #224, opened 2023-07-18
ERATRANS-717: Media Document can be multiple.

julien-:


julien-:
Finally not using this PR because we only use 1 File per Media but if someone needs to render several Files in a Media it could be usefull.

@todo : Implements the "translations" rendering

PR #226, opened 2023-07-28
ELA-650: Link lists 'Card' and 'Teaser' display plugin.

julien-:


modules/oe_whitelabel_link_lists/src/Plugin/LinkDisplay/ColumnLinkDisplayPluginBase.php
```diff
@@ -0,0 +1,133 @@
...
+    return [
+      'columns' => 1,
+    ] + parent::defaultConfiguration();
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildConfigurationForm(array $form, FormStateInterface $form_state) {
+    $id = Html::getUniqueId('columns');
```
brummbar: This is not needed here.

modules/oe_whitelabel_link_lists/src/Plugin/LinkDisplay/ColumnLinkDisplayPluginBase.php
```diff
@@ -0,0 +1,133 @@
...
+    $form['columns'] = [
+      '#type' => 'number',
+      '#title' => $this->t('Columns'),
+      '#min' => 1,
+      '#max' => 3,
+      '#default_value' => $this->configuration['columns'] ?? 1,
+      '#id' => $id,
+    ];
+
+    $form['equal_height'] = [
```
brummbar, line 82: These options are configurable here, but they don't exist in the default configuration specified above.

modules/oe_whitelabel_link_lists/src/Plugin/LinkDisplay/ColumnLinkDisplayPluginBase.php
```diff
@@ -0,0 +1,133 @@
...
+      '#title' => $this->t('Equal height'),
+      '#default_value' => $this->configuration['equal_height'] ?? TRUE,
+    ];
+
+    $form['background_color'] = [
+      '#type' => 'select',
+      '#title' => $this->t('Background color'),
+      '#options' => [
+        'bg-white' => $this->t('White'),
+        'bg-light' => $this->t('Light'),
```
brummbar, line 94: These values are arbitrary. That's why we need a technical analysis before writing code. We need to discuss what is going to be added here, and how.

modules/oe_whitelabel_link_lists/src/Plugin/LinkDisplay/CardDisplayPlugin.php
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildItems(LinkCollectionInterface $links): array {
+    $items = [];
+
+    /** @var \Drupal\oe_link_lists\LinkInterface $link */
+    foreach ($links as $link) {
+      $entity = $link instanceof EntityAwareLinkInterface ? $link->getEntity() : NULL;
```
brummbar: All this method is duplicated in the teaser display plugin. This means it should be in a base class.

modules/oe_whitelabel_link_lists/src/Plugin/LinkDisplay/CardDisplayPlugin.php
```diff
@@ -0,0 +1,124 @@
...
+        $link_here = Link::fromTextAndUrl($this->t('here'),
+          Url::fromUri('internal:' . '/admin/structure/' . $sub_path . '/manage/' . $bundle . '/display#edit-modes')
+        )->toString();
+        $message = new FormattableMarkup('The <b>@view_display</b> view mode is not available for <b>@bundle</b>. Please enable it @link_here', [
+          '@view_display' => $view_display_id,
+          '@bundle' => $entity->bundle(),
+          '@link_here' => $link_here,
+        ]);
+        \Drupal::messenger()->addError($message);
+        return $items;
```
brummbar: This is not a message that we should show to the editors, as they might not have access to the view mode configuration. Maybe we can instead add an error on pre-save?

composer.json
```diff
@@ -82,10 +82,20 @@
             }
         },
         "patches": {
+            "drupal/core": {
```
brummbar: Why this is needed here? We don't have field definitions.

modules/oe_whitelabel_link_lists/src/Plugin/LinkDisplay/ColumnLinkDisplayPluginBase.php
```diff
@@ -0,0 +1,133 @@
...
+    $build = parent::build($links);
+
+    $items = $this->buildItems($links);
+    if (empty($items)) {
+      return $build;
+    }
+
+    // Set additional attributes.
+    foreach ($items as &$item) {
+      $attributes = new Attribute($item['attributes'] ?? []);
```
brummbar, line 125: The buildItems() function return entries all wrapped in `['content' => // the build array]` and nothing else, so this will be always empty.

brummbar (changes requested):
Additional remark: the PR uses the listing pattern, which is a listing of cards, but "exploits" the body parameter to render a full view mode in it, card or teaser. 
This is incorrect (but clever ^_^) from a semantic point of view, and causes for example borders in the card while it should not have borders.
I will create another ticket on my board to fix this situation.

composer.json
julien-: You are right. Here is PR https://github.com/openeuropa/oe_content/pull/599 to update the current used patch

modules/oe_whitelabel_link_lists/src/Plugin/LinkDisplay/ColumnLinkDisplayPluginBase.php
julien-: Fixed

modules/oe_whitelabel_link_lists/src/Plugin/LinkDisplay/ColumnLinkDisplayPluginBase.php
julien-: This values are HTML classes used by OE Whitelabel.

modules/oe_whitelabel_link_lists/src/Plugin/LinkDisplay/ColumnLinkDisplayPluginBase.php
julien-: I let the possibility to alter / create custom DisplayPlugin

modules/oe_whitelabel_link_lists/src/Plugin/LinkDisplay/CardDisplayPlugin.php
julien-: YES ... I will move it to parent ColumnLinkDisplayPluginBase class. Do you agree with that ?

modules/oe_whitelabel_link_lists/src/Plugin/LinkDisplay/CardDisplayPlugin.php
julien-: The best way is to not offer the possibility to select a non-existent ViewDisplay but we know we have an issue between Source and Display ... I will go with a validate()

modules/oe_whitelabel_link_lists/src/Plugin/LinkDisplay/CardDisplayPlugin.php
```diff
@@ -0,0 +1,35 @@
...
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
+    return new static(
+      $configuration,
+      $plugin_id,
+      $plugin_definition,
+      $container->get('entity.repository'),
+      $container->get('entity_type.manager')
+    );
+  }
```
brummbar, line 33: This doesn't need to be here since we don't use it.

modules/oe_whitelabel_link_lists/src/Plugin/LinkDisplay/ColumnLinkDisplayPluginBase.php
```diff
@@ -0,0 +1,215 @@
...
+        $sub_path = $entity_type_id == 'node' ? 'types' : $entity->getEntityTypeId();
+        $link_here = Link::fromTextAndUrl($this->t('here'),
+          Url::fromUri('internal:' . '/admin/structure/' . $sub_path . '/manage/' . $bundle . '/display#edit-modes')
+        )->toString();
+        $message = new FormattableMarkup('The <b>@view_display</b> view mode is not available for <b>@bundle</b>. Please enable it @link_here', [
+          '@view_display' => $view_display_id,
+          '@bundle' => $entity->bundle(),
+          '@link_here' => $link_here,
+        ]);
+        \Drupal::messenger()->addError($message);
```
brummbar, line 205: We need to move this code to a validation when the link list is saved.

modules/oe_whitelabel_link_lists/src/Plugin/LinkDisplay/ColumnLinkDisplayPluginBase.php
brummbar: It's not ok to store classes as content. Also, this is limiting. Let's remove equal height and background color from this one. They have to be extensions on the project for now.

brummbar (changes requested):
If patches from other components are needed, they must be added in the composer.json.

I'm still conflicted about the equal height and background classes.

julien-:
Done in #239

PR #227, opened 2023-07-31
ELA-650: Link lists 'View mode' display plugin.

julien-:


julien-:
Will not be integrated

PR #233, opened 2023-08-30
ELA-776: Apply media browser instead of file upload.

serotonine:


brummbar:
PR is unclear, has no tests, doesn't do what written in the title.

PR #252, opened 2024-01-23
OEL-2794: Update components to run on Drupal 10.2.x

AaronGilMartinez:


.drone.yml
```diff
@@ -91,7 +104,7 @@ pipeline:
     when:
       event: tag
       matrix:
-        CORE_VERSION: "10.0.0"
+        CORE_VERSION: "10.1.0"
```
AaronGilMartinez, line 108: Changed release condition, is this OK? Should we use 10.2 instead? I think they are the same

drishu: its ok like this

README.md
```diff
@@ -28,6 +28,15 @@ Ideally this project should be managed with [Docker](https://www.docker.com/get-
 
 Check the [composer.json](composer.json) for required PHP version and other dependencies.
 
+#### Drupal 10.1
```
drishu: remove this as oe_content is a dev dependency

composer.json
```diff
@@ -103,7 +103,19 @@
         },
         "_readme": [
             "Explicit requirement of nikic/php-parser ^4 as later versions are not compatible with grumphp @see https://github.com/phpro/grumphp/issues/1119"
-        ]
+        ],
+        "patches": {
```
drishu: do we really need this in our project?

drishu (changes requested):
we can move on to php 8.2 in docker compose for local development

README.md
AaronGilMartinez: Removed

composer.json
AaronGilMartinez: Removed

drishu (approved):

PR #253, opened 2024-01-23
OEL-2767: Styles for subscription feature.

tibi2303:


tests/src/Functional/ActionBarGroupTest.php
```diff
@@ -0,0 +1,222 @@
...
+  /**
+   * Test that the Action Bar group exists in the display and has no children.
+   *
+   * @param string $bundle
+   *   The node type bundle.
+   * @param string $display
+   *   The display of the node type.
+   */
+  protected function doEmptyGroupTest(string $bundle, string $display): void {
+    $node_type = $this->container->get('entity_type.manager')->getStorage('node_type')->load($bundle);
```
drishu: dont use container->get in functional tests, rather Drupal::service as the container needs reloading sometimes is better practice

tests/src/Functional/ActionBarGroupTest.php
```diff
@@ -0,0 +1,222 @@
...
+
+  /**
+   * Test that the Action Bar group exists in the display and has no children.
+   *
+   * @param string $bundle
+   *   The node type bundle.
+   * @param string $display
+   *   The display of the node type.
+   */
+  protected function doEmptyGroupTest(string $bundle, string $display): void {
```
drishu: the function name is lacking and the function descritption is a mismatch; Also this looks rather like a separate test to me

tests/src/Functional/ActionBarGroupTest.php
```diff
@@ -0,0 +1,222 @@
...
+      throw new \InvalidArgumentException(sprintf('A bundle of type node is expected, %s not found.', $bundle));
+    }
+
+    if (empty($fields)) {
+      throw new \InvalidArgumentException("An array with field names is expected");
+    }
+
+    $bundle_fields = \Drupal::service('entity_field.manager')->getFieldDefinitions('node', $bundle);
+    if ($diff = array_diff($fields, array_keys($bundle_fields))) {
+      throw new \InvalidArgumentException(sprintf('The %s bundle does not have the expected fields: %s.', $bundle, $diff = implode(', ', $diff)));
```
drishu: does it complain about expected if you remove $diff = ?

tests/src/Functional/ActionBarGroupTest.php
```diff
@@ -0,0 +1,222 @@
...
+   * The function only accepts text fields.
+   *
+   * @param string $bundle
+   *   The node type bundle.
+   * @param string $display
+   *   The display of the node type.
+   * @param array $fields
+   *   The fields to be checked.
+   */
+  protected function doFieldsGroupTest(string $bundle, string $display, array $fields): void {
```
drishu: the function name is lacking and the function description is a mismatch; 
this function does a lot, looks like a separate test to me, we could use static provider functions and loop through the scenarios (same for the other one)

tests/src/Functional/ActionBarGroupTest.php
```diff
@@ -0,0 +1,222 @@
...
+    if ($diff = array_diff($fields, array_keys($bundle_fields))) {
+      throw new \InvalidArgumentException(sprintf('The %s bundle does not have the expected fields: %s.', $bundle, $diff = implode(', ', $diff)));
+    }
+
+    // Test that the fields can be added to the group.
+    $node_fields = [];
+    foreach ($fields as $field) {
+      $node_fields[$field] = "$field " . $this->randomMachineName();
+    }
+    $node = $this->createNode(['type' => $bundle] + $node_fields);
```
drishu: if we want to test that fields can be added to the group then doing it through the UI would have more value

tests/src/Functional/ActionBarGroupTest.php
```diff
@@ -0,0 +1,222 @@
...
+    $this->doEmptyGroupTest('oe_sc_publication', 'oe_w_content_banner');
+    $this->doEmptyGroupTest('oe_project', 'oe_w_content_banner');
+
+    // Test to place a hidden field in the group.
+    $this->assertHiddenNodeDisplayField('oe_sc_event', 'oe_w_content_banner', 'oe_content_short_title');
+
+    $this->doFieldsGroupTest('oe_sc_event', 'oe_w_content_banner', ['oe_content_short_title']);
+
+    // Test to place a content field in the group.
+    $this->assertSettingsNodeDisplayField([
```
drishu: I dont see the value in these

tests/src/Traits/NodeFieldDisplayTrait.php
```diff
@@ -0,0 +1,83 @@
...
+   * Asserts that a field is hidden in a node type display.
+   *
+   * @param string $bundle
+   *   The node type bundle.
+   * @param string $display
+   *   The display of the node type.
+   * @param string $field_name
+   *   The field to check.
+   */
+  protected function assertHiddenNodeDisplayField(string $bundle, string $display, string $field_name): void {
```
drishu: the naming of this is confusing, it can be a bit longer as long as its clear what it does

resources/sass/components/_flag.scss
```diff
@@ -0,0 +1,70 @@
+// Styles to make the subscribe message a bootstrap tooltip.
+.flag {
+  position: relative;
+  .js-flag-message {
+    --bs-tooltip-zindex: 1080;
+    --bs-tooltip-max-width: 200px;
+    --bs-tooltip-padding-x: 0.5rem;
+    --bs-tooltip-margin-bottom: 0.5rem;
+    --bs-tooltip-padding-y: 0.25rem;
+    --bs-tooltip-margin: ;
```
drishu: there is something missing here

templates/content/field-group-html-element--node--group-action-bar.html.twig
```diff
@@ -0,0 +1,30 @@
+{#
+/**
+ * @file
+ * Theme override for the "Action bar" fieldgroup html element.
+ *
+ * Available variables:
```
drishu: these are already present in the parent template no use to keep track and maintain these, same in the other templates

drishu (changes requested):
~~more importantly there are elements not themed or not sufficiantley~~

~~- The modal is far from the designs (including the buttons)~~
~~- the remove buttons in the manage page should be danger class and have icon~~
~~- the table in the manage page needs more attention~~
~~- I see a complicated test for the action bar but nothing for the rendering of the other elements like the manage page~~

resources/sass/components/_flag.scss
AaronGilMartinez: Removing, the variable is not needed, since we are recreating tooltip CSS we can drop it. The style comes from a Bootstrap deprecated variable.

templates/content/field-group-html-element--node--group-action-bar.html.twig
AaronGilMartinez: Removed the variables docs

drishu (approved):

PR #254, opened 2024-02-12
OEL-2820: FieldFormatters.

julien-:
# Render a **FeaturedMedia** field using **Image style**.
You can select the setting and it will output, as waited by our Patterns, in the format of array [src, alt]

image

# Provide a **File** in the **FileValueObject** format
# Return SVG files without image_style
```mermaid
classDiagram

    class FeaturedMediaFormatterBase:::add {
        EntityRepositoryInterface $entityRepository

        defaultSettings() array
        settingsForm(array $form, FormStateInterface $form_state)
       settingsSummary() array
    }
    

    class FeaturedMediaImageFormatter:::add {

      function viewElements(FieldItemListInterface $items, $langcode) array
    }
    FeaturedMediaFormatterBase --|> FeaturedMediaImageFormatter

    class FeaturedMediaImageValueObjectFormatter:::add {

      function viewElements(FieldItemListInterface $items, $langcode = NULL) array ['src','alt']
      function needsEntityLoad(EntityReferenceItem $item)

    }
    FeaturedMediaFormatterBase --|> FeaturedMediaImageValueObjectFormatter 

    class ImageFormatterBase:::existing_drupal {
    }
    class EntityReferenceImageValueObjectFormatter:::add {
      FileUrlGeneratorInterface $fileUrlGenerator
      EntityTypeManagerInterface $entityTypeManager
      LanguageManagerInterface $languageManager

      function defaultSettings() array
      function settingsForm(array $form, FormStateInterface $form_state) array
      function settingsSummary() array
      function view(FieldItemListInterface $items, $langcode = NULL) array
      function viewElements(FieldItemListInterface $items, $langcode = NULL) array ['src','alt']
    }
    ImageFormatterBase --|> EntityReferenceImageValueObjectFormatter

class FileFormatterBase:::existing_drupal {
    }
    class FileValueObjectFormatter:::add {
    
      function viewElements(FieldItemListInterface $items, $langcode) array
    }
    FileFormatterBase --|> FileValueObjectFormatter

    classDef add fill:darkgreen
    classDef existing_drupal fill:dodgerblue
```

### DIFF
--- a/modules/oe_whitelabel_helper/src/Plugin/Field/FieldFormatter/EntityReferenceImageValueObjectFormatter.php
+++ b/modules/oe_whitelabel_helper/src/Plugin/Field/FieldFormatter/EntityReferenceImageValueObjectFormatter.php
@@ -1,0 +1,226 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_whitelabel_helper\Plugin\Field\FieldFormatter;
+
+use Drupal\Core\Cache\CacheableMetadata;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Field\FieldDefinitionInterface;
+use Drupal\Core\Field\FieldItemListInterface;
+use Drupal\Core\Field\Plugin\Field\FieldType\EntityReferenceItem;
+use Drupal\Core\File\FileUrlGeneratorInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Language\LanguageInterface;
+use Drupal\Core\Language\LanguageManagerInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\image\Plugin\Field\FieldFormatter\ImageFormatterBase;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Plugin implementation of the 'Featured media as label' formatter.
+ *
+ * @FieldFormatter(
+ *   id = "oe_whitelabel_helper_entityreference_imageobjectvalue",
+ *   label = @Translation("ImageValueObject for pattern"),
+ *   description = @Translation("Return an object {src,alt}."),
+ *   field_types = {
+ *     "entity_reference"
+ *   }
+ * )
+ */
+class EntityReferenceImageValueObjectFormatter extends ImageFormatterBase implements ContainerFactoryPluginInterface {
+
+  /**
+   * The file URL generator.
+   *
+   * @var \Drupal\Core\File\FileUrlGeneratorInterface
+   */
+  protected $fileUrlGenerator;
+
+  /**
+   * The Get EntityTypeManagerInterface.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected $entityTypeManager;
+
+  /**
+   * The language manager service.
+   *
+   * @var \Drupal\Core\Language\LanguageManagerInterface
+   */
+  protected $languageManager;
+
+  /**
+   * Constructs an ImageFormatter object.
+   *
+   * @param array $parent_params
+   *   Array of parent parameters (PHPMD is blocking when 10 parameters).
+   * @param \Drupal\Core\File\FileUrlGeneratorInterface $file_url_generator
+   *   The file URL generator.
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   The entity_type_manager.
+   * @param \Drupal\Core\Language\LanguageManagerInterface $language_manager
+   *   The language manager service.
+   */
+  public function __construct(array $parent_params, FileUrlGeneratorInterface $file_url_generator = NULL, EntityTypeManagerInterface $entity_type_manager, LanguageManagerInterface $language_manager) {
+    parent::__construct($parent_params['plugin_id'], $parent_params['plugin_definition'], $parent_params['field_definition'], $parent_params['settings'], $parent_params['label'], $parent_params['view_mode'], $parent_params['third_party_settings']);
+
+    $this->fileUrlGenerator = $file_url_generator;
+    $this->entityTypeManager = $entity_type_manager;
+    $this->languageManager = $language_manager;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
+    $parent_params = [
+      'plugin_id' => $plugin_id,
+      'plugin_definition' => $plugin_definition,
+      'field_definition' => $configuration['field_definition'],
+      'settings' => $configuration['settings'],
+      'label' => $configuration['label'],
+      'view_mode' => $configuration['view_mode'],
+      'third_party_settings' => $configuration['third_party_settings'],
+    ];
+    return new static(
+      $parent_params,
+      $container->get('file_url_generator'),
+      $container->get('entity_type.manager'),
+      $container->get('language_manager'),
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function defaultSettings() {
+    return [
+      'image_style' => '',
+    ] + parent::defaultSettings();
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function settingsForm(array $form, FormStateInterface $form_state) {
+    $image_styles = image_style_options(FALSE);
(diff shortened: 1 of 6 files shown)


julien-:
Please, why this close ? I am using this patch in all my projects and FieldFormatter should be the Drupal direction that OEL should had been taken by providing custom entities.

**Closed** but : no question, no alert, no answer, no technical justification.

raqfialho:
Closed to match the status in Jira and since there was not discussion about it up to this point. Reopening to be further reviewed.

AaronGilMartinez:
Hi @julien- thanks for the contribution, using field formatters for the paragraphs is a good idea since it brings configuration for for fields that may need it, like images.

This PR brings a lot changes and will take time to get merged so I suggest to split it into several PRs since formatters can be delivered individually. We can start with one PR with one formatter for the moment, go through review, merge it, and then continue with the others. We can keep this PR in the meantime and when everything is done just close it.

Try to keep changes to the minimum and remember to provide tests to cover the functionalities.

modules/oe_whitelabel_helper/src/Plugin/Field/FieldFormatter/EntityReferenceImageValueObjectFormatter.php
```diff
@@ -0,0 +1,249 @@
...
+ * )
+ */
+class EntityReferenceImageValueObjectFormatter extends ImageFormatterBase implements ContainerFactoryPluginInterface {
+
+  /**
+   * The file URL generator.
+   *
+   * @var \Drupal\Core\File\FileUrlGeneratorInterface
+   */
+  protected $fileUrlGenerator;
```
donquixote, line 39: We should use typed properties.

```suggestion
  protected FileUrlGeneratorInterface $fileUrlGenerator;
```

Then, if we keep the constructor, we should use constructor property promotion.
If we populate the properties in the ::create() method and drop the constructor, then we need to keep these properties but yes we should add the type.

modules/oe_whitelabel_helper/src/Plugin/Field/FieldFormatter/EntityReferenceImageValueObjectFormatter.php
```diff
@@ -0,0 +1,249 @@
...
+   *   Array of parent parameters (PHPMD is blocking when 10 parameters).
+   * @param \Drupal\Core\File\FileUrlGeneratorInterface $file_url_generator
+   *   The file URL generator.
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   The entity_type_manager.
+   * @param \Drupal\Core\Language\LanguageManagerInterface $language_manager
+   *   The language manager service.
+   */
+  public function __construct(array $parent_params, ?FileUrlGeneratorInterface $file_url_generator = NULL, EntityTypeManagerInterface $entity_type_manager, LanguageManagerInterface $language_manager) {
+    parent::__construct($parent_params['plugin_id'], $parent_params['plugin_definition'], $parent_params['field_definition'], $parent_params['settings'], $parent_params['label'], $parent_params['view_mode'], $parent_params['third_party_settings']);
```
donquixote, line 68: This way to treat parent parameters as an array is odd. I don't see this being done elsewhere.
I understand you want to reduce the amount of copying, e.g. now at least you don't need to do it int he doc comment. If you really wanted to go in this direction, you could use array unpacking like this:
```suggestion
    parent::__construct(...$parent_params);
```
This works with named and unnamed parameters.

_However_
I think we should stick with a solution that is commonly used elsewhere.

There are two ways this is typically done:
- Simply put all the parameters. (when using phpstorm these are created automatically), OR
- Populate the new properties in the ::create() method. This avoids repeating the parent constructor signature. I see this being done even in Drupal core.

For plugins where we never call the constructor directly, option 2 is the way to go.

modules/oe_whitelabel_helper/src/Plugin/Field/FieldFormatter/EntityReferenceImageValueObjectFormatter.php
```diff
@@ -0,0 +1,249 @@
...
+      'view_mode' => $configuration['view_mode'],
+      'third_party_settings' => $configuration['third_party_settings'],
+    ];
+    return new static(
+      $parent_params,
+      $container->get('file_url_generator'),
+      $container->get('entity_type.manager'),
+      $container->get('language_manager'),
+    );
+  }
```
donquixote, line 94: If we go with option two as mentioned above, this becomes:

```suggestion
  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
    $instance = parent::create($container, $configuration, $plugin_id, $plugin_definition);
    $instance->fileUrlGenerator = $container->get('file_url_generator');
    $instance->entityTypeManager = $container->get('entity_type_manager');
    $instance->languageManager = $container->get('language_manager');
    return $instance;
  }
```

The constructor is no longer needed.

Examples I find in Drupal core 11.2:
- AvifImageEffect
- WorkspaceForm
- WorkspaceSelection
- ActiveWorkspaceTestForm

modules/oe_whitelabel_helper/src/Plugin/Field/FieldFormatter/EntityReferenceImageValueObjectFormatter.php
```diff
@@ -0,0 +1,249 @@
...
+   * {@inheritdoc}
+   */
+  public function viewElements(FieldItemListInterface $items, $langcode) {
+    $elements = [
+      'src' => '',
+      'alt' => '',
+    ];
+
+    /** @var \Drupal\media\Entity\Media[] $media_entities */
+    $media_entities = $this->getEntitiesToView($items, $langcode);
```
donquixote, line 153: We would get complaints here with higher levels of phpstan (and also my IDE).
We can fix it like this:

```suggestion
  public function viewElements(FieldItemListInterface $items, $langcode) {
    assert($items instanceof EntityReferenceFieldItemListInterface);
    $elements = [
      'src' => '',
      'alt' => '',
    ];

    /** @var \Drupal\media\Entity\Media[] $media_entities */
    $media_entities = $this->getEntitiesToView($items, $langcode);
```
(we cannot use assert for the `$media_entities` because we cannot easily use "instanceof" on an array. So `@var` is fine for that one.)

modules/oe_whitelabel_helper/src/Plugin/Field/FieldFormatter/EntityReferenceImageValueObjectFormatter.php
```diff
@@ -0,0 +1,249 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function viewElements(FieldItemListInterface $items, $langcode) {
+    $elements = [
+      'src' => '',
+      'alt' => '',
+    ];
```
donquixote, line 150: These values are not allowed in a renderable array.
We either have a key starting with '#' then the value can be anything (although it depends on the code that reads these specific keys).
Or we have a key not starting with '#', then the value _must_ be another render element.

modules/oe_whitelabel_helper/src/Plugin/Field/FieldFormatter/EntityReferenceImageValueObjectFormatter.php
```diff
@@ -0,0 +1,249 @@
...
+        $alt_text = $file_item['alt'] ?? '';
+
+        if (!empty($image_style) && $file_entity->getMimeType() !== 'image/svg+xml') {
+          $elements['src'] = $this->fileUrlGenerator->transformRelative($image_style->buildUrl($file_uri));
+        }
+        else {
+          $elements['src'] = $this->fileUrlGenerator->generateString($file_uri);
+        }
+
+        $elements['alt'] = $alt_text;
```
donquixote, line 207: We have two levels of foreach to iterate through what could be multiple files in multiple media entities.
But then we always write to the top-level of the `$elements` array, overwriting values from previous iterations.

The code needs to explicit about whether it wants to render only the first item, or all items (and return some kind of list, as other formatters do).

donquixote (changes requested):
This needs:
- A clear value proposition:
  - What is the use case? I see we are talking about entity references, but which kind of content architecture scenario are we trying to support here?
  - How would you solve the problem without these formatters, and why is that not desirable?
  - What do these formatters allow that was not possible otherwise?
- PhpUnit Tests.
- Code fixes as mentioned (but only after the first points).

This review is incomplete, I first want to get a better idea of the goal.

PR #255, opened 2024-02-12
OEL-2823: Search form placeholder translatable.

julien-:


brummbar:
Fixed in #291.

julien-:
I can confirm that it is not fixed. This PR is still relevant

PR #259, opened 2024-02-14
OEL-2504: Add an optional title, description and image on the listing page.

AaronGilMartinez:


composer.json
```diff
@@ -36,7 +36,7 @@
         "openeuropa/oe_authentication": "^1.11",
         "openeuropa/oe_contact_forms": "~1.11",
         "openeuropa/oe_content": "^3",
-        "openeuropa/oe_content_extra": "^1.1.0",
+        "openeuropa/oe_content_extra": "dev-OEL-2504",
```
AaronGilMartinez, line 39: @todo Change version when openeuropa/oe_content_extra is released

modules/oe_whitelabel_list_pages/oe_whitelabel_list_pages.module
```diff
@@ -26,3 +31,61 @@ function oe_whitelabel_list_pages_form_oe_list_pages_facets_form_alter(&$form, F
...
+    $cacheability->applyTo($variables);
+    return;
+  }
+
+  // Get the media source.
+  $source = $media->getSource();
+
+  $is_image = $source instanceof MediaAvPortalPhotoSource || $source instanceof Image;
+
+  // If it's not an image and not a video, bail out.
```
drishu:
```suggestion
  // If it's not an image bail out.
```

modules/oe_whitelabel_list_pages/config/install/core.entity_view_display.node.oe_list_page.full.yml
```diff
@@ -0,0 +1,75 @@
+langcode: en
```
drishu, line 1: To have a consisten approach and be in line with the design we need to implement the same solution using oe_w_content_banner as with the other content types

templates/list_pages/node--oe-list-page--full.html.twig
```diff
@@ -26,7 +32,7 @@
       <div class="row">
         <div class="col-md-6 col-lg-8 align-self-center">
           <h4 class="mb-0">
-            <span class="text-capitalize">{{ label }}</span>
+            <span class="text-capitalize">{{ 'Results'|t }}</span>
```
drishu, line 29: why hard code it?

drishu (changes requested):

modules/oe_whitelabel_list_pages/config/install/core.entity_view_display.node.oe_list_page.full.yml
AaronGilMartinez: Ok, changed.

modules/oe_whitelabel_list_pages/oe_whitelabel_list_pages.module
AaronGilMartinez: Changed

templates/list_pages/node--oe-list-page--full.html.twig
AaronGilMartinez: Didn't see that the editor should have the possibility to set/edit this, but it's true that this could be an inprovement in the future. Same for the filters title for example.

AaronGilMartinez: Changed images from drone too

drishu: just do `{{ label|t }}`

modules/oe_whitelabel_list_pages/oe_whitelabel_list_pages.info.yml
```diff
@@ -6,9 +6,9 @@ core_version_requirement: ^10
 
 dependencies:
   - extra_field:extra_field
-  - oe_list_pages:oe_list_page_content_type
-  - oe_list_pages:oe_list_pages
+  - oe_content_extra:oe_content_extra_list_pages
+  - oe_whitelabel:oe_whitelabel_helper
```
drishu, line 10: dont add oe_whitelabel:oe_whitelabel_helper dep as this does only declare the content banner view display, no reference to the block is present

templates/list_pages/node--oe-list-page--full.html.twig
```diff
@@ -26,7 +26,7 @@
...
```
drishu, line 29: this change is ok but breaks BC, so we can't do it

templates/list_pages/node--oe-list-page--oe-w-content-banner.html.twig
```diff
@@ -0,0 +1,12 @@
+{#
+/**
+ * @file
+ * News content banner display.
+ */
+#}
+{{ pattern('content_banner', {
+  title: label,
```
drishu, line 8: since we use the label in the full template perhaps this is where the optional title comes in, do be discussed with business what the optiginal intent for the optional title was

drishu (changes requested):

modules/oe_whitelabel_list_pages/config/install/core.entity_view_display.node.oe_list_page.full.yml
```diff
@@ -0,0 +1,134 @@
...
+  oe_content_content_owner: true
+  oe_content_legacy_link: true
+  oe_content_navigation_title: true
+  oe_content_short_title: true
+  oe_featured_media: true
+  oe_summary: true
+
+    - text
+    - user
+id: node.oe_list_page.full
```
intelektron, line 72: Aren't these lines doubled in this file? `id: node.oe_list_page.full` is listed twice

modules/oe_whitelabel_list_pages/config/post_updates/00002_content_banner/core.entity_view_display.node.oe_list_page.full.yml
```diff
@@ -0,0 +1,134 @@
...
+  oe_content_content_owner: true
+  oe_content_legacy_link: true
+  oe_content_navigation_title: true
+  oe_content_short_title: true
+  oe_featured_media: true
+  oe_summary: true
+
+    - text
+    - user
+id: node.oe_list_page.full
```
intelektron, line 72: These lines are doubled too

modules/oe_whitelabel_list_pages/oe_whitelabel_list_pages.module
```diff
@@ -26,3 +31,61 @@ function oe_whitelabel_list_pages_form_oe_list_pages_facets_form_alter(&$form, F
     $form['actions']['reset']['#attributes']['class'] = array_unique(array_merge($form['actions']['reset']['#attributes']['class'] ?? [], $reset_classes));
   }
 }
+
+/**
+ * Implements template_preprocess_node() for the List pages node type.
+ */
+function oe_whitelabel_list_pages_preprocess_node__oe_list_page(&$variables) {
```
intelektron, line 38: Perhaps introducing a field formatter would be a cleaner approach? WDYT @drishu ?

templates/list_pages/node--oe-list-page--full.html.twig
```diff
@@ -1,5 +1,5 @@
 {% apply spaceless %}
-  {{ content|without('extra_field_oe_list_page_filters:node:oe_list_page', 'extra_field_oe_list_page_results:node:oe_list_page', 'extra_field_oe_list_page_selected_filters:node:oe_list_page', 'extra_field_oe_list_page_total_count:node:oe_list_page', 'extra_field_oe_list_page_sort:node:oe_list_page') }}
+  {{ content|without('extra_field_oe_list_page_filters:node:oe_list_page', 'extra_field_oe_list_page_results:node:oe_list_page', 'extra_field_oe_list_page_selected_filters:node:oe_list_page', 'extra_field_oe_list_page_total_count:node:oe_list_page', 'extra_field_oe_list_page_sort:node:oe_list_page', 'oe_summary', 'oe_featured_media') }}
```
intelektron, line 2: These should be hidden for the full view in the UI, the configs are probably broken (I commented them too)

PR #261, opened 2024-02-28
UCPKN-2600: Accessibility statement link in the footer.

peterserfozo:


msnassar (commented):
It has been tested on UCPKN. Works as expected. However, it would be nice if the oe_corporate_blocks module do the preprocess for the accessibility link (the same way it does for the site_owner) `oe_corporate_blocks_preprocess_set_site_owner`. By doing this, we won't need to copy  `_oe_whitelabel_preprocess_accessibility_link` from the oe_theme and both oe_theme and oe_whitelabel will have same preprocess. I would like you know your thoughts?

dxvargas (dismissed):

brummbar:
@msnassar I like that, let's do it there. I'll check with core team.

brummbar:
@msnassar @peterserfozo confirmed that we can move this to corporate_blocks module as it's already depending on site_info.

drishu (dismissed):

drishu (dismissed):

tests/src/Kernel/FooterBlockTest.php
```diff
@@ -23,6 +23,8 @@ class FooterBlockTest extends SparqlKernelTestBase {
     'oe_whitelabel_helper',
     'rdf_skos',
     'system',
+    'ui_patterns',
+    'ui_patterns_library',
```
brummbar, line 27: Why this?

templates/overrides/navigation/oe-corporate-blocks-eu-footer.html.twig
```diff
@@ -1,4 +1,14 @@
 {% extends "@oe_whitelabel/overrides/navigation/oe-corporate-blocks-neutral-footer.html.twig" %}
+
+{% block accessibility_link %}
+  {% set _accessibility_link = pattern('link', {
```
brummbar, line 4: I am not a fan of using the link pattern, it doesn't support things like url attributes that might be in the url object.
We might leave it for now though.

templates/overrides/navigation/oe-corporate-blocks-neutral-footer.html.twig
```diff
@@ -70,6 +70,17 @@
...
   <p> {{ site_owner ? 'This site is managed by the @name'|t({'@name': site_owner}) }} </p>
+
+  {% if accessibility_link %}
+    {% set _accessibility_link = pattern('link', {
+      label: 'Accessibility'|t,
+      path: accessibility_link,
+      standalone: true,
+    }) %}
+    {% block accessibility_link %}
+      <p> {{ _accessibility_link }} </p>
```
brummbar: Nit: trim spaces.

brummbar (changes requested):

templates/overrides/navigation/oe-corporate-blocks-eu-footer.html.twig
drishu: we resolved that by patching with our bcl_link twig function

tests/src/Kernel/FooterBlockTest.php
drishu: for the usage of the link pattern

AaronGilMartinez (dismissed):

brummbar (dismissed):

brummbar (approved):

PR #262, opened 2024-03-12
OEL-2847: Create only the new field group in the update path.

AaronGilMartinez:


modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.post_update.php
```diff
@@ -27,5 +28,29 @@ function oe_whitelabel_extra_project_post_update_00001(): void {
...
+    'format_settings' => $default_settings,
+    'region' => 'content',
+    'weight' => 20,
+  ];
+
+  $display = EntityViewDisplay::load('node.oe_project.oe_w_content_banner');
+
+  // If no display was found, we bail out.
+  if (!isset($display)) {
+    return;
```
brummbar: Let's add some messages to inform users. Here let's return (without t())
`Content banner view display not found for event content type.`

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.post_update.php
```diff
@@ -27,5 +28,29 @@ function oe_whitelabel_extra_project_post_update_00001(): void {
...
+  }
+  // If there is a group_action_bar, we bail out.
+  if ($display->getThirdPartySetting('field_group', 'group_action_bar')) {
+    return;
```
brummbar: `Action bar field group already exists for event content banner view display.`

modules/oe_whitelabel_extra_project/oe_whitelabel_extra_project.post_update.php
```diff
@@ -27,5 +28,29 @@ function oe_whitelabel_extra_project_post_update_00001(): void {
...
+  }
+
+  $display->setThirdPartySetting('field_group', 'group_action_bar', $field_group);
+  $display->save();
```
brummbar, line 55: Last we return:
`Action bar field group was added to the event content banner view display.`

brummbar (changes requested):

brummbar (approved):

PR #267, opened 2024-04-24
EPHEH-318: Add Color Scheme

drishu:


modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.module
```diff
@@ -700,3 +735,21 @@ function _oe_whitelabel_paragraphs_preprocess_paragraph_media(array &$variables)
...
+}
+
+/**
+ * Implements hook_preprocess_paragraph() for oe_quote paragraph.
+ */
+function oe_whitelabel_paragraphs_preprocess_paragraph__oe_quote(array &$variables): void {
+  /** @var \Drupal\oe_whitelabel\ColorSchemePreprocess $color_scheme_preprocess */
+  $color_scheme_preprocess = \Drupal::classResolver(ColorSchemePreprocess::class);
+  $color_scheme_preprocess->injectColorScheme($variables);
+}
```
drishu, line 775: I'm contemplating a dedicated module

src/ColorSchemePreprocess.php
```diff
@@ -0,0 +1,92 @@
...
+      'text_colored' => FALSE,
+      'background' => FALSE,
+      'primary_background' => FALSE,
+      'secondary_background' => FALSE,
+      'danger_background' => FALSE,
+      'success_background' => FALSE,
+    ];
+
+    if (array_diff_key($options, $default_options)) {
+      throw new \InvalidArgumentException('Invalid options provided.');
```
brummbar: if people want to override this, we throw an error and they cannot reuse the method. Maybe we should just ignore options we don't handle?

src/ColorSchemePreprocess.php
```diff
@@ -0,0 +1,92 @@
...
+
+    if ($options['danger_background'] === TRUE) {
+      $variables['attributes']->addClass('text-bg-danger');
+      $variables['attributes']->removeClass(['bg-default', 'text-color-default']);
+    }
+
+    if ($options['success_background'] === TRUE) {
+      $variables['attributes']->addClass('text-bg-success');
+      $variables['attributes']->removeClass(['bg-default', 'text-color-default']);
+    }
```
brummbar: Does this mean that multiple background options can be used at once?

brummbar (changes requested):

src/ColorSchemePreprocess.php
drishu: we can be more flexible sure

src/ColorSchemePreprocess.php
drishu: No, its not meant to be used that way, but now I see it could be abused

src/ColorSchemePreprocess.php
```diff
@@ -62,31 +50,14 @@ public function injectColorScheme(array &$variables, array $options = []): void
...
-      $variables['attributes']->removeClass(['bg-default', 'text-color-default']);
-    }
+    match ($options['background']) {
+      'default' => $variables['attributes']->addClass('bg-default'),
+      'primary' => $variables['attributes']->addClass('text-bg-primary'),
+      'secondary' => $variables['attributes']->addClass('text-bg-secondary'),
+      'danger' => $variables['attributes']->addClass('text-bg-danger'),
+      'success' => $variables['attributes']->addClass('text-bg-success'),
+      default => FALSE,
+    };
```
brummbar: Can we change it to:
```
$class = match ($options['background']) {
  'default' => 'bg-default',
  'primary' => 'text-bg-primary',
  'secondary' => 'text-bg-secondary',
  'danger' => 'text-bg-danger',
  'success' => 'text-bg-success',
  default => FALSE,
};
if ($class !== FALSE) {
  $variables['attributes']->addClass($class);
}

brummbar (changes requested):

brummbar (dismissed):

brummbar (approved):

PR #268, opened 2024-04-29
UCPKN-3056: Agenda theming.

kp77:
Hi @brummbar,

The PR is still wip, without tests. There are some changes since we last discussed, most importanty that I ended up adding templates/preprocesses only for the bundles that are shipped by oe_agenda, and I don't override the general entity templates. I think it's cleaner like this, because we use bundle fields in all the templates. 

This affected the session templates a bit, so now I don't have a `oe-agenda-session.html.twig` that the `oe-default` and `oe-break` templates are extending, instead the `oe-break` template extends the `oe-default` (but otherwise the idea is the same). Let me know what you think, I will add comments to the code where I have something to say.

### DIFF
--- a/modules/oe_whitelabel_agenda/oe_whitelabel_agenda.module
+++ b/modules/oe_whitelabel_agenda/oe_whitelabel_agenda.module
@@ -65,6 +65,9 @@ function oe_whitelabel_agenda_preprocess_oe_agenda_session__oe_default(array &$v
         'name' => 'box-arrow-up-right',
         'size' => 'xs',
       ];
+      $link['#fields']['attributes'] = [
+        'target' => '_blank',
+      ];
     }
     $definition[] = $link;
 
--- a/modules/oe_whitelabel_agenda/oe_whitelabel_agenda.module
+++ b/modules/oe_whitelabel_agenda/oe_whitelabel_agenda.module
@@ -7,8 +7,12 @@
 
 declare(strict_types=1);
 
+use Drupal\Core\Field\FieldItemListInterface;
 use Drupal\Core\Render\Element;
+use Drupal\Core\Render\RendererInterface;
+use Drupal\link_description\Plugin\Field\FieldType\LinkDescriptionItem;
 use Drupal\oe_agenda\Entity\DayInterface;
+use Drupal\oe_agenda\Entity\SessionInterface;
 
 /**
  * Implements hook_preprocess_oe_agenda() for oe_default agenda.
@@ -21,6 +25,70 @@ function oe_whitelabel_agenda_preprocess_oe_agenda__oe_default(array &$variables
     $variables['days'][] = [
       'title' => $day->get('oe_day_date')->first()->view('default'),
       'content' => $variables['content']['oe_agenda_days'][$index],
+    ];
+  }
+}
+
+/**
+ * Implements hook_preprocess_oe_agenda_session() for oe_default session.
+ */
+function oe_whitelabel_agenda_preprocess_oe_agenda_session__oe_default(array &$variables): void {
+  $session = $variables['entity'];
+  assert($session instanceof SessionInterface);
+  $renderer = \Drupal::service('renderer');
+  assert($renderer instanceof RendererInterface);
+
+  // Build items for description list pattern displaying session links.
+  $items = [];
+  foreach (['oe_session_venue', 'oe_session_online_link'] as $index => $field_name) {
+    $field = $session->get($field_name);
+    assert($field instanceof FieldItemListInterface);
+    if ($field->isEmpty() || !$field->access()) {
+      continue;
+    }
+    $items[$index]['term'] = $field->getFieldDefinition()->getLabel() . ':';
+    $definition = [];
+    $link_description_item = $field->first();
+    assert($link_description_item instanceof LinkDescriptionItem);
+
+    // Add link to definition.
+    $link = [
+      '#type' => 'pattern',
+      '#id' => 'link',
+      '#fields' => [
+        'label' => $link_description_item->getTitle(),
+        'path' => $link_description_item->getUrl(),
+      ],
+    ];
+    if ($link_description_item->isExternal()) {
+      $link['#fields']['icon'] = [
+        'name' => 'box-arrow-up-right',
+        'size' => 'xs',
+      ];
+    }
+    $definition[] = $link;
+
+    // Add link description to definition.
+    if (!empty($link_description_item->description)) {
+      $definition[] = [
+        '#type' => 'processed_text',
+        '#text' => $link_description_item->description,
+        '#format' => 'plain_text',
+      ];
+    }
+
+    $items[$index]['definition'] = $renderer->render($definition);
+  }
+
+  // Build description list with session links.
+  if ($items) {
+    $variables['session_links'] = [
+      '#type' => 'pattern',
+      '#id' => 'description_list',
+      '#orientation' => 'horizontal',
+      '#fields' => [
+        'items' => $items,
+      ],
     ];
   }
 }
--- a/modules/oe_whitelabel_agenda/oe_whitelabel_agenda.module
+++ b/modules/oe_whitelabel_agenda/oe_whitelabel_agenda.module
@@ -7,6 +7,7 @@
 
 declare(strict_types=1);
 
+use Drupal\Core\Cache\CacheableMetadata;
 use Drupal\Core\Field\FieldItemListInterface;
 use Drupal\Core\Render\Element;
 use Drupal\Core\Render\RendererInterface;
@@ -37,13 +38,22 @@ function oe_whitelabel_agenda_preprocess_oe_agenda_session__oe_default(array &$v
   assert($session instanceof SessionInterface);
   $renderer = \Drupal::service('renderer');
   assert($renderer instanceof RendererInterface);
+  $cacheability = CacheableMetadata::createFromRenderArray($variables);
 
   // Build items for description list pattern displaying session links.
   $items = [];
   foreach (['oe_session_venue', 'oe_session_online_link'] as $index => $field_name) {
     $field = $session->get($field_name);
     assert($field instanceof FieldItemListInterface);
-    if ($field->isEmpty() || !$field->access()) {
+    if ($field->isEmpty()) {
+      // Skip if field is empty.
+      continue;
+    }
+    $access = $field->access('view', $variables['user'], TRUE);
+    $cacheability->addCacheableDependency($access);
+    if (!$access->isAllowed()) {
+      // Skip if user doesn't have access to the field.
+      $cacheability->applyTo($variables);
       continue;
     }
     $items[$index]['term'] = $field->getFieldDefinition()->getLabel() . ':';
@@ -94,4 +104,6 @@ function oe_whitelabel_agenda_preprocess_oe_agenda_session__oe_default(array &$v
       ],
     ];
   }
+
+  $cacheability->applyTo($variables);
 }
--- a/modules/oe_whitelabel_agenda/oe_whitelabel_agenda.module
+++ b/modules/oe_whitelabel_agenda/oe_whitelabel_agenda.module
@@ -1,0 +1,26 @@
+<?php
+
+/**
+ * @file
+ * OE Whitelabel Agenda module.
+ */
+
+declare(strict_types=1);
+
+use Drupal\Core\Render\Element;
+use Drupal\oe_agenda\Entity\DayInterface;
+
+/**
+ * Implements hook_preprocess_oe_agenda() for oe_default agenda.
+ */
+function oe_whitelabel_agenda_preprocess_oe_agenda__oe_default(array &$variables): void {
+  foreach (Element::children($variables['content']['oe_agenda_days']) as $index) {
+    $day = $variables['content']['oe_agenda_days'][$index]['#oe_agenda_day'];
+    assert($day instanceof DayInterface);
+    // Pass agenda days as accordion items to the template.
+    $variables['days'][] = [
+      'title' => $day->get('oe_day_date')->first()->view('default'),
+      'content' => $variables['content']['oe_agenda_days'][$index],
+    ];
+  }
+}
--- a/resources/sass/components/_agenda.scss
+++ b/resources/sass/components/_agenda.scss
@@ -7,11 +7,18 @@
     color: #6c757d !important;
   }
 
-  .bcl-description-list {
-    font-size: 0.875rem;
-  }
+  .session-content {
+    & > div:last-child,
+    .session-details > *:last-child {
+      margin-bottom: 0 !important;
+    }
 
-  .session-details > *:last-child {
-    margin-bottom: 0 !important;
+    .bcl-description-list {
+      font-size: 0.875rem;
+      dl:last-child,
+      dd > div *:last-child {
+        margin-bottom: 0 !important;
+      }
+    }
   }
 }
--- a/resources/sass/components/_agenda.scss
+++ b/resources/sass/components/_agenda.scss
@@ -8,7 +8,7 @@
   }
 
   .session-content {
-    & > div:last-child,
+    & > *:last-child,
     .session-details > *:last-child {
       margin-bottom: 0 !important;
     }
--- a/resources/sass/components/_agenda.scss
+++ b/resources/sass/components/_agenda.scss
@@ -1,0 +1,13 @@
+.oe-agenda {
+  .accordion-button {
+    font-weight: bold;
+  }
+
+  .text-muted {
+    color: #6c757d !important;
+  }
+
+  .bcl-description-list {
+    font-size: 0.875rem;
+  }
+}
--- a/resources/sass/components/_agenda.scss
+++ b/resources/sass/components/_agenda.scss
@@ -10,4 +10,8 @@
   .bcl-description-list {
     font-size: 0.875rem;
   }
+
+  .session-details > *:last-child {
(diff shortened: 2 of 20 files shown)


modules/oe_whitelabel_agenda/oe_whitelabel_agenda.module
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Implements hook_preprocess_oe_agenda_session() for oe_default session.
+ */
+function oe_whitelabel_agenda_preprocess_oe_agenda_session__oe_default(array &$variables): void {
+  $session = $variables['entity'];
+  assert($session instanceof SessionInterface);
+  $renderer = \Drupal::service('renderer');
+  assert($renderer instanceof RendererInterface);
+
+  // Build items for description list pattern displaying session links.
```
kp77, line 43: Originally, I wanted to build the description list in the template, without a preprocess. I decided to do it in a preprocess because:
1. we need to add the external icon for external URLs
2. the description list output for the description text is using nl2br.. here I changed that to processed text so the new lines will be paragraphs, which I think is better formatting

templates/agenda/oe-agenda-session--oe-default.html.twig
```diff
@@ -0,0 +1,70 @@
+{#
+/**
+ * @file
+ * Theme override to display an agenda session entity of oe_default type.
+ *
+ * @see oe_whitelabel_agenda_preprocess_oe_agenda_session__oe_default()
+ * @see ./modules/contrib/oe_agenda/templates/oe-agenda-session.html.twig
+ */
+#}
+<div{{ attributes.addClass(['row', 'mb-4-5',]) }}>
```
msnassar: I think we won't need the last comma as the elements are in one line...

`mb-4-5` I think this margin is added for the spacing between sessions? if yes, I wonder if the session template should handle this spacing... I believe the spacing between sessions should be handled in the sessions field.

In addition, by having this space here, in case the day has one session, we will have extra unwanted space... 

Also, If we want to keep the spacing consistent between sessions, if the last element in a session is paragraph, I think we should eat the margin-bottom added by the paragraph `margin-bottom: 1rem;`

modules/oe_whitelabel_agenda/oe_whitelabel_agenda.module
```diff
@@ -0,0 +1,94 @@
...
+      '#fields' => [
+        'label' => $link_description_item->getTitle(),
+        'path' => $link_description_item->getUrl(),
+      ],
+    ];
+    if ($link_description_item->isExternal()) {
+      $link['#fields']['icon'] = [
+        'name' => 'box-arrow-up-right',
+        'size' => 'xs',
+      ];
```
msnassar, line 77: I think, external link should also open in new tab.

templates/agenda/oe-agenda-session--oe-default.html.twig
```diff
@@ -0,0 +1,70 @@
...
+      <span class="text-muted">{{ content.oe_session_hours|field_value }}</span>
+    {% endblock %}
+  </div>
+  <div class="col">
+    {% block content %}
+      {% if not entity.get('name').isEmpty() %}
+        {# Display session label only if name field is not empty. #}
+        {% include '@oe-bcl/bcl-heading/heading.html.twig' with {
+          title: label,
+          title_tag: 'h4'
```
msnassar: As already clarified, the size/weight of the title is bigger than the one in the mockups. As per our discussion, this should double check with Dora...

msnassar (changes requested):
Hello @kp77 Looks good to me. Just some miner issues...

modules/oe_whitelabel_agenda/oe_whitelabel_agenda.module
kp77: I'm not sure about this, I couldn't find any such requirement. Also, it's generally considered bad practice from [usability perspective](https://ux.stackexchange.com/a/105371/94122). We don't do this with external links on UCPKN either (we have some cases of new tab opening, but for technical reasons).

templates/agenda/oe-agenda-session--oe-default.html.twig
kp77: Yes, I confirmed this with Dora. The weights are ok as is (500 on headings, 400 on body text). This is the default style from BCL and it's also the same on the mockup. For the session title, we should use the same font size as for the headings inside the session, fixed here: [624d9c2](https://github.com/openeuropa/oe_whitelabel/pull/268/commits/624d9c2cf7563c2f4d3175efa18969d162b8deec)

modules/oe_whitelabel_agenda/oe_whitelabel_agenda.module
kp77: Confirmed with BA, we will open the external links in new tab here. Fixed in: [2f372e1](https://github.com/openeuropa/oe_whitelabel/pull/268/commits/2f372e17614f81e4c6104fe38577d8536a0be8d3)

templates/agenda/oe-agenda-session--oe-default.html.twig
kp77: Spacing class moved outside session template in: [56d50a3](https://github.com/openeuropa/oe_whitelabel/pull/268/commits/56d50a319812559444c80473e0c9fa3852b73e2d)

Session details last child bottom margin is removed in: [abba1fc](https://github.com/openeuropa/oe_whitelabel/pull/268/commits/abba1fc3c559212821c13a142667679d01e9c05e)

templates/agenda/oe-agenda-session--oe-default.html.twig
```diff
@@ -0,0 +1,73 @@
...
+        </div>
+      {% endif %}
+
+      {# Venue / Online link as description list. #}
+      {{ session_links }}
+
+      {# Details. #}
+      <div class="session-details">
+        {{ content.oe_session_details|field_value }}
+      </div>
```
msnassar: If no details added, we will have empty div...
      `<div class="session-details"></div>`

templates/agenda/oe-agenda-session--oe-default.html.twig
```diff
@@ -0,0 +1,73 @@
+{#
```
msnassar, line 1: Every div in this template adds space to the bottom (Including the last div (could be any of these divs)).
This result in having an extra space between sessions all the time and inconstance spacing between them...

templates/agenda/oe-agenda-session--oe-break.html.twig
```diff
@@ -0,0 +1,20 @@
...
+
+{% block content %}
+  {% set _title = entity.get('name').isEmpty() ? 'Break'|t : label %}
+  {% include '@oe-bcl/bcl-heading/heading.html.twig' with {
+    title: _title,
+    title_tag: 'h4',
+    attributes: create_attribute().addClass('text-muted'),
+  } only %}
+
+  {{ content.oe_session_details|field_value }}
```
msnassar: Same spacing problem here with the p tag (Similar to the one in the default template)

msnassar (changes requested):
Hello @kp77 
Please look into the new comments below...

templates/agenda/oe-agenda-session--oe-break.html.twig
kp77: Fixed in [39c1bf3](https://github.com/openeuropa/oe_whitelabel/pull/268/commits/39c1bf33450e7ef48e25f95f80de3de67d8cad88)

templates/agenda/oe-agenda-session--oe-default.html.twig
kp77: Fixed in [39c1bf3](https://github.com/openeuropa/oe_whitelabel/pull/268/commits/39c1bf33450e7ef48e25f95f80de3de67d8cad88)

templates/agenda/oe-agenda-session--oe-default.html.twig
kp77: Fixed the title size also in Break session, here: [b965d77](https://github.com/openeuropa/oe_whitelabel/pull/268/commits/b965d77fee90df1ddfe197c51a35cc82ac545332)

templates/agenda/oe-agenda-session--oe-default.html.twig
kp77: A css solution to the inconsistent spacing (targeting last:child elements) was added in commit: [3239e00](https://github.com/openeuropa/oe_whitelabel/pull/268/commits/3239e00a5bd887ce1b69ac26557ff827e7af2a16)

kp77:
Hi @msnassar,

I added 3 new commits since last review round:
1. [UCPKN-3056: Normalize spacing between session sections.](https://github.com/openeuropa/oe_whitelabel/pull/268/commits/3239e00a5bd887ce1b69ac26557ff827e7af2a16) is the proposed solution for the inconsistency between sessions and session sections (this is what UCPKN-3105 ticket was created for as a follow up)
2. [UCPKN-3056: Set correct size/spacing for break/default session title.](https://github.com/openeuropa/oe_whitelabel/pull/268/commits/b965d77fee90df1ddfe197c51a35cc82ac545332) fixed session title font size and spacing to align with mockup
3. [UCPKN-3056: Display hours as muted only for break session.](https://github.com/openeuropa/oe_whitelabel/pull/268/commits/11664e035bc95516017f1a7e7714191f944b41f0) the `text-muted` class was incorrectly applied on the default session hours, but they should be muted only in break session

resources/sass/components/_agenda.scss
```diff
@@ -7,11 +7,18 @@
     color: #6c757d !important;
   }
 
-  .bcl-description-list {
-    font-size: 0.875rem;
-  }
+  .session-content {
+    & > div:last-child,
```
msnassar: This won't be applied in case session has a title only as the title tag is `h4`

templates/agenda/oe-agenda-session--oe-break.html.twig
```diff
@@ -8,6 +8,18 @@
...
 
+{% block hours %}
+  {{ pattern('icon', {
+    name: 'clock',
+    path: bcl_icon_path,
+    size: 'xs',
+    attributes: create_attribute({
+      class: ['mb-1', 'me-2'],
+    }),
+  }) }}
```
msnassar: To not repeat the icon here, I think we can use `macro`... We define it in the default template and  use `from `in order to use it in this template
https://twig.symfony.com/doc/3.x/tags/macro.html

templates/agenda/oe-agenda-session--oe-default.html.twig
msnassar: Resolved by https://citnet.tech.ec.europa.eu/CITnet/jira/browse/UCPKN-3105

msnassar (changes requested):
Hello @kp77 
Please have a look into the new comments above...

resources/sass/components/_agenda.scss
kp77: Fixed in [f774329](https://github.com/openeuropa/oe_whitelabel/pull/268/commits/f77432933c47302ed806f8a1b066d5c3c8755126)

templates/agenda/oe-agenda-session--oe-break.html.twig
kp77: Added a macro in [3a8f0ac](https://github.com/openeuropa/oe_whitelabel/pull/268/commits/3a8f0aca720bdcd44eded9b2de01f82a3024fc2e) (we don't need from as we extend)

templates/agenda/oe-agenda-session--oe-default.html.twig
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * @file
+ * Theme override to display an agenda session entity of oe_default type.
+ *
+ * @see oe_whitelabel_agenda_preprocess_oe_agenda_session__oe_default()
+ * @see ./modules/contrib/oe_agenda/templates/oe-agenda-session.html.twig
+ */
+#}
+{% macro icon(name = 'clock', size = 'xs', attributes) %}
```
msnassar: I understand the reason behind assigning default values to the parameters. But I am not sure we should do this. As this macro is generic, I prefer to have them passed instead.
Also, I believe it is better to add macros to the end of the file...

msnassar (changes requested):
Hello @kp77 Please see one miner issue...

templates/agenda/oe-agenda-session--oe-default.html.twig
kp77: This macro is added to the session template, and the default icon we use in the session is what the defaults of the macro are assigning. What is the reason not to assign default values for the macro and making it easier to insert the default session icon for hours?

Regarding where to put the macro: I would have also put it to the end of the file, but the only macro examples I found in Whitelabel are placing the macro at the top of the file. Not sure if it's 'better'  to put it to the end, I think it's just a question of preference/convention, and I was trying to follow the Whitelabel practice.

msnassar: Yes, it is part of the session template... But if we have another icons that we would like to display, I think we will use the same macro but in this case we will pass values to the parameters!
The reasons I prefer not to set default values:
- Code readability. `{{ _self.icon() }}` doesn't tell much! specially if we will use the macro to display other icons (which is not the case now).
- Less confusing for other developers who would like to extend/override the template and also reuse the macro.

Anyway. lets see the opinion from the other reviewers ;)

kp77: Well, for me `{{ _self.icon() }}` tells something: it tells that there's a macro in this template (or the parent template) that defines an icon for the template. If I don't know what it outputs and how, I should probably go and check it.

The original purpose of this macro is to output the icon we need - as we need it - in multiple places without repeating ourselves. If I have to pass everything in the different places I need the icon (and that includes also the spacing classes passed in the attributes), that goes against the original idea of using the macro (I might as well just call the pattern twice..)

The fact that this macro also can be used to output other icons in other sizes/attributes is just a bonus IMO.

brummbar: Checking core and other modules/themes, there are never defaults. Let's not use them.

modules/oe_whitelabel_agenda/oe_whitelabel_agenda.module
```diff
@@ -0,0 +1,97 @@
...
+    // Add link description to definition.
+    if (!empty($link_description_item->description)) {
+      $definition[] = [
+        '#type' => 'processed_text',
+        '#text' => $link_description_item->description,
+        '#format' => 'plain_text',
+      ];
+    }
+
+    $items[$index]['definition'] = $renderer->render($definition);
```
brummbar, line 93: I don't think we need to render, do we? Passing a render array should work.

modules/oe_whitelabel_agenda/oe_whitelabel_agenda.module
brummbar: Which business analyst? For UCKPN or OEL? Because normally I wouldn't open in new tabs either for usability.

templates/agenda/oe-agenda-session--oe-default.html.twig
kp77: Removed the defaults in [737e683](https://github.com/openeuropa/oe_whitelabel/pull/268/commits/737e6833bd909c2f26011a6d4c10d123be861917)

modules/oe_whitelabel_agenda/oe_whitelabel_agenda.module
kp77: We need to render, because we need to pass a string as item definition to the description list pattern. (It also accepts array, but not a render array, but a structured one)

modules/oe_whitelabel_agenda/oe_whitelabel_agenda.module
kp77: The requirement was created for OEL, if it's problematic we should probably discuss again.

drishu (dismissed):

modules/oe_whitelabel_agenda/oe_whitelabel_agenda.module
```diff
@@ -0,0 +1,97 @@
...
+  assert($session instanceof SessionInterface);
+  $renderer = \Drupal::service('renderer');
+  assert($renderer instanceof RendererInterface);
+
+  // Build items for description list pattern displaying session links.
+  $items = [];
+  foreach (['oe_session_venue', 'oe_session_online_link'] as $index => $field_name) {
+    $field = $session->get($field_name);
+    assert($field instanceof FieldItemListInterface);
+    if ($field->isEmpty() || !$field->access()) {
```
brummbar: We need to gather the access cache metadata and bubble it up here.

brummbar (dismissed):
We miss tests.

modules/oe_whitelabel_agenda/oe_whitelabel_agenda.module
kp77: This is done in 9cb8051

kp77:
Hi @brummbar,

The changes since your last review are all commits after and including: [UCPKN-3056: Add session link fields access cache metadata to render array.](https://github.com/openeuropa/oe_whitelabel/pull/268/commits/9cb8051914a632e5e06455ea1e5e94fc70715553)

Please have a look.

drishu (dismissed):

drishu (approved):

msnassar (approved):

PR #273, opened 2024-05-13
OEL-1771: Moved patterns tests from oe_whitelabel. 

tibi2303:


AaronGilMartinez (changes requested):
Changes seems right but to make sure we need drone green and conflicts solved.

To get changes from oe_bootstrap_theme branch you can do:
`composer require openeuropa/oe_bootstrap_theme:"dev-OEL-1771"`

And then test all together, when bootstrap is released we can change composer to get new version.

Can you check both, please?

composer.json
```diff
@@ -12,7 +12,7 @@
         "drupal/daterange_compact": "^2.0.1",
         "drupal/twig_field_value": "^2.0.2",
         "openeuropa/composer-artifacts": "^1.0.0-alpha1",
-        "openeuropa/oe_bootstrap_theme": "^1.4.0"
+        "openeuropa/oe_bootstrap_theme": "^1.10.0"
```
drishu, line 15: if you need to upgrade then best to do as @AaronGilMartinez instructed

drishu (changes requested):

composer.json
tibi2303: tried but it failed:
https://drone.fpfis.eu/openeuropa/oe_whitelabel/2098/15

https://github.com/openeuropa/oe_whitelabel/pull/273/commits/229a46b428035c32f4c4efb03f7bf244bdca0b40

Fra helped me out to patch and update to this version. I am not sure if we should keep it or not.

composer.json
```diff
@@ -71,6 +71,11 @@
     "extra": {
         "composer-exit-on-patch-failure": true,
         "enable-patching": true,
+        "patches": {
+            "openeuropa/oe_bootstrap_theme": {
+                "new pattern assert": "https://github.com/openeuropa/oe_bootstrap_theme/compare/1.10.0...OEL-1771.patch"
```
drishu:
```suggestion
                "new pattern assert": "https://github.com/openeuropa/oe_bootstrap_theme/compare/1.10.0...OEL-1771.diff"
```

tibi2303: done.

drishu (dismissed):

AaronGilMartinez (approved):

drishu (approved):

PR #275, opened 2024-05-29
OEL-1863: Added file assert for PersonContentRenderTest.

tibi2303:


AaronGilMartinez (dismissed):

AaronGilMartinez (dismissed):

tests/src/Functional/PersonContentRenderTest.php
```diff
@@ -143,13 +144,30 @@ public function testPersonPage(): void {
...
-      'Person document test',
-      $files->filter('p')->text()
-    );
+    $files = $content->filter('.bcl-file-container');
 
+    $document_file = File::create([
+      'uri' => $this->getTestFiles('text')[0]->uri,
+    ]);
+    $document_file->save();
+    $absolute_file_url = $document_file->createFileUrl(FALSE);
```
drishu: absolute_file_url is not needed, you can call $document_file->createFileUrl(FALSE); directly in $expected_document

tests/src/Functional/PersonContentRenderTest.php
```diff
@@ -143,13 +144,30 @@ public function testPersonPage(): void {
...
+        'icon' => 'file-text-fill',
+        'url' => $absolute_file_url,
+      ],
+      'translations' => NULL,
+      'link_label' => 'Download',
+    ];
     $this->assertCount(3, $files);
+    $assert = new FilePatternAssert();
+    foreach ($files as $file) {
+      $file_html = $file->ownerDocument->saveHTML($file);
```
drishu: $file should be a Crawler object, if not, use ->eq(), and then you can use $file->html() to get the html

drishu (changes requested):

tests/src/Functional/PersonContentRenderTest.php
tibi2303: done. thanks

tests/src/Functional/PersonContentRenderTest.php
tibi2303: done. thanks

drishu (dismissed):

AaronGilMartinez (dismissed):

AaronGilMartinez (approved):

drishu (approved):

PR #278, opened 2024-05-31
OEL-2604: Change alignment of the sorting options

AaronGilMartinez:


drishu (changes requested):
I dont think a change in the template is needed, I think its rather a css issue.
If I remove `.oe-list-pages-sort-form .form-item{margin:0 0 0 1.5rem}` for instance then it looks ok above the 1200px breakpoint
And then some work on the sort label's and select's margins are needed.

On other list pages its ok because the sort option are smaller, so what if we get much longer ones?

drishu (dismissed):

resources/sass/components/_list_pages.scss
```diff
@@ -17,6 +17,12 @@
   margin-right: 1.5rem;
 }
 
+.oe-list-pages-sort-form .form-select {
+  @media (min-width: 576px) {
```
brummbar: This is not working:
image

brummbar (changes requested):

resources/sass/components/_list_pages.scss
```diff
@@ -1,39 +1,22 @@
...
-  margin-right: 1.5rem;
-}
+  display: flex;
+  justify-content: flex-end;
+  align-items: center;
 
-.oe-list-pages-sort-form .form-select {
-  @media (min-width: 576px) {
-    max-width: calc(100% - 75px);
+  .form-label {
```
drishu: dont think this is ok, and its worse for me now, see
Screenshot 2024-06-05 at 18 15 43

drishu (changes requested):

drishu (dismissed):
The solution looks better now

brummbar:
Still has some issues on some viewports:
image

AaronGilMartinez (commented):
Approving, can't approve my own PR

brummbar (approved):

PR #279, opened 2024-06-04
OEL-2739: AV Portal Video is not sized correctly in featured media pattern

AaronGilMartinez:


templates/overrides/field/field--media--av-portal-video.html.twig
```diff
@@ -0,0 +1,11 @@
+{#
```
brummbar, line 1: If we do this we can remove the workaround at oe_whitelabel_preprocess_paragraph__oe_av_media().

brummbar (dismissed):

templates/overrides/field/field--media--av-portal-video.html.twig
AaronGilMartinez: True

intelektron (approved):

drishu (approved):

PR #284, opened 2024-06-06
OEL-3000: Add (Sub)SectionPattern field group formatter.

donquixote:
Changes:
- New base classes for field group formatters.
- New field group formatter SectionPattern.
- New field group formatter SubSectionPattern.
- New kernel test as alternative for the functional test. (It is faster!)

Up for discussion:
- Old vs new base classes. The new base class has some benefits (imo) but it can create confusion to have multiple options. I think we don't want to remove the old base class, it will create BC issues.
- Do we need the pattern variant option for 'section' pattern? It is only relevant if a sub-theme introduces variants for this pattern.
- Do we want two separate formatters or a single formatter with config? I prefer separate, to make it easier to apply them en masse without going into the settings sub-form. We could use a plugin deriver, but for now I just propose two separate classes.
- Do we need more levels like sub-sub-section?
- Do we want to use this field group formatter for existing cts?
- Should we test this in combination with in-page nav?
- Is `.mb-5` a good choice for the section margin? In our project we use `.my-5`, but we have a trick to clear the top margin (the `.eat-margin-top` trick).


modules/oe_whitelabel_helper/src/Plugin/field_group/FieldGroupFormatter/PatternFormatterBase.php
```diff
@@ -39,13 +32,16 @@ abstract class PatternFormatterBase extends FieldGroupFormatterBase implements C
...
-  public function __construct(array $configuration, string $plugin_id, array $plugin_definition, UiPatternsManager $patterns_manager) {
+  public function __construct(
+    array $configuration,
+    string $plugin_id,
+    array $plugin_definition,
+    protected UiPatternsManager $patternsManager,
+  ) {
     parent::__construct($plugin_id, $plugin_definition, $configuration['group'], $configuration['settings'], $configuration['label']);
-    $this->configuration = $configuration;
-    $this->patternsManager = $patterns_manager;
```
brummbar, line 48: How come removing this? Even if it's not used, classes extending this one might be using these properties.

modules/oe_whitelabel_helper/src/Plugin/field_group/FieldGroupFormatter/PreRenderingFieldGroupFormatterBase.php
```diff
@@ -0,0 +1,75 @@
...
+use Drupal\Core\Render\Element;
+use Drupal\Core\Security\Attribute\TrustedCallback;
+use Drupal\field_group\FieldGroupFormatterBase;
+
+/**
+ * Special base class for field group formatters.
+ *
+ * This is meant for field group formatters that want their logic to run in the
+ * '#pre_render' step, when the final list of visible fields has been correctly
+ * determined.
```
brummbar, line 16: What is the rationale for using vs not using this class?

brummbar (changes requested):
I am not sure I understand the needs for the "better detection of empty elements".
I do not agree on having a dedicated formatter per variation of classes, with tiny differences like attributes/tags.

What I did in the branch POC/OEL-3000 is:
* install ui_patterns_field_group module.
* created an extra field that renders a label only if the section fields are not empty.
* created a field group using the pattern formatter. Moved the fields above like in this picture:

image
With this little piece of code and the configuration in the page, the section was rendered only when the inner fields were not empty. The classes can be customised per field group instance.

The goal of the new patterns that we will be implementing will be to move as much towards this approach.

modules/oe_whitelabel_helper/src/Plugin/field_group/FieldGroupFormatter/PatternFormatterBase.php
donquixote: For the `patternManager`, this is just replaced with constructor property promotion, so it will be ok for extending classes.
The only concern could be calling code using named argument syntax, because the parameter name changes from `$patterns_manager` to `$patternsManager`.
We can remove this change from the PR, I think I just had it somewhere and kept it in.

For the other change, indeed removing `$configuration` would break existing extending classes if they rely on this property. So bad idea, and should not be part of the PR.

In the end I would be ok removing both commits.

modules/oe_whitelabel_helper/src/Plugin/field_group/FieldGroupFormatter/PreRenderingFieldGroupFormatterBase.php
donquixote: ### Emptiness check
Field group has a built-in mechanism to detect emptiness.
However, this mechanism only happens after the individual field group formatters already did their `preRender()` method. This means the `preRender()` method does not know if the field group is empty or not.

The built-in field group formatters like Accordion or Fieldset use dedicated element types and '#theme_wrappers' which expect the content to be direct children of the root element of the group, and in the end nothing is printed if the group has no content.

In the existing pattern-based field group formatters in `oe_whitelabel`, we actually render the content to determine whether it is empty. E.g. in `DescriptionListPattern`, we do this:

```php
      $field_markup = $this->renderer->render($field_element);
      if (trim((string) $field_markup) === '') {
        continue;
      }
```

This probably works most of the time, but it means we are replacing an existing default solution with a custom solution.

### Flexibility and DX
With the `PatternFormatterBase`, a subclass has to implement two methods, `getPatternId()` and `getFields()`, where the `getFields()` has to do the emptiness check in the subclass. The `DescriptionListPattern` even needs to override the `preRender()` method itself, to customize pattern settings.

The `PreRenderingFieldGroupFormatterBase` reduces the complexity of the extending class, while at the same time allowing usage beyond patterns.

The `PreRenderingPatternFormatterBase` brings back the mechanism to choose a pattern variant.

### Can it fully replace the old base class?
I think yes, in theory. But we have to keep the old class around for BC, and even the exiting subclasses we can't simply rewrite, because that would also have a BC impact.

(I think I was somehow involved in the design of the old base class, although I see Sergio as the commit author. So maybe not as much.)

Obviously this would cause a situation where we have different alternative base classes, which can cause confusion.

donquixote:
> I am not sure I understand the needs for the "better detection of empty elements".

For this you can check my other comment..

> I do not agree on having a dedicated formatter per variation of classes, with tiny differences like attributes/tags.

The alternatives would be:
- The `ui_patterns_field_group` module (see below).
- A single field group formatter class with a plugin deriver.
  This makes sense if we have a range of variations, e.g. h2, h3, h4..
- A single field group formatter class with a config option to choose the heading level (which should also set the spacing).
- (In a possible future, similar to what we did in SRB, we would have had a pattern contract for a "content + label" pattern, and the "section" pattern would implement that contract.)

The motivation here was to be able to quickly configure content type displays with many different sections, where a "section pattern" is the natural default formatter. The most common behavior should be achievable without going into the options subform.

The `ui_patterns_field_group` is the opposite of this, and requires lots of maintenance to achieve the basic behavior.
The need for an extra field for the title.. 

> 
> What I did in the branch POC/OEL-3000 is:
> 
> * install ui_patterns_field_group module.
> * created an extra field that renders a label only if the section fields are not empty.

We need to avoid that individual websites need to create extra fields where they have to hard-code the field names.
And I think your emptiness check in the extra field is not even fully reliable, because it does not cover the case where a field has content, but the visitor has no access to it.

> * created a field group using the pattern formatter. Moved the fields above like in this picture:
> [...] 
> With this little piece of code and the configuration in the page, the section was rendered only when the inner fields were not empty.

But this "little piece of code" would have to be replicated across different website projects, for different field groups.

And if you add a new field into the group, now you need to update the field mapping in the group settings, and you need to add the new field in the condition of the extra field.


> The classes can be customised per field group instance.

We should avoid having people customize classes for each field group, if the main use case is to have standardized section and heading levels with standardized spacing.

> 
> The goal of the new patterns that we will be implementing will be to move as much towards this approach.

I think the pattern integration modules shipped with ui_patterns are a dead end, because they have to accommodate a wide range of possible pattern parameter signatures. This always deletes to complex configuration forms, with many degrees of freedom, and no validation to determine if a field mapping makes sense with the parameter type.

----

The goal should be to reduce complexity and unnecessary choice in the website projects,

PR #286, opened 2024-06-12
OEL-2727: _oe_whitelabel_starter_event_preprocess_date disregards site timezone

### LINKED ISSUE #234: _oe_whitelabel_starter_event_preprocess_date disregards site timezone
`$dates->get('start_date')` and `$dates->get('end_date')` have timezone set to UTC, but the sites have Europe/Brussles (as per php timezone set on server) when installed.
The core formatter handles this by fetching the timezone, I think we should do the same, something like
`\Drupal::config('system.date')->get('timezone.default')` and also account for users timezone (will be null if not allowed or not set so we can check for that).
**Steps to reproduce:**

- install oe_whitelabel
- create an event, set start date with His 00:00:00 and end date with his 23:59:59
- visit /nodes
- you will see the start date 1 day earlier than set

The issue becomes bigger if we would display the hour (ex requirement for teaser on one of the subsites)

AaronGilMartinez:


tests/src/Functional/ContentEventRenderTest.php
```diff
@@ -210,6 +213,91 @@ public function testEventRenderingTeaser(): void {
...
+      'end_year' => '2024',
+      'date_time' => '2024-06-07',
+    ];
+
+    $build = $builder->view($node, 'teaser');
+
+    $html = (string) $this->container->get('renderer')->renderRoot($build);
+
+    $card_assert->assertVariant('search', $html);
+    $card_assert->assertPattern($expected, $html);
```
drishu, line 352: Lets do 1 more thing, disable the user configurabele timezone, check that even though the user has a different time zone, the user still sees the sites timezone

AaronGilMartinez: Done, also added check to verify that user timezone overrides site timezone when logged in.

modules/oe_whitelabel_starter_event/oe_whitelabel_starter_event.module
```diff
@@ -186,26 +186,33 @@ function _oe_whitelabel_starter_event_preprocess_date(array &$variables): void {
...
 
+  // Displayed date depends on the timezone.
+  $cache = CacheableMetadata::createFromRenderArray($variables);
+  $cache->addCacheContexts(['timezone']);
+  $cache->applyTo($variables);
+
   $dates = $node->get('oe_sc_event_dates')->first();
-  /** @var \Drupal\Core\Datetime\DrupalDateTime $start */
-  $start = $dates->get('start_date')->getValue();
+  $start = $dates->get('start_date')->getValue()->getTimestamp();
```
intelektron, line 195: The second way to do it may be 

```php
use Drupal\Core\Datetime\DrupalDateTime;
// ...
$start = new DrupalDateTime($dates->get('start_date')->value);
// ...
'year' => $start->format('Y'),
```

But I am not sure if it is much better

intelektron (approved):
Approving with a small notice

modules/oe_whitelabel_starter_event/oe_whitelabel_starter_event.module
drishu: there is also a 3rd variant with `$start->setTimeZone(timezone_open($timezone));` like done in DateTimeFormatterBase, but I don't mind using a service.

drishu (approved):

PR #288, opened 2024-06-12
EPHEH-465: Refactor color scheme injection.

drishu:
After the showcase in figma, it became clear that just background parameter would not be sufficient as the banner variants need diferent combinations of classes from the new color scheme api.

### DIFF
--- a/modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.module
+++ b/modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.module
@@ -522,10 +522,6 @@ function oe_whitelabel_preprocess_paragraph__oe_facts_figures(array &$variables)
  * Implements hook_preprocess_paragraph() for oe_carousel paragraph.
  */
 function oe_whitelabel_preprocess_paragraph__oe_carousel(array &$variables): void {
-  /** @var \Drupal\oe_whitelabel\ColorSchemePreprocess $color_scheme_preprocess */
-  $color_scheme_preprocess = \Drupal::classResolver(ColorSchemePreprocess::class);
-  $color_scheme_preprocess->injectColorScheme($variables);
-
   /** @var \Drupal\paragraphs\Entity\Paragraph $paragraph */
   $paragraph = $variables['paragraph'];
   $variables['items'] = [];
@@ -686,10 +682,6 @@ function oe_whitelabel_paragraphs_preprocess_paragraph__oe_gallery(&$variables) 
   // title tag, we set it up here so that further preprocess hooks can change
   // it.
   $variables['title_tag'] = 'h2';
-
-  /** @var \Drupal\oe_whitelabel\ColorSchemePreprocess $color_scheme_preprocess */
-  $color_scheme_preprocess = \Drupal::classResolver(ColorSchemePreprocess::class);
-  $color_scheme_preprocess->injectColorScheme($variables);
 }
 
 /**
--- a/modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.module
+++ b/modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.module
@@ -68,6 +68,10 @@ function oe_whitelabel_paragraphs_theme_suggestions_field_alter(array &$suggesti
  * Implements hook_preprocess_paragraph().
  */
 function oe_whitelabel_preprocess_paragraph__oe_links_block(array &$variables): void {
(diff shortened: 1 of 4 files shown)


modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.module
```diff
@@ -68,6 +68,12 @@ function oe_whitelabel_paragraphs_theme_suggestions_field_alter(array &$suggesti
  * Implements hook_preprocess_paragraph().
  */
 function oe_whitelabel_preprocess_paragraph__oe_links_block(array &$variables): void {
+  /** @var \Drupal\oe_whitelabel\ColorSchemePreprocess $color_scheme_preprocess */
+  $color_scheme_preprocess = \Drupal::classResolver(ColorSchemePreprocess::class);
```
drishu, line 72: I've updated the list of paragraphs that recive or not the color scheme and have or not background color to match the requirements

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.module
```diff
@@ -272,12 +278,19 @@ function oe_whitelabel_preprocess_paragraph__oe_banner(array &$variables): void
   $variant = $paragraph->get('oe_paragraphs_variant')->value ?? 'default';
   $variables['variant'] = str_replace('oe_banner_', '', $variant);
 
-  if ($variables['variant'] === 'default') {
-    $color_scheme_preprocess->injectColorScheme($variables);
-  }
-  else {
-    $color_scheme_preprocess->injectColorScheme($variables, FALSE);
-  }
+  match($variables['variant']) {
```
drishu, line 281: I had to complicate things here so that the correct color contrasts are applied (for example on the shaded banner, the best text color contrast is with white always)
Screenshot 2024-06-12 at 16 51 28

modules/oe_whitelabel_paragraphs/tests/src/Kernel/Paragraphs/ColorSchemeTest.php
```diff
@@ -79,10 +79,8 @@ public function testColorSchemeInParagraphs(): void {
       $html = $this->renderParagraph($paragraph);
       $crawler = new Crawler($html);
 
-      $element = $crawler->filter($data['wrapper_selector'] . '.foo-bar.text-color-default');
+      $element = $crawler->filter($data['wrapper_selector'] . '.foo-bar');
       $this->assertCount(1, $element, sprintf('Element "%s" has color scheme applied.', $data['wrapper_selector']));
-      $has_background = (bool) $crawler->filter('.bg-default')->count();
-      $this->assertTrue($data['has_background'] === $has_background, sprintf('Element "%s" does not have ".bg-default" class.', $data['wrapper_selector']));
```
drishu, line 85: had to delete the negative testing because of the increased complexity of the injectColorScheme "API"

src/ColorSchemePreprocess.php
```diff
@@ -16,19 +17,58 @@ class ColorSchemePreprocess {
...
     }
 
-    $variables['attributes']['class'][] = Html::getClass($variables['elements']['#oe_color_scheme']);
-    $variables['attributes']['class'][] = 'text-color-default';
+    $variables['attributes'] = new Attribute($variables['attributes'] ?? []);
+    $variables['attributes']->addClass(Html::getClass($variables['elements']['#oe_color_scheme']));
 
-    if ($background) {
-      $variables['attributes']['class'][] = 'bg-default';
+    if (!empty($options['text_colored'])) {
```
drishu: would like to make sure that its bool, could do `(isset && === TRUE)` but like this the complexity is lower as empty checks isset and the value, downside is `0`, 'FALSE`, `''` are interpreted as false. 
Was contemplating also the symfony options resolver but seems like bit much for a preprocess

intelektron: I have some second thoughts here:

1. If somebody makes a typo in the option id, they will never know, there is no validation. Maybe it would be good to introduce constants for options (like `const PRIMARY_BACKGROUND = 'primary_background`), or an `enum`, or defining a list of allowed options and throwing an `InvalidArgumentException`
2. Using `empty` is discouraged by the latest trends, and if we ever introduce PHPStan to the pipeline, it will start to throw errors. Perhaps `TRUE === $options['text_colored'] ?? FALSE` would be more readable? Or maybe we should define an array of default options, so they are always set.

intelektron (approved):

tibi2303 (approved):

PR #291, opened 2024-06-26
OEL-2728: Added translatability fix.

tibi2303:


AaronGilMartinez (changes requested):
Changes look ok, drone is failing and branch is outdated, but I think updating the branch will fix drone.

While testing found an issue that can be investigated in a follow up ticket.
The issue is present for Listing item block description:
image

AaronGilMartinez (approved):

PR #300, opened 2024-08-07
OEL-1686: Updated tests regarding bcl-offcanvas duplicate class.

tibi2303:


composer.json
```diff
@@ -27,6 +27,7 @@
         "drupal/extra_field": "^2.3",
         "drupal/facets_form": "^1.0.0-alpha6",
         "drupal/field_group": "^3.4",
+        "drupal/paragraphs": "1.17",
```
drishu:
```suggestion
        "drupal/paragraphs": "~1.17.0",
```
we need to lock it in such a way that it still permits patch releases

composer.json
```diff
@@ -107,7 +108,8 @@
             ]
         },
         "_readme": [
-            "Explicit requirement of nikic/php-parser ^4 as later versions are not compatible with grumphp @see https://github.com/phpro/grumphp/issues/1119"
+            "Explicit requirement of nikic/php-parser ^4 as later versions are not compatible with grumphp @see https://github.com/phpro/grumphp/issues/1119",
+            "Lock drupal/paragraphs to 1.17.x to apply patch correctly"
```
drishu: I dont understand from this message the exact reason, which patch?

drishu (changes requested):

drishu (approved):

AaronGilMartinez (approved):

PR #307, opened 2024-08-27
OEL-3315: Remove empty title wrapper.

### LINKED ISSUE #302: Empty div with margin in Banner when paragraph has no title
#### Problem
The Banner paragraph type's Title field is [not required](https://github.com/openeuropa/oe_paragraphs/blob/master/modules/oe_paragraphs_banner/config/install/field.field.paragraph.oe_banner.field_oe_title.yml#L13). When the field is left empty, there is an empty div with some bottom margin in the output.

Tested on version 1.8.0

oe_whitelabel_banner_paragraph_without_title

#### Reason
The banner paragraph template [passes the title](https://github.com/openeuropa/oe_whitelabel/blob/1.8.0/templates/paragraphs/paragraph--oe-banner.html.twig#L31) as the field render array, that is not null when the field is empty (contains cache metadata for example).

#### Proposed solution
If the Title field is empty, null should be passed as pattern title. This can be achieved by using `field_value` filter.

```twig
{{ pattern('banner', {
  'variant': variant,
  'title': content.field_oe_title|field_value,
```

AaronGilMartinez:


drishu (dismissed):
The PR follows our usual approach for resolving the optional title, so its ok for me

modules/oe_whitelabel_paragraphs/tests/src/Kernel/Paragraphs/MediaParagraphsTest.php
```diff
@@ -672,13 +679,33 @@ public function testBanner(): void {
...
-    $this->assertStringContainsString('Example', trim($crawler->filter('a.btn')->text()));
-    $this->assertStringContainsString('#chevron-right', trim($crawler->filter('a.btn')->html()));
+  protected function assertBannerRendering(Crawler $crawler, array $expected_values = []): void {
+    $values = array_merge([
+      'title' => 'Banner',
+      'description' => 'Description',
+      'link_label' => 'Example',
+    ], $expected_values);
+    // Check optional title.
+    if (!empty($values['title'])) {
```
brummbar: empty() with `''` plays a bit weird as we all know. You are passing title = '', but it will actually do the empty scenario.
Let's clear this up: use NULL comparison, we need to pass NULL if we don't expect a title.

brummbar (changes requested):

modules/oe_whitelabel_paragraphs/tests/src/Kernel/Paragraphs/MediaParagraphsTest.php
AaronGilMartinez: Replaced

brummbar: You still need to change !empty with === NULL.

drishu (approved):

drishu (approved):

brummbar (approved):

PR #311, opened 2024-09-13
OEL-2289: Added border-left on the search input in header.

tibi2303:


drishu (approved):

donquixote:
This change seems absolutely correct and achieves the result.

The only question I could have is why this class was added originally. Maybe there is a use case where we need this class?
But I think it is just a leftover.
For the record, it was introduced here, https://github.com/openeuropa/oe_whitelabel/pull/136/commits/0ea1a7bfbc48021bbef7d9a2dba7c21c259c1128#diff-22ce6c40b73de01104a912a6b2eabefd32912f3daa5c2258c686eca60665c456R12

donquixote (approved):
Seems good to me and it works.

PR #314, opened 2024-09-27
OEL-3411: Set default button label in search block header, if not already set.

tibi2303:


modules/oe_whitelabel_search/oe_whitelabel_search.post_update.php
```diff
@@ -19,3 +19,16 @@ function oe_whitelabel_search_post_update_00001(&$sandbox) {
   $block->set('settings', $settings);
   $block->save();
 }
+
+/**
+ * Set default button label in search block if it's not already set.
+ */
+function oe_whitelabel_search_post_update_00031(&$sandbox) {
```
AaronGilMartinez: I think update number is not ok.

modules/oe_whitelabel_search/oe_whitelabel_search.post_update.php
```diff
@@ -19,3 +19,16 @@ function oe_whitelabel_search_post_update_00001(&$sandbox) {
...
+  $block = Block::load('oe_whitelabel_search_form');
+  $settings = $block->get('settings');
+  if (empty($settings['button']['label'])) {
+    $settings['button']['label'] = 'Search';
+  }
+  $block->set('settings', $settings);
+  $block->save();
```
AaronGilMartinez: We are saving the block even if we didn't change the settings.

modules/oe_whitelabel_search/oe_whitelabel_search.post_update.php
```diff
@@ -19,3 +19,16 @@ function oe_whitelabel_search_post_update_00001(&$sandbox) {
   $block->set('settings', $settings);
   $block->save();
 }
+
+/**
+ * Set default button label in search block if it's not already set.
+ */
+function oe_whitelabel_search_post_update_00031(&$sandbox) {
+  $block = Block::load('oe_whitelabel_search_form');
```
AaronGilMartinez: We need to set this setting in current theme search block instances rather than the block provided by the theme, sites will have its own configuration for the search block as oesa does with oe_showcase_theme.
To do so the logic is the same but we have to:

1. Get blocks from entity type manager, and filter theme (current) and plugin ('whitelabel_search_block').
2. Check if there is value, if its we skip, if not we set the value.

AaronGilMartinez (changes requested):

AaronGilMartinez (dismissed):

drishu (dismissed):
If the subsite doesnt want the improvement in is config and can be easly manged

AaronGilMartinez (approved):

drishu (approved):

PR #316, opened 2024-09-30
OEL-3412: Added cck before page_top.

tibi2303:


drishu (changes requested):
I remember we chose the option to show the wrapper div permanently, check what html attribute has to be added so that it does not show as a accessibility issue of empty-div/hidden content. Also first try to inject it via a preprocess instead of overriding templates

tibi2303:
done, thanks andras.. it's better like this.

oe_whitelabel.theme
```diff
@@ -155,6 +155,23 @@ function oe_whitelabel_preprocess_page(&$variables) {
...
+/**
+ * Implements hook_preprocess_html().
+ */
+function oe_whitelabel_preprocess_html(array &$variables) {
+  $cck_container = [
+    '#type' => 'html_tag',
+    '#tag' => 'div',
+    '#attributes' => [
+      'id' => 'cck_here',
+      'role' => 'alertdialog',
```
drishu: lets stick to the ticket as discussed in the daily and use "alert" and remove the "aria-live"
"aria-live="assertive" should only be used for time-sensitive/critical notifications that absolutely require the user's immediate attention. " so dont think it is what we need

oe_whitelabel.theme
```diff
@@ -155,6 +155,23 @@ function oe_whitelabel_preprocess_page(&$variables) {
...
+      'aria-live' => 'assertive',
+    ],
+  ];
+
+  $variables['page_top']['cck_container'] = $cck_container;
```
drishu, line 171: make sure its the first one

drishu (changes requested):

oe_whitelabel.theme
tibi2303: it's not possible to use unshift because the drupal tests will fail on the admin toolbar not finding specific elements.

drishu (dismissed):

tests/src/Traits/CckContainerTrait.php
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Contains methods to assert the presence of specific blocks and elements.
+ */
+trait CckContainerTrait {
+
+  /**
+   * Asserts the HTML of the "cck_here" container.
+   */
+  protected function assertCckContainer(): void {
+    $cck_container = $this->cssSelect('#cck_here')[0];
```
AaronGilMartinez:
```suggestion
    $cck_container = $this->cssSelect('body :nth-child(2)')[0];
```

AaronGilMartinez: This way we check the location too, we are moving to second place inside body.
Since we are doing the check in the setUp, toolbar shouldn't be a problem, unless we log in with a user before calling the parent setUp.

AaronGilMartinez (changes requested):

tests/src/Traits/CckContainerTrait.php
tibi2303: done, indeed, thanks

AaronGilMartinez (approved):

drishu (approved):

PR #317, opened 2024-10-04
OEL-3418: Changed list title for listing and search.

tibi2303:


AaronGilMartinez (dismissed):

templates/list_pages/node--oe-list-page--full.html.twig
```diff
@@ -25,10 +25,10 @@
     <div class="col-12 col-lg-9">
       <div class="row">
         <div class="col-md-6 col-xl-8 align-self-center">
-          <h4 class="mb-0">
```
drishu, line 28: it is ok to fix the tags, but we are breaking the page visually, we need to ensure that it looks as before

drishu (changes requested):

drishu (approved):

PR #320, opened 2024-10-08
OEL-3420: Added check for rich text paragraph title.

tibi2303:


templates/paragraphs/paragraph--oe-rich-text.html.twig
```diff
@@ -6,7 +6,9 @@
  * @see ./modules/contrib/paragraphs/templates/paragraph.html.twig
  */
 #}
-{% include '@oe-bcl/bcl-heading/heading.html.twig' with {
-  title: content.field_oe_title,
-} only %}
+{% if content.field_oe_title is not empty %}
```
drishu: use the |field_value filter to avoid render meta giving false negatives

drishu (changes requested):
We can add a few lines in the paragraphs test to assert the absense of the wrapper and presence when needed.

templates/paragraphs/paragraph--oe-rich-text.html.twig
tibi2303: done

modules/oe_whitelabel_paragraphs/tests/src/Kernel/Paragraphs/ParagraphsTestBase.php
```diff
@@ -46,6 +46,7 @@ abstract class ParagraphsTestBase extends AbstractKernelTestBase {
     'paragraphs',
     'text',
     'typed_link',
+    'twig_field_value',
```
drishu: remove

modules/oe_whitelabel_paragraphs/tests/src/Kernel/Paragraphs/RichTextTest.php
```diff
@@ -108,7 +108,7 @@ public function testRendering(): void {
       'second option',
```
drishu, line 115: add 
```
/**
   * {@inheritdoc}
   */
  protected static $modules = [
    'twig_field_value',
  ];
```
the reason is that only this one needs it, lets start to be more efficient, the tests are slow enough

drishu (changes requested):

modules/oe_whitelabel_paragraphs/tests/src/Kernel/Paragraphs/ParagraphsTestBase.php
tibi2303: removed and added on the necessary tests.

drishu (approved):

PR #321, opened 2024-10-09
OEL-3419: Added roles on nav items in header.

tibi2303:


templates/overrides/page/page.html.twig
```diff
@@ -79,7 +79,9 @@
           right: page.header_top,
           color_scheme: 'light',
           background: false,
-          attributes: create_attribute().addClass(['w-100', 'shadow-sm']),
+          attributes: create_attribute()
```
drishu: remember the prefferred format of 
```
create_attribute([
  'class': ...

drishu (changes requested):
lets update the banch and wait for green drone

templates/overrides/page/page.html.twig
tibi2303: done

drishu (approved):

PR #324, opened 2024-10-14
OEL-3422: Added aria-label on language switcher link.

tibi2303:


templates/overrides/navigation/links--language-block.html.twig
```diff
@@ -9,31 +9,26 @@
...
-) %}
-  {% set title = _language.link.label %}
-  {% set link_title %}
-    <span class="d-block d-lg-inline-block">{{ title }}</span>
-  {% endset %}
+{% set extra_attributes = create_attribute({
+  'class': ['text-center'],
+  'data-bs-toggle': 'modal',
+  'data-bs-target': '#' ~ _language_link.target,
+  'aria-label': ('Change language. Current language is '|t) ~ _language_link.label ~ '.',
```
drishu: this is not the way to use the t function, please add a placehoder and set _language_link.label as parameter

tests/src/Kernel/MultilingualBlockTest.php
```diff
@@ -87,6 +87,7 @@ public function testBlockRendering(): void {
     $link = $crawler->filter('div.language-switcher > a');
     $this->assertSame('English', trim($link->text()));
     $this->assertSame('#', $link->attr('href'));
+    $this->assertSame('Change language. Current language is English.', $link->attr('aria-label'));
```
drishu: set another language and assert the language changes

drishu (changes requested):

tests/src/Kernel/MultilingualBlockTest.php
tibi2303: added

templates/overrides/navigation/links--language-block.html.twig
tibi2303: yes, thanks

drishu (approved):

PR #328, opened 2024-11-07
OEL-2118: Fix AddressInlineFormatter php notices

drishu:
Fix AddressInlineFormatter php notices when displaying the formatter settings form.


tibi2303:
on my side seems okay, i will approve and i also change the commit message. it had a wrong task id.

tibi2303 (approved):

PR #331, opened 2024-12-09
OEL-3503: Support slim select V2

piotrsmykaj:


modules/oe_whitelabel_helper/oe_whitelabel_helper.module
```diff
@@ -92,7 +92,13 @@ function oe_whitelabel_helper_requirements_alter(array &$requirements): void {
   if (isset($requirements['slim_select_library'])) {
     $theme_handler = \Drupal::service('theme_handler');
     $theme_path = $theme_handler->getTheme('oe_bootstrap_theme')->getPath();
-    $path = '/' . $theme_path . '/assets/js/slimselect.min.js';
+    $version = \Drupal::config('slim_select.settings')->get('version');
```
drishu: Let's move all sim select code into a class under /source SlimSelectProcess (or if u have a better idea). Mark it as internal. Also this part that you added for version 2 repeats, so you can add it into a private method.

modules/oe_whitelabel_helper/oe_whitelabel_helper.module
```diff
@@ -92,7 +92,13 @@ function oe_whitelabel_helper_requirements_alter(array &$requirements): void {
...
+    if (str_starts_with($version, '2.')) {
```
drishu: the saved version starts with v
```suggestion
    if (str_starts_with($version, 'v2.')) {
```

drishu (changes requested):

modules/oe_whitelabel_helper/oe_whitelabel_helper.module
piotrsmykaj: ok please check my changes.

modules/oe_whitelabel_helper/src/SlimSelectHelper.php
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * Alter requirements for the Slim Select.
+   *
+   * @param array $requirements
+   *   The list of requirements.
+   */
+  public function alterRequirements(array &$requirements) {
+    if (isset($requirements['slim_select_library'])) {
+      $path = \Drupal::service('oe_whitelabel_helper.slim_select')->getJsFilePath();
```
drishu:
```suggestion

```

modules/oe_whitelabel_helper/src/SlimSelectHelper.php
```diff
@@ -0,0 +1,74 @@
...
+      $requirements['slim_select_library'] = [
+        'title' => t('Slim Select library'),
+        'severity' => REQUIREMENT_OK,
+        'value' => t('Library available at :path.', [
+          ':path' => $path,
```
drishu:
```suggestion
          ':path' => $this->getJsFilePath(),
```

modules/oe_whitelabel_helper/src/SlimSelectHelper.php
```diff
@@ -0,0 +1,74 @@
...
+   * @param array $libraries
+   *   The list of libraries.
+   * @param string $extension
+   *   The extension.
+   */
+  public function libraryInfoAlter(array &$libraries, string $extension) {
+    if ('slim_select' !== $extension) {
+      return;
+    }
+    $path = \Drupal::service('oe_whitelabel_helper.slim_select')->getJsFilePath();
```
drishu:
```suggestion

```

modules/oe_whitelabel_helper/src/SlimSelectHelper.php
```diff
@@ -0,0 +1,74 @@
...
+    $libraries['slim.select']['js'] = [
+      $path => [
```
drishu:
```suggestion
      $this->getJsFilePath() => [
```

modules/oe_whitelabel_helper/src/SlimSelectHelper.php
```diff
@@ -0,0 +1,74 @@
...
+ */
+class SlimSelectHelper {
+
+  /**
+   * Alter requirements for the Slim Select.
+   *
+   * @param array $requirements
+   *   The list of requirements.
+   */
+  public function alterRequirements(array &$requirements) {
```
drishu:
```suggestion
  public function alterRequirements(array &$requirements): void {
```

modules/oe_whitelabel_helper/src/SlimSelectHelper.php
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * Alter a library info for the SLim Select.
+   *
+   * @param array $libraries
+   *   The list of libraries.
+   * @param string $extension
+   *   The extension.
+   */
+  public function libraryInfoAlter(array &$libraries, string $extension) {
```
drishu:
```suggestion
  public function libraryInfoAlter(array &$libraries, string $extension): void {
```

modules/oe_whitelabel_helper/src/SlimSelectHelper.php
```diff
@@ -0,0 +1,74 @@
...
+    unset($libraries['slim.select']['css']);
+  }
+
+  /**
+   * Get the Slim Select library JS path.
+   *
+   * @return string
+   *   The Slim Select JS path.
+   */
+  private function getJsFilePath() {
```
drishu:
```suggestion
  private function getJsFilePath(): string {
```

modules/oe_whitelabel_helper/src/SlimSelectHelper.php
```diff
@@ -0,0 +1,74 @@
...
+    $theme_handler = \Drupal::service('theme_handler');
```
drishu: here we need to use injection here, let the class implement ContainerInjectionInterface and the classResolver will take care

modules/oe_whitelabel_helper/oe_whitelabel_helper.services.yml
```diff
@@ -4,3 +4,5 @@ services:
     arguments: ['@plugin.manager.block']
     tags:
       - { name: twig.extension }
+  oe_whitelabel_helper.slim_select:
```
drishu: A few thoughts about this:
-  services should be reusable classes (as in can be used anywhere), this class is specific for the requirements alter and the info alter, so it is not suited for a service.
- we dont want this to be used as like an api so it should have been marked with `public: false` but that brings a whole different set of problems when trying to use in a preprocess, so I think we should drop the service part.

We still want code reuse and OOP so I thought more along the lines of a helper class in oe_whitelabel/src (see we have 2 more there), And then we call the class with \Drupal::classResolver(SlimSelectHelper::class);

drishu (changes requested):

src/SlimSelectHelper.php
```diff
@@ -0,0 +1,107 @@
...
+    if ($this->themeHandler->themeExists($theme_name)) {
+      $theme_path = $this->themeHandler->getTheme($theme_name)->getPath();
+      $version = \Drupal::config('slim_select.settings')->get('version');
+      if (str_starts_with($version, 'v2.')) {
+        return '/' . $theme_path . '/assets/js/slim-select-2/slimselect.min.js';
+      }
+      return '/' . $theme_path . '/assets/js/slimselect.min.js';
+    }
+
+    return '';
```
piotrsmykaj, line 96: @drishu I added the condition to check if the theme exists.

modules/oe_whitelabel_helper/oe_whitelabel_helper.services.yml
piotrsmykaj: oky good idea, it's done

modules/oe_whitelabel_helper/src/SlimSelectHelper.php
piotrsmykaj: done

oe_whitelabel.theme
```diff
@@ -239,24 +239,7 @@ function _oe_whitelabel_process_element_select(array &$element): array {
...
-    $path => [
-      'minified' => TRUE,
-      'attributes' => [
-        'defer' => TRUE,
-      ],
-    ],
-  ];
-  // The slim.select css is already present in the parent theme.
-  unset($libraries['slim.select']['css']);
+  \Drupal::service('oe_whitelabel_helper.slim_select')->libraryInfoAlter($libraries, $extension);
```
drishu: \Drupal::classResolver

src/SlimSelectHelper.php
```diff
@@ -0,0 +1,107 @@
...
+  protected $themeHandler;
+
+  /**
+   * Constructs a SlimSelectHelper object.
+   *
+   * @param \Drupal\Core\Extension\ThemeHandlerInterface $themeHandler
+   *   The theme handler.
+   */
+  public function __construct(ThemeHandlerInterface $themeHandler) {
+    $this->themeHandler = $themeHandler;
```
drishu: Lets go for the modern property promotion

drishu (changes requested):
I'm stopping my review here as I found a blocking issue for me. I've created and used a tag for oebt 0.3503.202502031520, so that I have the assets built, and indeed the slimselect 2 folder is present, but when I create a multiselect field I get this:
image

I've asked @tibi2303 to look into the bcl releases but something is defenetly wrong.

oe_whitelabel.theme
piotrsmykaj: done

piotrsmykaj:
> I'm stopping my review here as I found a blocking issue for me.

image

@drishu @tibi2303 with BCL 0.3501.202502031900 it looks on my local as you can see on the screenshot above. At the HEH project, we applied some customizations and arrows seems fine with it: https://git.fpfis.tech.ec.europa.eu/ep/ep-historia-museum-dev/-/blob/master/lib/themes/ep_historia_museum/src/scss/components/bcl_theme/_multiselect.scss?ref_type=heads#L44

I am wondering why the style from slim select is not applied here because multi select elements have also a class form-select:
 
https://preview-630--oelibrary.netlify.app/default/iframe.html?id=features-group--content-list&viewMode=story ?
 
I noticed this:
Could not load content for https://preview-630--oelibrary.netlify.app/node_modules/slim-select/src/slim-select/slimselect.scss (HTTP error: status code 404, net::ERR_HTTP_RESPONSE_CODE_FAILURE)

drishu (approved):
Approved, lets not forget to update oebt before merge

drishu (dismissed):

drishu (approved):

PR #336, opened 2025-02-27
OEL-3081: Add the theme option to apply Slim select to all elements

### LINKED ISSUE #265: Make slim_select work for single-value select elements
Currently, oe_whitelabel adds slim_select for multiple-value select elements.

See `_oe_whitelabel_process_element_select()`.
https://github.com/openeuropa/oe_whitelabel/blob/3ec646ddd82fe1938bbab66d5ebd4dc0c322ee93/oe_whitelabel.theme#L210-L220

The slim_select library can also be useful for single-value selects with many options.

Currently there is one obstacle:
The BCL targets a class `'placeholder'` which is also used in slim_select, causing an undesirable side effect.

Other than that, we need to decide if it is safe to apply it to all single-value selects, or if we need to decide it case by case.

piotrsmykaj:


AaronGilMartinez (dismissed):

AaronGilMartinez:
The branch will need rebase before merging.

piotrsmykaj:
ok rebased and also removed the dev version from the composer.json
`"openeuropa/oe_bootstrap_theme": "0.3081.202503271630"`

AaronGilMartinez (approved):

PR #340, opened 2025-03-12
OEL-3341: Handle link access in banner paragraph button

piotrsmykaj:


modules/oe_whitelabel_paragraphs/tests/src/Kernel/Paragraphs/MediaParagraphsTest.php
```diff
@@ -672,6 +688,30 @@ public function testBanner(): void {
...
+      'uri' => 'entity:node/' . $node->id(),
+      'title' => 'Example node',
+    ]);
+    $paragraph->save();
+    $html = $this->renderParagraph($paragraph);
+    $crawler = new Crawler($html);
+    $this->assertStringContainsString('Example node', trim($crawler->filter('a.btn')->text()));
+    $node->setUnpublished();
+    $node->save();
+    \Drupal::entityTypeManager()->getAccessControlHandler('node')->resetCache();
```
AaronGilMartinez: I don't like the idea of reseting cache when we actually made changes to handle the access with cache, this will prevent the test to fail in case something is not ok with the cache.

Also, the renderParagraph() function is rendering root which is ok for isolated calls but not for rendering taking into account previous contexts.

If you remove any code related to cache the test will pass anyway.

We cannot check agaisng possible tags because entity tags aren't added with the URL access as cache dependency.

Other option would be use functional test, but the cache dependecy added by the URL is the context 'user.permissions', and that will be added anyway.

In conclusion, I think is better just to check that the link is hidden for the user without access and forget about testing the cache.

Other checks in the test have cache depedencies that are not tested.

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.module
```diff
@@ -614,8 +614,25 @@ function _oe_whitelabel_set_banner_link(Paragraph $paragraph, array &$variables)
...
-  $variables['url'] = $link->getUrl();
+  if (!$link) {
+    return;
+  }
+
+  $cacheability = CacheableMetadata::createFromRenderArray($variables);
+
+  // Run access checks on the URL.
+  $url = $link->getUrl();
+  $access = $url->access($variables['user'], TRUE);
```
AaronGilMartinez:
```suggestion
  $access = $url->access(return_as_object: TRUE);
```

modules/oe_whitelabel_paragraphs/tests/src/Kernel/Paragraphs/MediaParagraphsTest.php
```diff
@@ -42,6 +52,12 @@ protected function setUp(): void {
     // Call the install hook of the Media module.
     $this->container->get('module_handler')->loadInclude('media', 'install');
     media_install();
+
+    $this->createContentType([
+      'type' => 'paragraphs_test',
+      'name' => 'Paragraphs Test',
+    ]);
+    $this->setCurrentUser($this->createUser(['access content']));
```
AaronGilMartinez: Since we are not running access checks in other tests I would move this to the link access verification part.
This will be executed for all the tests, we only need it there.

AaronGilMartinez (changes requested):

modules/oe_whitelabel_paragraphs/tests/src/Kernel/Paragraphs/MediaParagraphsTest.php
piotrsmykaj: removed user access and testing it as anonymous, as you suggested in another comment

AaronGilMartinez (approved):

PR #358, opened 2025-05-23
OEL-3817: Preserve the query parameters when submitting the SearchForm.

### LINKED ISSUE #242: Preserve the query parameters when submitting the SearchForm
In search pages, when submitting the [SearchForm](https://github.com/openeuropa/oe_whitelabel/blob/1.x/modules/oe_whitelabel_search/src/Form/SearchForm.php), other query parameters (e.g. facets filters and sorting) are getting lost.

**Steps to reproduce:**
- Go to https://digit-oesa.prd.fpfis.tech.ec.europa.eu/en/search
- Select a Category and click Refine.
- Select Change Sort by e.g. A-Z
-  Now, use the free text search box and submit

**Expected result:**
- I expect that the category and sorting criteria are preserved and applied.

**Actual result:**
- I lose the previously applied filters and sorting criteria

**Proposed solution:**
Preserve all query parameters when submitting the SearchForm.
```
  /**
   * {@inheritdoc}
   */
  public function submitForm(array &$form, FormStateInterface $form_state): void {
    $config = $form_state->get('oe_whitelabel_search_config');
    $url = Url::fromUri('base:' . $config['form']['action'], [
      'language' => $this->languageManager->getCurrentLanguage(),
      'absolute' => TRUE,
      'query' => $this->requestStack->getCurrentRequest()->query->all(),
    ]);
    $url->mergeOptions([
      'query' => [
        $config['input']['name'] => $form_state->getValue('search_input'),
      ],
    ]);
    $form_state->setRedirectUrl($url);
  }
```
Lets discuss it  if you agree with the approach.

enriquelacoma:


nagyad:
The "Merge branch '1.x' into OEL-3817" commit should not be there, it will pollute the commit list. Please remove it. Make sure to do rebase next time.

modules/oe_whitelabel_search/tests/src/Functional/SearchFormTest.php
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  protected $defaultTheme = 'oe_whitelabel';
+
+  /**
+   * Test the search form keeps query parameters.
+   */
+  public function testSearchQueryParams() {
```
AaronGilMartinez, line 32: The test covers a part of the form functionality, the one related to parameters. I think would be nice to add a todo to complete the testing coverage and other options for the form like language, the action, and not sure about search_api autocomplete.

modules/oe_whitelabel_search/tests/src/Functional/SearchFormTest.php
```diff
@@ -0,0 +1,50 @@
...
+ * Search form test.
+ */
+class SearchFormTest extends BrowserTestBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected static $modules = [
+    'block',
+    'oe_bootstrap_theme_helper',
```
AaronGilMartinez, line 20: This is not related with the ticket but It seemed odd to have this module here since is not a dependency for search module. After a bit of research it seems that oe_whitelabel uses a twig filter provided by oe_bootstrap_theme_helper. This shouldn't be fixed here, not scope of the issue.

AaronGilMartinez (changes requested):

modules/oe_whitelabel_search/tests/src/Functional/SearchFormTest.php
enriquelacoma: @AaronGilMartinez I think that's outside the scope of the issue and should be done on another issue

AaronGilMartinez (approved):

PR #361, opened 2025-06-03
ELA-1301: Fix search block redirection.

### LINKED ISSUE #360: Search block form wrong redirection
Release 1.18.0 introduced an issue to the redirection of the search block form.

In case the form action is configured with a prepended "/" slash, it ends up having two slashes "//" which drupal Url::fromUserInput does not interpret correctly.

Image

I think there are a couple of things we could do here, but it depends on what we want to achieve.

Are there case scenarios where the search block form should redirect to an external third party website? Ideally, but require the more work, we could replace the textfield for the form action by a select list of views, to provide the route to Url::fromRoute().

I will create a PR with the easiest way of fixing this (without having to execute a hook_update).

Aporie:


modules/oe_whitelabel_search/src/Plugin/Block/SearchBlock.php
```diff
@@ -62,12 +69,15 @@ class SearchBlock extends BlockBase implements ContainerFactoryPluginInterface {
...
    * @param \Drupal\Core\Extension\ModuleHandlerInterface $module_handler
    *   The module handler.
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   The entity type manager.
    */
-  public function __construct(array $configuration, $plugin_id, $plugin_definition, ConfigFactoryInterface $config_factory, FormBuilderInterface $form_builder, ModuleHandlerInterface $module_handler) {
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, ConfigFactoryInterface $config_factory, FormBuilderInterface $form_builder, ModuleHandlerInterface $module_handler, EntityTypeManagerInterface $entity_type_manager) {
     parent::__construct($configuration, $plugin_id, $plugin_definition);
     $this->configFactory = $config_factory;
     $this->formBuilder = $form_builder;
```
Aporie, line 69: Please note that formBuilder and configFactory are two services not use in the block. But I've kept them in case anyone extended this class.

tibi2303: thanks a lot @Aporie. Your contrib is very well received, i improve the tests and we are ready for another review from a collegue.

modules/oe_whitelabel_search/src/Form/SearchForm.php
```diff
@@ -101,7 +101,9 @@ public function buildForm(array $form, FormStateInterface $form_state, ?array $c
    */
   public function submitForm(array &$form, FormStateInterface $form_state): void {
     $config = $form_state->get('oe_whitelabel_search_config');
-    $url = Url::fromUserInput('/' . $config['form']['action'], [
+    // Makes sure to have a relative path,
+    // with a single prepended slash.
+    $url = Url::fromUserInput('/' . ltrim($config['form']['action'], '/'), [
```
donquixote, line 119: In the projects I can find, the search url config value is `search`, not `/search`.
What is interesting is that there is no validation, and no explanation in the form whether a leading slash is expected or not.

Fixing this here is _one_ possible solution, but it means that the unchecked config value is passed around in the block and in the form.

modules/oe_whitelabel_search/src/Form/SearchForm.php
```diff
@@ -101,7 +101,9 @@ public function buildForm(array $form, FormStateInterface $form_state, ?array $c
    */
   public function submitForm(array &$form, FormStateInterface $form_state): void {
     $config = $form_state->get('oe_whitelabel_search_config');
-    $url = Url::fromUserInput('/' . $config['form']['action'], [
+    // Makes sure to have a relative path,
+    // with a single prepended slash.
```
donquixote: Some thoughts on this comment:
- "Makes sure" -> "Make sure". The former would make sense as a function/method doc (meaning "This method makes sure that..."), but for inline comment the latter form is more common.
- Line break after comma feels unnecessary.
- "prepended slash" -> "leading slash" is the common term for this.
- It is quite obvious from the code that we want to have a single leading slash. What is less obvious is where this value is coming from, and which format we can expect.

```suggestion
    // The format of the search destination path setting is currently not
    // enforced. It may or may not have a leading slash.
```

donquixote: (going to push)

modules/oe_whitelabel_search/tests/src/Functional/SearchFormTest.php
```diff
@@ -85,4 +106,20 @@ public function testEmptySearchRemovesQueryParam(): void {
...
   }
 
+  /**
+   * Sets the default search block form action.
+   *
+   * @param string $action
+   *   The form action.
+   */
+  protected function setSearchBlockFormAction(string $action): void {
+    $block = $this->container->get('entity_type.manager')->getStorage('block')->load('oe_whitelabel_search_form');
```
donquixote, line 133: We don't need to be so verbose in a test, we don't care as much about dependency injection, mostly because this code won't be used anywhere else.
`Block::load()` is fine, and it works better for static analysis.
Alternatively, `$this->container->get(EntityTypeManagerInterface::class)->...` is another way to better support static analysis.

modules/oe_whitelabel_search/tests/src/Functional/SearchFormTest.php
```diff
@@ -47,6 +47,27 @@ public function testSearchQueryParams() {
...
+    $page->fillField('search_input', $search_text);
+    $page->pressButton('Search');
+
+    $current_url = $this->getSession()->getCurrentUrl();
+    $current_path = parse_url($current_url, PHP_URL_PATH);
+    $parsed_url = UrlHelper::parse($current_url);
+
+    $this->assertStringEndsWith('/search', $current_path);
+    $this->assertStringNotContainsString('//', $current_path);
+    $this->assertEquals(['text' => $search_text], $parsed_url['query']);
```
donquixote: We want to avoid uncertainty and flexibility in a test.
If we know the exact value of something, there is no need to make flexible assertions.

In this case, the only thing that we might not know fully is the base url of the website. It could change between environments (local docker vs CI).

Going to push a version that simplifies this.

modules/oe_whitelabel_search/tests/src/Functional/SearchFormTest.php
```diff
@@ -47,6 +47,27 @@ public function testSearchQueryParams() {
     $this->assertEquals($options['query'], $parsed_url['query']);
   }
 
+  /**
+   * Test the search form redirects correctly when action starts with slash.
+   */
+  public function testSearchActionWithLeadingSlash(): void {
+    $this->setSearchBlockFormAction('/search');
```
donquixote: If we have a dedicated test method for an action with leading slash, we should also cover actions without leading slash, or with more fragments.

(Going to push something)

AaronGilMartinez (approved):

PR #366, opened 2025-06-16
OEL-3909: [oe_whitelabel] Error on phpunit tests

enriquelacoma:


templates/overrides/navigation/oe-corporate-blocks-ec-footer.html.twig
```diff
@@ -44,7 +44,7 @@
         'items': [
         {
           'type': 'links',
-          'links': bcl_footer_links(corporate_footer.service_navigation),
+          'links': bcl_footer_links(corporate_footer.social_media_common_links),
```
AaronGilMartinez: I think we should require 5.0.0-beta5 as minimum, otherwise people could have an older version and this would break the template.

oe_whitelabel.theme
```diff
@@ -297,17 +297,3 @@ function oe_whitelabel_preprocess_oe_corporate_blocks_eu_footer(&$variables): vo
     }
   }
 }
-
-/**
- * Implements hook_preprocess_HOOK().
- */
-function oe_whitelabel_preprocess_oe_corporate_blocks_ec_footer(&$variables): void {
```
AaronGilMartinez, line 304: Since we are removing a code that unsets a link that is not there anymore, I'm thinking about updating tests/src/Kernel/FooterBlockTest.php
We have some checks to see that the link is not displayed.

AaronGilMartinez (changes requested):

templates/overrides/navigation/oe-corporate-blocks-ec-footer.html.twig
drishu: maybe this then also:
```
"conflict": {
        "openeuropa/oe_corporate_blocks": "<4.19"
    },
```

drishu: Add an if statement for the var in the template to make it compatible with both

drishu (changes requested):

templates/overrides/navigation/oe-corporate-blocks-ec-footer.html.twig
```diff
@@ -39,12 +39,23 @@
...
+        {
+          'classes': 'col-12 col-lg-4',
+          'items': [
+            {
+              'type': 'links',
+              'links': bcl_footer_links(corporate_footer.service_navigation),
+            }
+          ],
+        },
+      {% endif %}
```
drishu: isn't it if service_navigation else social_media_common_links ?

drishu (changes requested):

drishu (approved):

AaronGilMartinez (approved):

PR #367, opened 2025-06-19
OEL-3916: Paragraph oe_accordion was ignoring field display settings.

julien-:
The oe_accordion preprocessing was ignoring field display configuration
by using viewField() without display settings. This prevented site builders
from controlling label visibility and field formatters through the UI.

Now retrieves and applies the configured display settings for the current
view mode, ensuring proper field rendering respects user configuration.

### DIFF
--- a/modules/oe_whitelabel_paragraphs/tests/src/Functional/ParagraphsTest.php
+++ b/modules/oe_whitelabel_paragraphs/tests/src/Functional/ParagraphsTest.php
@@ -124,6 +124,21 @@ class ParagraphsTest extends BrowserTestBase {
    * Test Accordion paragraphs form.
    */
   public function testAccordionParagraph(): void {
+    // Test label display settings first.
+    $display = \Drupal::service('entity_display.repository')->getViewDisplay('paragraph', 'oe_accordion_item', 'default');
+
+    // Configure title field to show label above.
+    $title_component = $display->getComponent('field_oe_text') ?: [];
+    $title_component['label'] = 'above';
+    $display->setComponent('field_oe_text', $title_component);
+
+    // Configure content field to hide label.
+    $content_component = $display->getComponent('field_oe_text_long') ?: [];
+    $content_component['label'] = 'hidden';
+    $display->setComponent('field_oe_text_long', $content_component);
+
+    $display->save();
+
     $this->drupalGet('/node/add/paragraphs_test');
     $page = $this->getSession()->getPage();
     $page->pressButton('Add Accordion');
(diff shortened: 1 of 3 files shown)


AaronGilMartinez:
Hi @julien-  thanks for the contribution, change seems ok only thing I miss is a check in the tests to see that the configuration is applied. Test for this paragraph is in modules/oe_whitelabel_paragraphs/tests/src/Functional/ParagraphsTest.php.

modules/oe_whitelabel_paragraphs/tests/src/Functional/ParagraphsTest.php
```diff
@@ -140,6 +155,25 @@ public function testAccordionParagraph(): void {
...
+    $field_label = $field_definition->getLabel();
+    $this->assertSession()->pageTextContains($field_label);
+
+    // Test with hidden label setting.
+    $title_component['label'] = 'hidden';
+    $display->setComponent('field_oe_text', $title_component);
+    $display->save();
+
+    // Clear render cache and reload page.
+    \Drupal::service('cache.render')->invalidateAll();
```
AaronGilMartinez: Invalidation of cache usually means something is not ok, like a missing dependency. But I would recommend to fix the test first. I have triggered drone with your changes and is failing https://drone.fpfis.eu/openeuropa/oe_whitelabel/2570
If you need to check the test locally you can run that specific test by launching:

`docker compose exec web ./vendor/bin/phpunit --testdox modules/oe_whitelabel_paragraphs/tests/src/Functional/ParagraphsTest.php --filter=testAccordionParagraph
`

AaronGilMartinez (changes requested):

modules/oe_whitelabel_paragraphs/tests/src/Functional/ParagraphsTest.php
AaronGilMartinez: Unfortunately forks doesn't trigger the pipelines.

julien-: Code updated. I let you check @AaronGilMartinez

AaronGilMartinez (approved):

modules/oe_whitelabel_paragraphs/tests/src/Functional/ParagraphsTest.php
AaronGilMartinez: Changes looks good to me, thank you @julien-

PR #373, opened 2025-07-28
OEL-4014: Moved search to header top.

tibi2303:


templates/search/block--whitelabel-search-block.html.twig
```diff
@@ -9,5 +9,22 @@
...
+      .addClass([
+        'dropdown-toggle',
+        'd-xl-none',
+        'px-2',
+      ])
+  }) }}
+  <div class="dropdown-menu px-3 px-xl-0 d-xl-block">
+    {{ content }}
+  </div>
+</div>
```
donquixote: This works!
I wonder if we should introduce a pattern / component for this kind of dropdown, which could be used like so:
```twig
{{ pattern('dropdown', {
  icon: 'search',
  content: content,
  breakpoint: 'xl'
}) }}
```

But for now let's keep it in here.

templates/search/block--whitelabel-search-block.html.twig
donquixote, line 1: The directory of this file is strange.
We have other files in `templates/overrides/search/` but this one in `templates/search/`.
But ok to leave it here for legacy reasons.

donquixote, line 1: There is one funny glitch, but maybe that's ok:
If you resize for mobile, click the icon, then resize to desktop, the search element remains "shifted".
You can then click anywhere else to bring it back to where it's meant to be.

templates/overrides/search/form--header-top--oe-whitelabel-search-form.html.twig
```diff
@@ -0,0 +1,9 @@
+{#
+/**
+ * @file
+ * Search form template for the header top region.
+ */
+#}
+{% extends "form.html.twig" %}
+{% set attributes = attributes.addClass('d-flex', 'ms-xl-4') %}
```
donquixote: It seems odd to have that horizontal margin on the `<form>` element instead of the outer block div.

(Maybe you have a good reason, e.g. it would be harder to do it differently.)

templates/overrides/navigation/block--oe-authentication-login-block.html.twig
```diff
@@ -23,6 +23,7 @@
     icon: {
       name: 'person-fill',
       size: 'xs',
+      attributes: create_attribute().removeClass('me-2-5')
```
donquixote: This is odd. We remove a class from an empty attributes object?

I think the only way to remove this is in BCL itself. Or if we pass the icon + text as the link text to bcl-link.

templates/overrides/navigation/block--oe-authentication-login-block.html.twig
```diff
@@ -13,7 +13,7 @@
 
 {% set title = content['#title']|render %}
 {% set link_title %}
-  <span class="d-block d-lg-inline-block">{{ title }}</span>
+  <span class="d-none d-xl-inline-block">{{ title }}</span>
```
donquixote, line 20: Looking at this template, I wonder if we actually benefit from bcl-link icon handling, or if it would be better to just pass the icon as part of the link_title.

OR have a version of bcl-link that handles this "hide text on mobile".

But, ok to keep it as is for now to avoid scope creep.

templates/search/block--whitelabel-search-block.html.twig
donquixote: One way to avoid it is to have two separate search forms.
But I am not sure if we want this.

resources/sass/components/_forms.scss
```diff
@@ -66,3 +66,41 @@
...
+  .dropdown-toggle {
+    &:after {
+      display: none;
+    }
+  }
+}
+.language-switcher,
+.oe-authentication {
+  a {
+    color: var(--bs-body-color);
```
donquixote: It is odd that we have to set the color like this:

Normally with bootstrap 5:
- Navigation links have the `nav-link` class on the `<a>` tag.
- We would get `.nav-link {.. color: var(--bs-nav-link-color) ..}`.
- We could set the `--bs-nav-link-color` to whatever we want for this area.

Instead:
- The `nav-link` class is not on the `<a>` but on a `<div>` around the `<a>`.
- Instead of setting the `--bs-nav-link-color` variable only for that top area, we have a rule with `.bcl-header .nav-link { color: ... }` which makes it blue.
- Then we have another rule to color the nav links in the navigation menu.

And now we target these individual links, instead of a more generic solution.

This will cause funky side effects on website projects if these organize these elements differently.
The developers will add custom css in their website subtheme only to counteract this css from the base theme, leading to overall messy css.

It would be cleaner to work with utility classes, and to use the bootstrap css variables in a smart way.

resources/sass/components/_forms.scss
```diff
@@ -66,3 +66,41 @@
...
+}
+
+@media (max-width: 1200px) {
+  .language-switcher,
+  .oe-authentication {
+    svg {
+      margin: 0 auto !important;
+    }
+  }
+}
```
donquixote, line 86: And also here, the selectors will match these elements even in a website subtheme where the same elements should be displayed differently.
We can achieve the same result with utility classes.

donquixote: Btw there is one trick for these margins:

Problem:
- We have two child elements, icon and text, where either one of them could be missing or hidden.
- We want a space between the items, but not before or after the first/last item.

Solution:
- Put a positive margin (`me-n-5`) on both child elements (icon and text).
- Put a negative margin (`me-n2-5`) on the wrapper element that contains the children.
- Maybe you need to use `ms-*` instead of `me-*` to avoid horizontal scrolling, I forgot which one it is.

This way you don't need any conditional CSS to check whether either one of the elements is hidden or missing.

templates/overrides/navigation/block--oe-authentication-login-block.html.twig
donquixote: Maybe we _do_ want to change this to have more control over the utility classes on each part.\
See my above comments about using utility classes instead of too much CSS.

templates/search/block--whitelabel-search-block.html.twig
donquixote: Actually: Should this dropdown behavior be configurable for the block?
This would allow websites to put this block in other places where it does not need to have this drpdown behavior.

modules/oe_whitelabel_search/oe_whitelabel_search.post_update.php
```diff
@@ -45,3 +45,18 @@ function oe_whitelabel_search_post_update_00002(&$sandbox) {
...
+
+/**
+ * Move the Whitelabel Search Block to the 'header_top' region.
+ */
+function oe_whitelabel_search_post_update_00004(&$sandbox) {
+  $block = Block::load('oe_whitelabel_search_form');
+  if (!$block || $block->getTheme() !== 'oe_whitelabel') {
+    return 'No update needed.';
+  }
+  if ($block->getRegion() !== 'header_top') {
```
donquixote: I would say we should only move the block if it is currently in 'navigation_right', ~(and if it is enabled)~.
If it is in another region or disabled, that means the site builder does not want the default behavior.

> ~(and if it is enabled)~

If the block is simply disabled, I think we can still move it, but keep it disabled.

donquixote: Also we can swap the condition so that we have symmetry with the earlier condition.
```suggestion
  if ($block->getRegion() === 'header_top') {
    return 'Search block already in correct region.';
  }
  if ($block->getRegion() !== 'navigation_right') {
    return 'The search block was intentionally moved to a different region.';
  }
  $block->setRegion('header_top')->save();
  return "Search block moved to 'header_top'.";
}
```

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
```diff
@@ -94,27 +94,57 @@ public function testNavigationRightSearchBlockRendering(): void {
...
+    $this->assertCount(1, $wrapper);
+
+    // Toggle button.
+    $toggle = $wrapper->filter('button.oe-whitelabel-search-form');
+    $this->assertCount(1, $toggle);
+    $this->assertSame('block-oe-whitelabel-search-form', $toggle->attr('id'));
+    $this->assertStringContainsString('dropdown-toggle', $toggle->attr('class') ?? '');
+    $this->assertStringContainsString('btn', $toggle->attr('class') ?? '');
+    $this->assertStringContainsString('btn-ghost', $toggle->attr('class') ?? '');
+    $this->assertStringContainsString('btn-md', $toggle->attr('class') ?? '');
```
donquixote: That's a lot of assert for a single class attribute.
Maybe we should have an ->assertClass() where we assert a normalized class array? (class names in alphabetic order to avoid fail if they get swapped).

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
```diff
@@ -94,27 +94,57 @@ public function testNavigationRightSearchBlockRendering(): void {
...
+    $this->assertCount(1, $icon);
+
+    // Dropdown menu container.
+    $dropdown = $wrapper->filter('div.dropdown-menu');
+    $this->assertCount(1, $dropdown);
+    $this->assertStringContainsString('px-3', $dropdown->attr('class') ?? '');
+    $this->assertStringContainsString('d-xl-block', $dropdown->attr('class') ?? '');
+
+    // The form.
+    $form = $dropdown->filter('form#oe-whitelabel-search-form');
```
donquixote: Do we have more forms within that block?

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
donquixote: Also below we simply assert the full class. What is different here?

resources/sass/components/_forms.scss
```diff
@@ -66,3 +66,41 @@
...
+  .dropdown-menu[data-bs-popper] {
+    left: auto;
+    right: 0;
+  }
+  .dropdown-toggle {
+    &:after {
+      display: none;
+    }
+  }
+}
```
donquixote: This CSS is ok I would say, because it is namespaced with `.search-dropdown`.
Sites that don't want this can simply override the template and not add that class.
OR if we introduce a setting in the search block, they can disable that setting to not have the dropdown.

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
donquixote, line 1: We are asserting all the elements and attributes.
But we don't test the actual behavior.
We could add a `FunctionalJavascript` test in oe_whitelabel_search module.

donquixote (changes requested):

resources/sass/components/_forms.scss
donquixote: Btw one reason to have `.nav-link` on the `<a>` instead of a wrapping element is that the clickable area is bigger, if we have the padding on `<a>`. So an improved usability.

resources/sass/components/_forms.scss
donquixote: Actually we can just use `gap`, much better.
(the negative margin trick is from old times where gap was not available)

All we need is to put `class="d-flex gap-2-5"` on the link.

templates/overrides/navigation/block--oe-authentication-login-block.html.twig
donquixote, line 1: I now have a version of this template which looks like below.
This means it does not use the bcl-link or pattern-link anymore.
It works without additional css.
(The color on the `<a>` is still wrong but this should be fixed elsewhere).

We can do this now or later.
```twig

{% set extra_attributes = create_attribute() %}
{% set extra_attributes = extra_attributes.addClass([
  'text-center'
]) %}

{% set link_title %}
  {{ pattern('icon', {
    name: 'person-fill',
    size: 'xs',
  }) }}
  <span class="d-none d-xl-inline-block">{{ content['#title'] }}</span>
{% endset %}
{% set link_attributes = create_attribute(content['#attributes'] ?? [])
  .addClass(['d-flex', 'gap-2-5', 'nav-link'])
%}
<div class="oe-authentication d-flex">
  {{ link(link_title, content['#url'], link_attributes) }}
</div>
```

templates/search/block--whitelabel-search-block.html.twig
tibi2303: ok, i will resolve this without change

templates/search/block--whitelabel-search-block.html.twig
tibi2303: ok, i will resolve this without change

templates/search/block--whitelabel-search-block.html.twig
tibi2303: no, i would not do this, dont like duplications.. still though the full search bar appears on xl view so we wont have any problem, because even on table the maximum size is 1200px i think which is still not xl

templates/overrides/navigation/block--oe-authentication-login-block.html.twig
tibi2303: leftover, should have been deleted, thanks.

templates/overrides/navigation/block--oe-authentication-login-block.html.twig
tibi2303: we have the possibility in bcl to use icons inside links and buttons, so why not use it?

tibi2303: i see your point but we first use the functionality of the template, then we see if we need to adapt or not

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
tibi2303: you are right, i am changing.

resources/sass/components/_forms.scss
tibi2303: i changed the classes and behaviour

templates/overrides/search/form--header-top--oe-whitelabel-search-form.html.twig
tibi2303: indeed, thanks

modules/oe_whitelabel_search/oe_whitelabel_search.post_update.php
tibi2303: changed

resources/sass/components/_forms.scss
tibi2303: we are limited by the icon template in link... we are adding me-2-5 on the icon from bcl and i cant remove it.. that's why that leftover which i removed

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
tibi2303: changed to simply form

templates/overrides/navigation/block--oe-authentication-login-block.html.twig
tibi2303: let's do it later, i see your point and yes would be nice

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
```diff
@@ -94,27 +94,51 @@ public function testNavigationRightSearchBlockRendering(): void {
...
+
+    // Dropdown menu container.
+    $dropdown = $wrapper->filter('div.dropdown-menu');
+    $this->assertCount(1, $dropdown);
+    $this->assertStringContainsString('dropdown-menu px-3 px-xl-0 d-xl-block ms-xl-4', $dropdown->attr('class'));
+
+    // The form.
+    $form = $dropdown->filter('form');
     $this->assertCount(1, $form);
-    $this->assertSame('oe-whitelabel-search-form', $form->attr('id'));
```
donquixote: We used to assert the id, then you changed that to filter by id, now we don't assert or filter by id anymore. (maybe you decided that we don't care about the id?)

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
donquixote: We decided that I will do this..

templates/search/block--whitelabel-search-block.html.twig
donquixote: Ok, maybe enough if a sub-theme can simply override this template.
Adding a setting to the block is not good because we want to separate backend from theme layer.

templates/overrides/navigation/block--oe-authentication-login-block.html.twig
donquixote: I still don't like that mix of utility classes and custom css. Especially not in a theme that is shared by multiple projects which may add their customizations.

This will trigger messy code in sub-themes that want to customize it.

If we don't want to change BCL, we cannot use `gap`, because the icon still has the margin-right.
We can use `me-n2-5` on the link, but this can cause horizontal scrollbar in some cases.

We could append an empty `<span class="ms-n2-5 d-xl-none"></span>` to the link text to eat up the bad margin.
It feels ugly, but now all the customization is in the template, and a sub-theme won't need to override custom css.

If we can do the follow-up before the next release of oe_whitelabel I am ok..

donquixote: > we have the possibility in bcl to use icons inside links and buttons, so why not use it?

the main reason to use bcl templates is if we can have simpler code in oe_whitelabel, and have a single source of change. So we can change the BCL template and we see the effect without having to update anything in oe_whitelable.
This advantage goes away if we have to pass more stuff into these templates, and especially if we have to write "custom" (theme-level) css to override styling from BCL.

resources/sass/components/_forms.scss
donquixote: (one open thread is enough, so closing this one.)

templates/search/block--whitelabel-search-block.html.twig
donquixote: (question still open)

donquixote:
One more problem I find:
The search dropdown goex beyond the content width, because the closest parent with `position:relative` is a container.
We need 12px margin-right on the dropdown, or change the parenting situation.

resources/sass/components/_forms.scss
```diff
@@ -66,3 +66,27 @@
...
 }
 
+
+.search-dropdown {
+  display: flex;
+  position: static;
+  .dropdown-menu[data-bs-popper] {
+    left: auto;
+    right: calc(var(--bs-gutter-x)*.5);
+    margin-left: calc(var(--bs-gutter-x)*.5);
```
donquixote: Code style, I think we want space around operators.
I have not found an actual rule, but the example code has spaces around the operator:
https://www.drupal.org/docs/develop/standards/css/css-formatting-guidelines
And it's more readable.
```suggestion
    right: calc(var(--bs-gutter-x) * .5);
    margin-left: calc(var(--bs-gutter-x) * .5);
```

donquixote (changes requested):

donquixote:
> One more problem I find:
> The search dropdown goex beyond the content width, because the closest parent with position:relative is a container.
> We need 12px margin-right on the dropdown, or change the parenting situation.

This problem is now fixed.

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
donquixote: @AaronGilMartinez What do you think, do we need a FunctionalJavascript test for the search dropdown behavior?

modules/oe_whitelabel_search/oe_whitelabel_search.post_update.php
```diff
@@ -45,3 +45,27 @@ function oe_whitelabel_search_post_update_00002(&$sandbox) {
   }
   return 'No search block needed to be updated.';
 }
+
+/**
+ * Move the Whitelabel Search Block to the 'header_top' region.
+ */
+function oe_whitelabel_search_post_update_00004(&$sandbox) {
```
AaronGilMartinez:
```suggestion
function oe_whitelabel_search_post_update_00003(&$sandbox) {
```

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
donquixote: I added one.

.drone.yml
```diff
@@ -43,6 +43,7 @@ pipeline:
     volumes:
       - /cache/${DRONE_REPO_NAME}:/cache
     commands:
+      - mkdir assets
```
donquixote: This change won't be needed if we do OEL-4038 first.

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
```diff
@@ -89,6 +89,13 @@ public function testNavigationRightSearchBlockRendering(): void {
       ->get('entity_type.manager')
       ->getStorage('block');
     $entity = $block_entity_storage->load('oe_whitelabel_search_form');
+
+    // Set the block region to 'navigation_right'.
+    $settings = $entity->get('settings');
+    $settings['form']['region'] = 'navigation_right';
+    $entity->set('settings', $settings);
+    $entity->save();
+
```
AaronGilMartinez, line 98: My guess is that we keep compatibility with header_top and navigation_right versions, right?

templates/overrides/navigation/block--oe-authentication-login-block.html.twig
```diff
@@ -6,17 +6,21 @@
...
+  'nav-link',
+  'text-dark',
+  'h-100',
+  'gap-2-5',
 ]) %}
 
 {% set title = content['#title']|render %}
 {% set link_title %}
-  <span class="d-block d-lg-inline-block">{{ title }}</span>
+  <span class="d-none d-xl-inline-block">{{ title }}</span>
```
AaronGilMartinez, line 20: This is changing the breakpoint, is it ok? Didn't see where it should break in the designs.

AaronGilMartinez (changes requested):

templates/overrides/navigation/block--oe-authentication-login-block.html.twig
tibi2303: there is no design for this actually, but in order to have enough space, the xl breakpoint is the best approach

modules/oe_whitelabel_search/oe_whitelabel_search.post_update.php
tibi2303: done, thanks

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
donquixote: Yes that was the idea.
If we treat oe_whitelabel like a real package to be used by others, keeping that theming is the less disruptive thing to do.

donquixote: Maybe we will change our mind about this in the future :)

AaronGilMartinez: Clear

templates/overrides/navigation/block--oe-authentication-login-block.html.twig
AaronGilMartinez: Seems sensible for me.

donquixote (approved):

AaronGilMartinez (approved):

PR #377, opened 2025-08-18
OEL-3730: Changed id to main and removed anchor plus test.

tibi2303:


tests/src/FunctionalJavascript/SkipToMainContentTest.php
```diff
@@ -0,0 +1,50 @@
...
+    $assert = $this->assertSession();
+    $skip_link = $assert->elementExists('css', 'a[href="#main-content"]');
+    $this->getSession()->executeScript(<<<'JS'
+      (function () {
+        var a = document.querySelector('a[href="#main-content"]');
+        if (!a) return;
+        a.focus();
+        a.click();
+      }());
+    JS);
```
donquixote, line 40: This does not capture how a user would find and interact with this link.
The goal is keyboard navigation:
- visit the page - at that point the skip link is hidden.
- hit "Tab" on the keyboard - at this point the skip link becomes visible, and has focus.
  (at some point we might want to assert the text of the link..)
- hit the enter key to trigger the link - this will focus `<main>`.

tests/src/FunctionalJavascript/SkipToMainContentTest.php
```diff
@@ -0,0 +1,50 @@
...
+    $hash = $this->getSession()->evaluateScript('return window.location.hash;');
+    $this->assertSame('#main-content', $hash, 'Location hash updated to #main-content.');
+    $active_id = $this->getSession()->evaluateScript('return document.activeElement ? document.activeElement.id : "";');
+    $this->assertSame('main-content', $active_id, 'Focus moved to #main-content.');
+    $main = $assert->elementExists('css', 'main#main-content');
+    $this->assertSame('-1', $main->getAttribute('tabindex'), '<main> has tabindex="-1".');
+    $this->assertCount(1, $this->getSession()->getPage()->findAll('css', 'main#main-content'));
```
donquixote: This seems redundant with earlier assertions.
I guess we want to assert that the element with that id is really the `<main>` element?

templates/overrides/page/page.html.twig
```diff
@@ -115,8 +115,7 @@
       {% endif %}
     </header>
   {% endblock %}
-  <main aria-label="{{ "Main content"|t }}">
-    <a id="main-content" tabindex="-1"></a>
+  <main aria-label="{{ "Main content"|t }}" id="main-content" tabindex="-1">
```
donquixote, line 118: One observation I make:
When the `<main>` element becomes active/focused, it gets `:focus-visible` class and gets an outline.
That outline looks a bit ugly (browser native styling).
Maybe it's ok for now, given that we have no other design specs.
So I'd say let's keep it, but we might get some interesting feedback.

tests/src/FunctionalJavascript/SkipToMainContentTest.php
tibi2303: this is what     $active_id = $this->getSession()->evaluateScript('return document.activeElement ? document.activeElement.id : "";'); does... gets the current focused element which should be #main-content. We assert in the test that focused element id is main-content    $this->assertSame('main-content', $active_id);... I dont see what we can do more than this.. it's the expected behaviour

tibi2303: This is functional javascript test.. we test the behaviour of the skip link

tests/src/FunctionalJavascript/SkipToMainContentTest.php
tibi2303: indeed, it's deleted

tibi2303: we already test this on line 45

templates/overrides/page/page.html.twig
tibi2303: okay, i suggest to keep it until somebody comes with changes here

piotrsmykaj: The issue described by @donquixote is a direct result of how tabindex="-1" affects the `<main>` element focus. Normally, a `<main>` element is not focusable. When a user clicks "Skip to main content" link, the browser scrolls to the element but does not transfer focus to it.

Adding `tabindex="-1"` makes the `<main>` element focusable so when the skip link is clicked, the browser moves the keyboard focus from the link to the `<main>` element, using the :focus-visible state, which is why the blue outline appears.

In my opinion, this is a correct behaviour for ensuring accessibility.

tests/src/FunctionalJavascript/SkipToMainContentTest.php
piotrsmykaj: @donquixote This is a valid point but actually this test focuses on the outcome. The goal is to confirm the link correctly moves focus and this test achieves that effectively.

piotrsmykaj (dismissed):

piotrsmykaj (approved):

PR #381, opened 2025-09-22
OEL-4016: Mega menu

donquixote:


modules/oe_whitelabel_helper/oe_whitelabel_helper.post_update.php
```diff
@@ -41,3 +42,12 @@ function oe_whitelabel_helper_post_update_00003(): void {
...
   \Drupal::service('config.installer')->installOptionalConfig($storage);
 }
+
+/**
+ * Place the OEL mega menu block.
+ */
+function oe_whitelabel_helper_post_update_00004(): void {
+  ConfigImporter::importSingle('module', 'oe_whitelabel_helper', '/config/post_updates/00004_megamenu', 'block.block.oe_whitelabel_oelmegamenu');
+
+  Block::load('oe_whitelabel_main_navigation')->setStatus(FALSE)->save();
```
piotrsmykaj: to make it consistent, why not doing the same logic as in `oe_bootstrap_theme` with all those checks?
```
[...]
  $old_navigation_block = Block::load('oe_whitelabel_main_navigation');
  if (!$old_navigation_block || $old_navigation_block->getTheme() !== 'oe_whitelabel') {
   return $report . "\nThe old navigation block was not found.";
  }
  if (!$old_navigation_block->status()) {
    return $report . "\nThe old navigation block was already disabled.";
  }
[...]
```

templates/overrides/page/page.html.twig
```diff
@@ -70,39 +70,50 @@
...
-          right: page.header_top,
-          color_scheme: 'light',
-          background: false,
-          attributes: create_attribute({
-            'class': ['w-100', 'shadow-sm'],
-            'role': 'navigation',
-            'aria-label': 'Site settings navigation'|t
-          })
-        }) }}
+        <nav class="w-100 shadow-sm navbar navbar-expand-lg navbar-light" role="navigation" aria-label="{{ "Site settings navigation"|t }}">
```
piotrsmykaj, line 93: what do you think about a new "megamenu" variant for navbar pattern?

templates/overrides/page/page.html.twig
```diff
@@ -70,39 +70,50 @@
...
-          color_scheme: 'dark',
-          background: false,
-          expanded_breakpoint: 'lg',
-          attributes: create_attribute({
-            'class': ['bcl-header__navbar'],
-            'role': 'navigation',
-            'aria-label': 'Menu'|t
-          })
-        }) }}
+        <nav class="bcl-header__navbar navbar navbar-expand-lg navbar-dark" role="navigation" aria-label="{{ 'Menu'|t }}">
```
piotrsmykaj, line 108: same question as above

modules/oe_whitelabel_helper/oe_whitelabel_helper.post_update.php
donquixote: The reason is that I wrote the one for oe_bootstrap_theme later.
But I can fix it.

templates/overrides/page/page.html.twig
donquixote: It is a valid idea. But I think it is not a blocker, and it would be risky to do it now.

templates/overrides/page/page.html.twig
donquixote: It is a valid idea. But I think it is not a blocker, and it would be risky to do it now.

piotrsmykaj (dismissed):

tibi2303 (dismissed):

tibi2303 (approved):

PR #390, opened 2025-10-29
UCPKN-3850: Add oe_whitelabel_modal_page submodule.

kp77:


modules/oe_whitelabel_modal_page/oe_whitelabel_modal_page.module
```diff
@@ -0,0 +1,77 @@
...
+ * Implements hook_ENTITY_TYPE_update() for modal entities.
+ */
+function oe_whitelabel_modal_page_modal_update(EntityInterface $entity): void {
+  // Invalidate modal_page cache tags when a modal is updated.
+  // This fixes an issue in modal_page module where cache is only invalidated
+  // in the form submission handler, not on programmatic saves.
+  // @see \Drupal\modal_page\Form\ModalForm::save()
+  // @todo Report this issue to modal_page module.
+  Cache::invalidateTags(['modal_page:' . $entity->id()]);
+}
```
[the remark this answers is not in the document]
kp77: I opened an issue about this, and added a patch, so this temporary fix is now removed.

https://www.drupal.org/project/modal_page/issues/3555830

composer.json
donquixote, line 1: We could also enable the submodule in `runner.yml.dist`.

donquixote:
I noticed a crazy problem here. I reduced it to the minimum reproducible steps.

Steps:
```
docker compose exec web ./vendor/bin/drush si -y minimal
docker compose exec web ./vendor/bin/drush en -y oe_multilingual
docker compose exec web ./vendor/bin/drush en -y modal_page
```
I get this error, triggered in `modal_page_install()`.

> Route "modal_page.settings" does not exist.

The reason is that `oe_multilingual` module somehow causes the route list to be populated, before the routes from `modal_page.routing.yml` are registered, and then `modal_page_install()` uses the outdated route list where the routes from `modal_page` are missing.

More precisely, `Drupal\oe_multilingual\Plugin\LanguageNegotiation\LanguageNegotiationAdmin::create()` (defined in parent class) gets the 'router' service, and that seems enough to warm up the route list.

composer.json
kp77: Sure, I can add it there too.

kp77:
> I noticed a crazy problem here. I reduced it to the minimum reproducible steps.
> ...

Hi @donquixote,

Thanks for checking the PR! It's strange, I could not reproduce it (not with the exact steps you provided, or otherwise).

But I know about this issue, and actually reported it to modal_page:
https://www.drupal.org/project/modal_page/issues/3543767

And I'm [using the patch](https://github.com/openeuropa/oe_whitelabel/pull/390/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R79) I created to fix it in this PR.

I wonder why you have it and I can't reproduce.. I tested with core version is 11.2.8. Can you check if that patch is applied to modal_page?

Note: I'm planning to rebase this branch, but for now I keep it like this until we figure this issue out.

donquixote:
> I wonder why you have it and I can't reproduce.. I tested with core version is 11.2.8. Can you check if that patch is applied to modal_page?

It happened because I merged a version of 1.x where we had another "patches" section, so the patches were not applied.
With the latest version of 1.x, this no longer happens.
Sorry for the noise.

composer.json
```diff
@@ -73,6 +74,16 @@
         }
     },
     "extra": {
+        "patches": {
+            "drupal/modal_page": {
+                "RouteNotFoundException thrown during module install: https://www.drupal.org/project/modal_page/issues/3543767": "https://git.drupalcode.org/project/modal_page/-/merge_requests/55.diff",
+                "Support for less-intrusive dialogs via Bootstrap Toasts: https://www.drupal.org/project/modal_page/issues/3535144": "https://git.drupalcode.org/project/modal_page/-/merge_requests/53.diff",
```
donquixote: We need stable patch urls.\
The content from an MR patch url will change whenever somebody pushes to the branch, including someone malicious.
A typical solution is to upload the patch as a file to the drupal.org issue, then link to that.

kp77: Yes, I'm going to replace these with static patch files. I just kept them like this temporarily because I was waiting for them to be released so there won't be need for them at all. These issues are RTBC, but not sure when the maintainer is going to release them. (sorry, just one of them is RTBC)

modules/oe_whitelabel_modal_page/oe_whitelabel_modal_page.info.yml
```diff
@@ -0,0 +1,7 @@
+name: OpenEuropa Whitelabel Modal
+type: module
+description: Modal component.
```
donquixote: What about
```suggestion
description: Integrates the Modal Page module with OpenEuropa Whitelabel theme.
```
(not sure if there is a convention for these descriptions)

modules/oe_whitelabel_modal_page/oe_whitelabel_modal_page.module
donquixote, line 1: Procedural hook implementations in this file should be moved to a class when we drop Drupal 10 support.
Until then we _could_ do the compatibility trick with `#[LegacyHook]` but honestly I don't mind to keep the procedural-only hooks until then.

modules/oe_whitelabel_modal_page/oe_whitelabel_modal_page.module
```diff
@@ -0,0 +1,51 @@
...
+    $base_themes = [];
+    foreach ($active_theme->getBaseThemeExtensions() as $base_theme) {
+      $base_themes[] = $base_theme->getName();
+    }
+    if (!in_array('oe_whitelabel', $base_themes, TRUE)) {
+      // Do nothing if the active theme or one of the base themes of the active
+      // theme is not oe_whitelabel.
+      return;
+    }
+  }
```
donquixote: There may be ways to optimize or simplify this further, but it's a pointless exercise, so let's keep it as is :)

donquixote: For comments in/on conditional blocks, my own rule of thumb is like this:
- Inside an `if` condition, I use comments with non-conditional phrasing.
- Above an `if` condition, I use comments with questions or conditional phrasing.
- Most of the time, it is enough to have one of them.
- By having a comment above the entire block, the reader will know to skip the block entirely, without having to dive into the conditions.

```
// Is it raining?
// (or)
// Check if it is raining.
// (or)
// If it is raining, we need to open the umbrella.
if (is_it_raining()) {
  // It is raining. Open the umbrella.
  open_umbrella();
}
else {
  // It is not raining. Wear sunglasses.
  wear_sunglasses();
}
```
So for the above code block:
```suggestion
  // Do nothing if the active theme or one of the base themes of the active
  // theme is not oe_whitelabel.
  $active_theme = \Drupal::theme()->getActiveTheme();
  if ($active_theme->getName() !== 'oe_whitelabel') {
    $base_themes = [];
    foreach ($active_theme->getBaseThemeExtensions() as $base_theme) {
      $base_themes[] = $base_theme->getName();
    }
    if (!in_array('oe_whitelabel', $base_themes, TRUE)) {
      return;
    }
  }
```

modules/oe_whitelabel_modal_page/oe_whitelabel_modal_page.module
```diff
@@ -0,0 +1,51 @@
...
+  if ($modal_entity_type instanceof ConfigEntityType) {
+    // Override modal_page module's form for editing modals.
+    $modal_entity_type->setFormClass('add', ModalForm::class);
+    $modal_entity_type->setFormClass('edit', ModalForm::class);
+  }
+}
+
+/**
+ * Implements hook_page_attachments_alter().
+ */
```
donquixote:
```suggestion
/**
 * Implements hook_page_attachments_alter().
 *
 * Removes the bootstrap library from modal_page module, if oe_whitelabel theme is active.
 */
```
(improve and insert line breaks as applicable)

modules/oe_whitelabel_modal_page/oe_whitelabel_modal_page.module
```diff
@@ -0,0 +1,51 @@
...
+
+declare(strict_types=1);
+
+use Drupal\Core\Config\Entity\ConfigEntityType;
+use Drupal\oe_whitelabel_modal_page\Form\ModalForm;
+
+/**
+ * Implements hook_entity_type_alter().
+ */
```
donquixote, line 17:
```suggestion
/**
 * Implements hook_entity_type_alter().
 *
 * Sets custom form classes for 'modal' entity type.
 */
```

modules/oe_whitelabel_modal_page/src/Form/ModalForm.php
```diff
@@ -0,0 +1,44 @@
...
+
+declare(strict_types=1);
+
+namespace Drupal\oe_whitelabel_modal_page\Form;
+
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\modal_page\Form\ModalForm as OriginalModalForm;
+
+/**
+ * Override of modal_page configuration form.
```
donquixote: The name of the entity type is 'modal', not 'modal_page', I think that is the more relevant piece of info here.

modules/oe_whitelabel_modal_page/oe_whitelabel_modal_page.module
```diff
@@ -0,0 +1,51 @@
...
+  $libraries = $attachments['#attached']['library'] ?? NULL;
+  if ($libraries) {
+    foreach ($libraries as $delta => $library) {
+      if (str_starts_with($library, 'modal_page/modal-page-bootstrap')) {
+        // Make sure that modal_page doesn't load any CDN Bootstrap libraries.
+        unset($attachments['#attached']['library'][$delta]);
+      }
+    }
+  }
+}
```
donquixote: This hook could be renamed `oe_whitelabel_page_attachments_alter()` and moved to `oe_whitelabel.theme`.
This way we don't need the first part where we check the active theme.
A drawback would be that this will fire even if the module is not enabled, but I think that is ok in this case, it is a cheap operation. (TBD, let me know if you disagree)

For the libraries part we _can_ use `preg_grep()` with `PREG_GREP_INVERT`.

modules/oe_whitelabel_modal_page/src/Form/ModalForm.php
donquixote, line 1: All of this form and permission aspect could be proposed as an MR for `modal_page` module, OR live as a separate contrib module.
(I think this was discussed earlier, but I don't remember the arguments and conclusion)

resources/patches/oe_bootstrap_theme_rename_bcl_toasts_dir.patch
```diff
@@ -0,0 +1,4 @@
+diff --git a/assets/bcl/bcl-toast/bcl-toasts.html.twig b/assets/bcl/bcl-toasts/bcl-toasts.html.twig
+similarity index 100%
+rename from assets/bcl/bcl-toast/bcl-toasts.html.twig
+rename to assets/bcl/bcl-toasts/bcl-toasts.html.twig
```
donquixote: I would prefer to see this fixed in `oe_bootstrap_theme`, so we don't have to merge with this patch.
Also I wonder about side effects for other things that use `toasts`.

tests/src/Functional/ModalPageTemplateOverrideTest.php
donquixote, line 1: This test is about templates in `oe_whitelabel`, which I think are active whenever `modal_page` module is installed, independent of `oe_whitelabel_modal_page` module.
So we should move the test to the top level.

tests/src/Functional/ModalPageBootstrapLibraryTest.php
donquixote, line 1: If we move the hook to `oe_whitelabel.theme`, then we should also move the test to `oe_whitelabel/tests/src`.

donquixote (changes requested):
@kp77 First round of feedback, before I have a deeper look at the actual templates and scss.

Main points:
We could move the library stuff to `oe_whitelabel.theme`.
What remains in the module is the permission management, and that seems to be unrelated to `oe_whitelabel`.
In the end we might have a new mini contrib module _or_ a patch to `modal_page`, and the rest will be in `oe_whitelabel` theme without any submodule.

donquixote:
> What remains in the module is the permission management, and that seems to be unrelated to oe_whitelabel.

We discussed this.
The reason why we want to have it as a submodule of oe_whitelabel is that we are specifically hiding form fields that don't make sense in combination with our templates for modal and toast.

So, based on this, I support keeping it here.
(but would still move the other hook to oe_whitelabel)

modules/oe_whitelabel_modal_page/src/Form/ModalForm.php
```diff
@@ -0,0 +1,44 @@
...
+      $form['modal_footer']['enable']['dont_show_again']['close_modal_cookie']['#access'] = FALSE;
+      $form['modal_buttons']['button_close']['top_right_button_class']['#access'] = FALSE;
+      $form['modal_buttons']['button_close']['top_right_button_label']['#access'] = FALSE;
+      $form['modal_buttons']['ok_button']['ok_button_class']['#access'] = FALSE;
+      $form['modal_buttons']['left_button']['left_button_class']['#access'] = FALSE;
+      $form['modal_buttons']['maximize_button']['#access'] = FALSE;
+      $form['modal_class']['#access'] = FALSE;
+      $form['parameters']['#access'] = FALSE;
+      $form['extras']['type']['#access'] = FALSE;
+    }
```
donquixote, line 39: One thing I would point out:
If an admin creates a modal and uses some non-default values for the advanced settings, then a webmaster editing the same modal will not see that this one is "special".

I think we can assume that this scenario does generally not happen in the projects we care about. But it can be relevant if we would publish this functionality to a wider audience, e.g. as a contribution to the module.

resources/patches/oe_bootstrap_theme_rename_bcl_toasts_dir.patch
donquixote: If we want to safely apply this as a patch, it should be a symlink, so that we support both `toast` and `toasts`.

templates/modal_page/modal-page-modal.html.twig
```diff
@@ -0,0 +1,124 @@
...
+ */
+#}
+{% set _attributes = create_attribute({
+  'class': ['js-modal-page-show']|merge(modal_class|split(' ')),
+  'data-modal-options': modal_options
+}) %}
+
+{% if enable_modal_header %}
+  {% set _header %}
+    {% apply spaceless %}
```
donquixote: The `spaceless` filter is deprecated as of Twig 3.12.
https://twig.symfony.com/doc/3.x/filters/spaceless.html

The alternative is this: https://twig.symfony.com/doc/3.x/templates.html#templates-whitespace-control

```twig
<h5 class="mb-0">
  {{- title -}}
</h5>
```

resources/patches/oe_bootstrap_theme_rename_bcl_toasts_dir.patch
kp77: I think there is no need for that, because when the directory name is 'toast' the component cannot be included, because this gives an error without the patch:

```twig
{% include '@oe-bcl/toasts' %}
```

and this also:
```twig
{% include '@oe-bcl/toast' %}
```
Reason: the component's name is 'toasts' not 'toast'.  However, the package name in BCL used 'toast' incorrectly and that's why this directory name became 'toast'.

This patch is only fixing the issue that currently it is not possible to include the toasts component.

The proper fix in BCL will also not support 'toast' (as it wasn't something that ever worked):
https://github.com/openeuropa/bootstrap-component-library/pull/705/files

donquixote:
Interesting thing I notice:
A youtube video keeps playing when I close the modal.
But this seems to be independent of oe_whitelabel theme, so should be reported and fixed in the modal_page module itself. I won't create the issue for now because maybe it is just me..

modules/oe_whitelabel_modal_page/tests/src/Functional/ModalFormFieldAccessTest.php
```diff
@@ -0,0 +1,62 @@
...
+    $assert_session->fieldNotExists('Class(es)', $assert_session->elementExists('xpath', '//details[./summary[.="MODAL FOOTER"]]'));
+    $assert_session->elementNotExists('xpath', '//details[./summary[.="Cookies"]]');
+    $assert_session->fieldNotExists('Label', $assert_session->elementExists('xpath', '//details[./summary[.="Button X close"]]'));
+    $assert_session->fieldNotExists('ok_button_class');
+    $assert_session->fieldNotExists('left_button_class');
+    $assert_session->elementNotExists('xpath', '//details[./summary[.="Maximize Button"]]');
+    $assert_session->elementNotExists('xpath', '//details[./summary[.="MODAL CLASS"]]');
+    $assert_session->fieldNotExists('Parameters');
+    $assert_session->fieldNotExists('Modal By');
+  }
```
donquixote, line 95: Maybe we should also test the opposite, so that we are sure the fields are hidden exactly when the user does not have that perm.
Could it be done with the same block of code with a boolean variable?
```php
$field_exists_if = $should_exist 
      ? $assert_session->fieldExists(...)
      : $assert_session->fieldNotExists(...);
$field_exists_if('Open this modal clicking on this element');
```
Not sure if that's a good idea :)

modules/oe_whitelabel_modal_page/tests/src/Functional/ModalPageBootstrapLibraryTest.php
```diff
@@ -0,0 +1,120 @@
...
+
+    // Assert that Bootstrap libraries are NOT loaded even with a sub-theme.
+    $this->assertStringNotContainsString('bootstrap.min.js', $html);
+    $this->assertStringNotContainsString('bootstrap.min.css', $html);
+  }
+
+  /**
+   * Tests that Bootstrap libraries are loaded when using a different theme.
+   */
+  public function testDefaultBootstrapLibraryLoad(): void {
```
donquixote: We could do all of these in a single test method, if this speeds up the overall test.
(It is cleaner and nicer in separate methods, but we need to make compromises)

modules/oe_whitelabel_modal_page/tests/src/Functional/ModalPageTemplateOverrideTest.php
```diff
@@ -0,0 +1,469 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function tearDown(): void {
+    /** @var \Drupal\modal_page\Entity\ModalInterface $modal */
+    $modal = \Drupal::entityTypeManager()->getStorage('modal')->load('test_modal');
+    if ($modal) {
+      $modal->delete();
```
donquixote: Why do we need this?
I can see why we would need it in an ExistingSite test, but here?

kp77:
> Interesting thing I notice: A youtube video keeps playing when I close the modal. But this seems to be independent of oe_whitelabel theme, so should be reported and fixed in the modal_page module itself. I won't create the issue for now because maybe it is just me..

Yes, I could reproduce this too (outside oe_whitelabel). I reported the issue: https://www.drupal.org/project/modal_page/issues/3559547

templates/modal_page/modal-page-modal.html.twig
kp77: I was trying to find where this came from, and I think I brought it from OE Bootstrap theme modal pattern: https://github.com/openeuropa/oe_bootstrap_theme/blob/1.x/templates/patterns/modal/pattern-modal.html.twig#L10

I will remove these, they are not necessary (the BCL component applies spaceless anyway on the whole template).

The `{{- title -}}` whitespace control is not really an alternative to this IMO, because it's not doing the same thing. But there is no whitespace in the title we need to remove.

Also, for me it's ok to even have the whitespace (and not remove it), I just thought it's OE practice to remove the whitespace between tags.

composer.json
kp77: Enabled in [a237d66](https://github.com/openeuropa/oe_whitelabel/pull/390/commits/a237d66f29c782d868c79147db11510fe28eb67c)

composer.json
kp77: Done in [854b77f](https://github.com/openeuropa/oe_whitelabel/pull/390/commits/854b77f440d7936cf9863d42e1209a8a103e0c36)

modules/oe_whitelabel_modal_page/oe_whitelabel_modal_page.info.yml
kp77: Done in [9eafc3f](https://github.com/openeuropa/oe_whitelabel/pull/390/commits/9eafc3f353781dd2efd6f2f3ea79308e3732950b)

modules/oe_whitelabel_modal_page/oe_whitelabel_modal_page.module
kp77: I kept the procedural.

modules/oe_whitelabel_modal_page/oe_whitelabel_modal_page.module
kp77: This condition was removed at the end.

modules/oe_whitelabel_modal_page/oe_whitelabel_modal_page.module
kp77: Done in [9eafc3f](https://github.com/openeuropa/oe_whitelabel/pull/390/commits/9eafc3f353781dd2efd6f2f3ea79308e3732950b)

modules/oe_whitelabel_modal_page/oe_whitelabel_modal_page.module
kp77: Done in [9eafc3f](https://github.com/openeuropa/oe_whitelabel/pull/390/commits/9eafc3f353781dd2efd6f2f3ea79308e3732950b)

modules/oe_whitelabel_modal_page/src/Form/ModalForm.php
kp77: Done in [9eafc3f](https://github.com/openeuropa/oe_whitelabel/pull/390/commits/9eafc3f353781dd2efd6f2f3ea79308e3732950b)

modules/oe_whitelabel_modal_page/oe_whitelabel_modal_page.module
kp77: Done in [02ec1c3](https://github.com/openeuropa/oe_whitelabel/pull/390/commits/02ec1c3b37bc08cb56cb424c5da1b1f60e5331fb)

tests/src/Functional/ModalPageTemplateOverrideTest.php
kp77: Done in [1cfd036](https://github.com/openeuropa/oe_whitelabel/pull/390/commits/1cfd036a09d0aad044545e04a3be1951546aebf0)

tests/src/Functional/ModalPageBootstrapLibraryTest.php
kp77: Done in [02ec1c3](https://github.com/openeuropa/oe_whitelabel/pull/390/commits/02ec1c3b37bc08cb56cb424c5da1b1f60e5331fb)

modules/oe_whitelabel_modal_page/src/Form/ModalForm.php
kp77: This is true, but not sure if it's a problem. Depends on the use case of the site what access they want to give to webmasters.

templates/modal_page/modal-page-modal.html.twig
kp77: Done in [061c1da](https://github.com/openeuropa/oe_whitelabel/pull/390/commits/061c1dae05adc01cbc1ce3c3e94a4f1f5e1a902d)

modules/oe_whitelabel_modal_page/tests/src/Functional/ModalFormFieldAccessTest.php
kp77: Done in [14b512a](https://github.com/openeuropa/oe_whitelabel/pull/390/commits/14b512ad61f2635aa5b6516d9a824fb2d6ed1112)

modules/oe_whitelabel_modal_page/tests/src/Functional/ModalPageBootstrapLibraryTest.php
kp77: Done in [e446a7b](https://github.com/openeuropa/oe_whitelabel/pull/390/commits/e446a7bb16c6dbac24e12f3d2fafa68b47ebd81f)

modules/oe_whitelabel_modal_page/tests/src/Functional/ModalPageTemplateOverrideTest.php
kp77: Removed the method in [131eb1c](https://github.com/openeuropa/oe_whitelabel/pull/390/commits/131eb1cd931381a01165fd495c3e28f5873e90b1)

modules/oe_whitelabel_modal_page/tests/src/Functional/ModalFormFieldAccessTest.php
```diff
@@ -28,13 +28,39 @@ class ModalFormFieldAccessTest extends BrowserTestBase {
...
+   * Tests that advanced fields are visible with permission.
+   */
+  public function testAdvancedFieldsVisibleWithPermission(): void {
+    $this->loginWithAdvancedPermission(TRUE);
+    $this->assertAdvancedFieldsVisibility(TRUE);
+  }
+
+  /**
+   * Logs in a modal user with or without the advanced permission.
+   */
```
donquixote, line 48: Do we want to add `@param` doc? I am honestly undecided.
Maybe the var name is clear enough. And if code style check is happy then feel free to click "resolve".

modules/oe_whitelabel_modal_page/tests/src/Functional/ModalFormFieldAccessTest.php
```diff
@@ -43,20 +69,23 @@ public function testAdvancedFieldsHiddenWithoutPermission(): void {
...
-    $assert_session->elementNotExists('xpath', '//details[./summary[.="Cookies"]]');
-    $assert_session->fieldNotExists('Label', $assert_session->elementExists('xpath', '//details[./summary[.="Button X close"]]'));
-    $assert_session->fieldNotExists('ok_button_class');
-    $assert_session->fieldNotExists('left_button_class');
-    $assert_session->elementNotExists('xpath', '//details[./summary[.="Maximize Button"]]');
-    $assert_session->elementNotExists('xpath', '//details[./summary[.="MODAL CLASS"]]');
-    $assert_session->fieldNotExists('Parameters');
-    $assert_session->fieldNotExists('Modal By');
+    $field_assert = $should_exist ? 'fieldExists' : 'fieldNotExists';
+    $element_assert = $should_exist ? 'elementExists' : 'elementNotExists';
```
donquixote, line 79: Alternative would be anonymous / arrow functions.
```suggestion
    $field_assert = $should_exist ? $assert_session->fieldExists(...) : $assert_session->fieldNotExists(...);
    $element_assert = $should_exist ? $assert_session->elementExists(...) : $assert_session->ementNotExists(...);
```
and
```php
    $field_assert('Open this modal clicking on this element');
```
(actually the var names would be different, e.g. `$assert_field_if`.)

The question is whether this would make the code below more readable.
Clearly it would be shorter, but the current version makes it clear that these are assert methods.

So ok for me to leave it as is.

modules/oe_whitelabel_modal_page/src/Form/ModalForm.php
donquixote: We decided to keep it here because we hide specifically those form fields that should not be used with our templates.

modules/oe_whitelabel_modal_page/README.md
```diff
@@ -0,0 +1,26 @@
+# OpenEuropa Whitelabel Modal Page
+
+This module provides integration with the [Modal Page](https://www.drupal.org/project/modal_page) contrib module, enabling modal dialogs and toast notifications styled with the Bootstrap Component Library (BCL).
+
+## Features
+
+- **Bootstrap Component Library integration**: Template overrides that render modals and toasts using BCL components
+- **Permission-based form simplification**: Hides advanced modal configuration fields from users without the "administer advanced modal page configuration" permission
+- **Library conflict prevention**: Automatically prevents modal_page from loading its own Bootstrap libraries to avoid conflicts with BCL
```
donquixote: It should be mentioned that two of these points are now implemented in the theme itself, independent of this module.

(I don't know if the top-level README.md has place for this.)

modules/oe_whitelabel_modal_page/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+To use this module, you need to require the Modal Page contrib module in your project's composer.json:
+
+```
+composer require drupal/modal_page:^6.0@beta
+```
+
+## Template overrides
+
+This module provides the following template overrides for the oe_whitelabel theme:
```
donquixote: Does it? "This module"?

resources/patches/oe_bootstrap_theme_rename_bcl_toasts_dir.patch
donquixote: nice, alright!

oe_whitelabel.theme
```diff
@@ -314,3 +314,20 @@ function oe_whitelabel_preprocess_oe_corporate_blocks_eu_footer(&$variables): vo
...
+ * Removes the bootstrap library from modal_page module.
+ */
+function oe_whitelabel_page_attachments_alter(array &$attachments): void {
+  $libraries = $attachments['#attached']['library'] ?? NULL;
+  if ($libraries) {
+    // Remove any modal_page Bootstrap libraries.
+    $attachments['#attached']['library'] = preg_grep(
+      '/^modal_page\/modal-page-bootstrap/',
+      $libraries,
+      PREG_GREP_INVERT
```
donquixote:
```suggestion
      PREG_GREP_INVERT,
```
It seems it is not enforced by code style, not sure why.

oe_whitelabel.theme
```diff
@@ -314,3 +314,20 @@ function oe_whitelabel_preprocess_oe_corporate_blocks_eu_footer(&$variables): vo
...
+ * Implements hook_page_attachments_alter().
+ *
+ * Removes the bootstrap library from modal_page module.
+ */
+function oe_whitelabel_page_attachments_alter(array &$attachments): void {
+  $libraries = $attachments['#attached']['library'] ?? NULL;
+  if ($libraries) {
+    // Remove any modal_page Bootstrap libraries.
+    $attachments['#attached']['library'] = preg_grep(
+      '/^modal_page\/modal-page-bootstrap/',
```
donquixote: Could be
```suggestion
      '/^modal_page\/modal-page-bootstrap(-\d+)?$/',
```
or
```suggestion
      '#^modal_page/modal-page-bootstrap(-\d+)?$#',
```

But it is good enough as it is :)
(click resolve)

tests/src/Functional/ModalPageBootstrapLibraryTest.php
```diff
@@ -0,0 +1,107 @@
...
+ * Tests that Bootstrap libraries are removed when oe_whitelabel is active.
+ */
+class ModalPageBootstrapLibraryTest extends BrowserTestBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected static $modules = [
+    'oe_whitelabel_helper',
+    'oe_whitelabel_modal_page',
```
donquixote: We don't need to enable the module for this test, do we?
```suggestion
    'modal_page',
    'oe_whitelabel_helper',
```

oe_whitelabel.info.yml
```diff
@@ -13,6 +13,10 @@ libraries-override:
     css:
       theme:
         css/system.admin.css: false
+  modal_page/modal-page:
+    css:
+      theme:
+        css/modal-page.css: false
```
donquixote, line 19: So, we could use the same mechanism to nuke `modal_page/modal-page-bootstrap-3` and `modal_page/modal-page-bootstrap-5`, but we are afraid that these will be renamed? (Ok for me)

donquixote, line 19: The goal is to remove the css but keep the js and the dependencies?
(ok for me)

resources/sass/default.style.scss
```diff
@@ -7,3 +7,4 @@
 @import "components/patterns";
 @import "components/blocks";
 @import "components/agenda";
+@import "components/modal_page";
```
donquixote, line 10: So, this scss will always be included, even when modal_page is not installed or no modal on the current page?

An alternative would be to generate a separate asset for this.
But, given that scss is so small, maybe not worth it.

oe_whitelabel.info.yml
donquixote, line 19: The `modal_page/modal-page-dialog-api` is ok and does not need removing?
(just asking, no clear opinion yet)

templates/modal_page/modal-page-modal.html.twig
```diff
@@ -0,0 +1,118 @@
...
+      </div>
+    </figure>
+  {% endif %}
+
+  {% if enable_do_not_show_again %}
+    <div>
+      <label class="modal-dont-show-again-label"><input type="checkbox" class="modal-page-please-do-not-show-again" value="{{ id }}"> {{ do_not_show_again_label }}</label>
+      <input type="hidden" id="cookie_expiration" name="cookie_expiration" value="{{ cookie_expiration_time }}" />
+    </div>
+  {% endif %}
```
donquixote, line 55: In the admin theme, this checkbox appears in the footer.
In oe_whitelabel it appears in the body.
Is this intended?

resources/sass/components/_modal_page.scss
```diff
@@ -0,0 +1,32 @@
...
+  .toast {
+    --bs-toast-bg: rgba(var(--bs-body-bg-rgb), 1);
+  }
+
+  .toast-header {
+    justify-content: space-between;
+  }
+
+  .toast-header .btn-close {
+    margin-left: auto;
```
donquixote, line 16: This is not needed with the `justify-content: space-between` on the parent element.
(unless we get more elements)

resources/sass/components/_modal_page.scss
```diff
@@ -0,0 +1,32 @@
...
+    flex-wrap: wrap;
+    align-items: center;
+    justify-content: flex-end;
+    background-color: var(--bs-toast-header-bg);
+    background-clip: padding-box;
+    border-bottom-left-radius: calc(var(--bs-toast-border-radius) - var(--bs-toast-border-width));
+    border-bottom-right-radius: calc(var(--bs-toast-border-radius) - var(--bs-toast-border-width));
+    border-top: var(--bs-toast-border-width) solid var(--bs-toast-header-border-color);
+    padding: var(--bs-toast-padding-y) var(--bs-toast-padding-x);
+  }
```
donquixote, line 39: Are all these styles needed?
Are they copied from somewhere or custom?
E.g. `flex-shrink` only works if the parent element has `display: flex`.

donquixote:
I observe another very curious thing.
(unrelated to this PR)
When I create a new modal, if the machine name is `example_modal`, it does not show up.
When the machine name is anything else, it works just fine.

donquixote (commented):
I have some feedback and questions, but overall I like it.

Before we merge:
- I would like Tiberiu to have a look at the templates and CSS.
- It would be cool if we can release the BCL toast typo fix first, which is already merged in BCL, just not released yet.

templates/modal_page/modal-page-modal.html.twig
kp77: Yes, this is one of the reasons of the override, our mockup places this to the bottom of body, but modal_page's template puts it in the footer.

resources/sass/default.style.scss
kp77: I used this solution because as I see this is the practice with other components included here  (not all components are necessarily used on the page that are included here). So we could put it to a separate asset (and with js I'd definitely do that), but with a small css override, I think this follows current pattern used in oe_whitelabel.

kp77:
> I observe another very curious thing. (unrelated to this PR) When I create a new modal, if the machine name is `example_modal`, it does not show up. When the machine name is anything else, it works just fine.

My guess is that you previously created another 'example_modal' and you maybe tested the 'Do not show again' feature that placed a cookie for that ID. Can you check the cookies if you have it maybe? (Otherwise there should be nothing special about the 'example_modal' ID per se, for me it works)

This is something I'm planning to report on modal_page btw, because I think it's not ideal to use these automatically generated machine name IDs for identifying the modals in cookies (exactly because of this reason). Probably it would be better to use uuid for that purpose, but this needs some analysis and it's not a trivial fix, so for now I postponed this issue.

oe_whitelabel.theme
kp77: These are function parameters not elements of an array, I don't think we need a comma here.

oe_whitelabel.info.yml
kp77: Yes, we could just remove the assets here. The idea was to contain this logic in the sub-module and that was the only reason I didn't do it here to be honest. But now that logic was moved to the theme anyway, so you're right, we might as well remove them the 'regular way' (and in this case I think the test class could be also removed, as this is not something that specifically needs to be tested IMO).

The only advantage of the current implementation is indeed - as you wrote - that it will work with any potential future version numbers of bootstrap and probably won't need special attention in the future.

tests/src/Functional/ModalPageBootstrapLibraryTest.php
kp77: True, now it's enough to enable modal_page. Fixed it in [360baf6](https://github.com/openeuropa/oe_whitelabel/pull/390/commits/360baf600e9da6b426bf3c6d1632f32c591751d2).

oe_whitelabel.info.yml
kp77: Yes, modal_page's css overrides are not needed for us (it would be more complicated and uglier to keep them and override them again). The js on the other hand is necessary as it supports the features that can be configured for the modals.

kp77: The 'modal-page-dialog-api' library is only attached by modal_page when the module is configured specifically to use the dialog api provider instead of bootstrap. I'm not sure we need to cover this case: if a project configures the module with that provider, then I think it's up to them to decide what happens with these libraries.

resources/sass/components/_modal_page.scss
kp77: This is needed when the 'Display title' is unchecked, but the close button is enabled  (so there's no title heading element in the header just the button) in 'toast' display mode. We have the same rule for the modal header btw, it's needed because the toast header has 'toast-header' class.

resources/sass/components/_modal_page.scss
kp77: Yes, these are copied from `.modal-footer` rules from Bootstrap, to achieve the same look in `.toast-footer` as in `.modal-footer`.

https://github.com/twbs/bootstrap/blob/main/scss/_modal.scss#L160

Reason: `.toast-footer` is not a thing in Bootstrap, there's no styling for it. It is allowed by Bootstrap to put any custom content here though, and what we needed is a footer similar to modal footer (with the same styling).

But indeed, there's a difference: in modal the [parent element](https://github.com/twbs/bootstrap/blob/main/scss/_modal.scss#L97) is `.modal-content` and that has `display: flex`. This is not the case for toast where the parent is directly the `.toast` element, and that doesn't have `display: flex`. I removed the `flex-shrink` in [016edff](https://github.com/openeuropa/oe_whitelabel/pull/390/commits/016edffb8c8df59b2d4c4cb99737688872b5662b) as it doesn't make sense here.

modules/oe_whitelabel_modal_page/README.md
kp77: I used the word 'feature' instead of 'module' in [ffd2c61](https://github.com/openeuropa/oe_whitelabel/pull/390/commits/ffd2c61e258ca20c04f524afc4917f37a00fa308). I think it's still better to add this info to this readme, because it will be relevant for those who enable this module.

modules/oe_whitelabel_modal_page/README.md
kp77: I added a note for both points about the fact that they are implemented in the theme in [ffd2c61](https://github.com/openeuropa/oe_whitelabel/pull/390/commits/ffd2c61e258ca20c04f524afc4917f37a00fa308). I think it's still better to add this info to this readme, because it will be relevant for those who enable this module.

oe_whitelabel.theme
kp77: I was not aware that trailing comma is needed also for argument lists, but apparently it is not failing in tests at the moment. The code has been removed anyway, so we can resolve this.

oe_whitelabel.info.yml
kp77: I simplified the library removal and moved it to oe_whitelabel_library_info_alter in [a5263bc](https://github.com/openeuropa/oe_whitelabel/pull/390/commits/a5263bc58b7f428e8a91260b22412333fe070dc7). Kept the tests as discussed.

modules/oe_whitelabel_modal_page/tests/src/Functional/ModalFormFieldAccessTest.php
kp77: Added docs in [360baf6](https://github.com/openeuropa/oe_whitelabel/pull/390/commits/360baf600e9da6b426bf3c6d1632f32c591751d2)

templates/modal_page/modal-page-modal.html.twig
```diff
@@ -0,0 +1,118 @@
...
+    {% endif %}
+
+    {% set _button_attributes = create_attribute({
+      'class': ['js-modal-page-ok-button'],
+      'data-bs-dismiss': _is_toast ? 'toast' : 'modal'
+    }) %}
+
+    {% if enable_left_button %}
+      {{ pattern('button', {
+        'variant': 'default',
```
tibi2303: there is no default variant (primary is default in bcl) and i am not sure what should go here - I think maybe secondary.

templates/modal_page/modal-page-modal.html.twig
```diff
@@ -0,0 +1,118 @@
...
+        'attributes': _button_attributes
+      }) }}
+    {% endif %}
+
+    {% if enable_right_button %}
+      {% if enable_redirect_link %}
+        {% set _button_attributes = _button_attributes.setAttribute('data-redirect', redirect_link) %}
+      {% endif %}
+      {{ pattern('button', {
+        'variant': 'primary',
```
tibi2303: default is primary so you should be able to safely delete this

templates/modal_page/modal-page-modal.html.twig
```diff
@@ -0,0 +1,118 @@
...
+    {% if display_title  %}
+      <h5 class="mb-0">{{ title }}</h5>
+    {% endif %}
+    {% if _is_toast is empty and display_button_x_close %}
+      {% include '@oe-bcl/button' with {
+        clean_class: TRUE,
+        attributes:  create_attribute({
+          'class': ['btn-close', 'js-modal-page-ok-button'],
+          'data-bs-dismiss': 'modal',
+          'aria-label': 'Close'
```
tibi2303: should we run it through |t?

tibi2303:
@kp77 thanks a lot of peter for this pr, 3 small remark and i will approve.

templates/modal_page/modal-page-modal.html.twig
donquixote: did we have a spec saying it should be in the body area?
to me it looks nicer in the footer..

kp77: Yes, this was in the requirements. I added a comment on the OEL review ticket and linked the wiki page and the figma mockup.

templates/modal_page/modal-page-modal.html.twig
kp77: Yes, you are right. Originally, the mockup used a button with no styling - or no variant, one that looks like here: https://getbootstrap.com/docs/5.3/components/buttons/#base-class

This is not supported by the BCL component at the moment (BCL always sets a variant class on the button), and I could only achieve it with this fake 'default' workaround.

But in the meantime the mockup was updated to use secondary, so we should change it to that.

However, this change revealed, that I didn't use proper margins on the toast footer elements (this was not visible that much when the button didn't have color). So to fix this, I also added the spacing fix inside the toast footer to ./resources/sass/components/_modal_page.scss

Another issue that this change revealed is that the _button_attributes variable (created to store the defaults for both buttons) is changed by the BCL component when passed to the first button. After that, the BCL component adds classes to the attributes (for the variant) and when I try to use it for the second button, it gets 'contaminated' by the first button's variant class. So I modified that part to avoid this.

All changes are in this commit: [25080bb](https://github.com/openeuropa/oe_whitelabel/pull/390/commits/25080bb5afef6d186c175ca5d2c7d448b96df6b3)

templates/modal_page/modal-page-modal.html.twig
kp77: Removed.

templates/modal_page/modal-page-modal.html.twig
kp77: Yes, added the filter, thanks for noticing!

tibi2303 (dismissed):

tibi2303 (dismissed):

donquixote (dismissed):

tibi2303 (dismissed):

tibi2303 (approved):

PR #393, opened 2025-11-13
OEL-4180 ff: Use remote patch urls.

donquixote:
- Follow-up to #386 

Local patch url does not work when the package is used as a dependency.
See https://github.com/cweagans/composer-patches/issues/549


composer.json
```diff
@@ -104,10 +104,10 @@
         },
          "patches": {
             "openeuropa/oe_paragraphs": {
-                "Support for Drupal 11.2. https://github.com/openeuropa/oe_paragraphs/pull/267:": "./patches/oe_paragraphs-PR_267-05112025.patch"
+                "Support for Drupal 11.2. https://github.com/openeuropa/oe_paragraphs/pull/267:": "https://raw.githubusercontent.com/openeuropa/oe_whitelabel/refs/tags/1.27.0/patches/oe_paragraphs-PR_267-05112025.patch"
```
donquixote, line 107: @AaronGilMartinez This solution is a bit dirty/unorthodox:
We need a stable url where we get this patch.
We tried to use a commit patch url from oe_content, but that did not apply cleanly.
Here we take advantage of the fact that we earlier released with the local patch.
So the main flaw of the solution is that it is not repeatable.

Another concern of me was that we want to avoid explicitly relying on github. But this is a temporary solution, we will drop these patches soon.

piotrsmykaj (approved):

AaronGilMartinez (approved):

PR #406, opened 2026-01-12
OEL-4322: Migrate Alert and Description list components.

piotrsmykaj:


templates/agenda/oe-agenda-session--oe-default.html.twig
```diff
@@ -7,14 +7,14 @@
...
-  }) }}
-{% endmacro %}
+{#{% macro icon(name, size, attributes) %}#}
+{#  {{ pattern('icon', {#}
+{#    name: name,#}
+{#    path: bcl_icon_path,#}
+{#    size: size,#}
+{#    attributes: attributes,#}
+{#  }) }}#}
+{#{% endmacro %}#}
```
donquixote, line 17: We should not remove the entire macro, only the `pattern()` part inside.
The `_self.macro` is still used below.

This said, I really think we should find a solution where we don't have to modify all these templates.

We should override/replace the twig function `pattern()` from the legacy module, and have it do something if there is no sdc/component with that name.

There are two options:
1. Render an empty string (this is equivalent to what we currently do in this PR, just less noise in the templates).
2. Render the old pattern. This is a lot more work, but would make the tests pass and avoid empty content. If we do this, we have to compare the effort to forking the ui_patterns v1 module.

donquixote: Other options I can think of:
- (not viable) Auto-migrate all patterns, even if they will be broken as component. We likely get errors and broken pages from broken components.
- First refactor all patterns to drop preprocess and prepare for migration (while on ui patterns v1). Then auto-migrate.

PR #407, opened 2026-01-15
OEL-4458: Multi-value 'site_owner' twig variable.

julien-:
The 'site_owner' variable in some twig templates used to be a single value.
In newer versions of oe_corporate_blocks it is an array.
The twig templates need to be updated accordingly.

(previously it was suggested to rename the variable, but that is not feasible given that oe_theme already uses the old variable name with multiple value)

### DIFF
--- a/templates/overrides/navigation/oe-corporate-blocks-neutral-footer.html.twig
+++ b/templates/overrides/navigation/oe-corporate-blocks-neutral-footer.html.twig
@@ -69,7 +69,10 @@
 
 {% set footer_top_left_content %}
   <p class="fw-bold mb-2">{{ site_specific_footer.site_identity }}</p>
-  <p> {{ site_owner ? 'This site is managed by the @name'|t({'@name': site_owner}) }} </p>
+
+  {% if site_owners %}
+    <p>{{ 'This site is managed by the @name'|t({'@name': site_owners|join(', ')}) }}</p>
+  {% endif %}
 
   {% if accessibility_link %}
     {% set _accessibility_link = pattern('link', {
--- a/templates/overrides/navigation/oe-corporate-blocks-neutral-footer.html.twig
+++ b/templates/overrides/navigation/oe-corporate-blocks-neutral-footer.html.twig
@@ -70,8 +70,8 @@
 {% set footer_top_left_content %}
   <p class="fw-bold mb-2">{{ site_specific_footer.site_identity }}</p>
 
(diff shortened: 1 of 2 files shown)


templates/overrides/navigation/oe-corporate-blocks-neutral-footer.html.twig
```diff
@@ -69,7 +69,10 @@
 
 {% set footer_top_left_content %}
   <p class="fw-bold mb-2">{{ site_specific_footer.site_identity }}</p>
-  <p> {{ site_owner ? 'This site is managed by the @name'|t({'@name': site_owner}) }} </p>
+
+  {% if site_owners %}
+    <p>{{ 'This site is managed by the @name'|t({'@name': site_owners|join(', ')}) }}</p>
+  {% endif %}
```
AaronGilMartinez, line 75: This fixes an issue thank you @julien- . 

We can fix this issue by using the existing name, personally I don't have a preference for the name but must be a reason to keep the same.

donquixote: We must keep the old name, because that's already used in `oe_theme`, see https://github.com/openeuropa/oe_theme/commit/de4c69196eb4368f97e29be4452625b8366ef8f8#diff-780633d7bdb3030cc3ee0933a53891fee572de8a07298ff9e9b0466aaed6fdf2

donquixote: Btw interesting how the old version had spaces around the text inside the `<p>` tags. Probably safe to remove, if anything they would be undesirable.

templates/overrides/navigation/oe-corporate-blocks-neutral-footer.html.twig
```diff
@@ -69,7 +69,10 @@
 
 {% set footer_top_left_content %}
   <p class="fw-bold mb-2">{{ site_specific_footer.site_identity }}</p>
-  <p> {{ site_owner ? 'This site is managed by the @name'|t({'@name': site_owner}) }} </p>
+
+  {% if site_owner %}
+    <p>{{ 'This site is managed by the @name'|t({'@name': site_owner|join(', ')}) }}</p>
```
donquixote, line 74: I tested how this looks like now.

I don't like it that much for long owner names. (see Jira comments)

PR #412, opened 2026-01-30
OEL-3347: Update banner template for copyright.

tibi2303:


modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.module
```diff
@@ -758,6 +758,10 @@ function _oe_whitelabel_paragraphs_preprocess_paragraph_media(array &$variables)
 
   $source = $media->getSource();
 
+  if ($media->hasField('field_media_copyright') && !$media->get('field_media_copyright')->isEmpty()) {
+    $variables['copyright'] = $media->get('field_media_copyright')->value;
+  }
+
```
donquixote, line 779: It hurts me that we have to hard-code this field name here, but I don't see a good alternative, so ok.

templates/paragraphs/paragraph--oe-banner.html.twig
```diff
@@ -25,6 +25,13 @@
 {% endif %}
 {% set _hero = 'hero' in pattern %}
 {% set _shade = 'shade' in variant %}
+{% set copyright_value = '' %}
+{% if not paragraph.get('field_oe_media').isEmpty() %}
+  {% set media_entity = paragraph.get('field_oe_media').entity %}
+  {% if media_entity and media_entity.hasField('field_media_copyright') %}
+    {% set copyright_value = media_entity.field_media_copyright.value %}
+  {% endif %}
+{% endif %}
```
donquixote: In the other paragraph we use preprocess, but here we do the same logic in the template?
Is there a difference?

README.md
```diff
@@ -97,6 +97,26 @@ Enable the theme and set as default:
...
+The theme ships recipes that add a reusable copyright field to media types and expose it in view modes.
+
+1) Apply the recipes (they are not auto-applied):
+
+```bash
+# Run in your Drupal root where the theme is installed.
+./vendor/bin/drush recipe "/var/www/html/build/themes/contrib/oe_whitelabel/recipes/media_image_copyright"
+./vendor/bin/drush recipe "/var/www/html/build/themes/contrib/oe_whitelabel/recipes/media_remote_video_copyright"
+./vendor/bin/drush recipe "/var/www/html/build/themes/contrib/oe_whitelabel/recipes/media_av_portal_photo_copyright"
+./vendor/bin/drush recipe "/var/www/html/build/themes/contrib/oe_whitelabel/recipes/media_av_portal_video_copyright"
```
donquixote: We no longer have video copyright recipe.

donquixote (commented):
I tried the paragraphs after enabling (and downloading) some more modules that were needed.

Some observations:
- I did not find how to add "content banner", I only found "banner". Not sure if these are the same.
- In "banner" I did not see how to add a picture / media.
- In gallery paragraph I did not see the copyright (I think not implemented in whitelabel, only in oebt and bcl).

README.md
tibi2303: removed.

templates/paragraphs/paragraph--oe-banner.html.twig
tibi2303: indeed, i removed it

README.md
```diff
@@ -97,6 +97,29 @@ Enable the theme and set as default:
...
+./vendor/bin/drush recipe "/var/www/html/build/themes/contrib/oe_whitelabel/recipes/media_image_copyright"
+./vendor/bin/drush recipe "/var/www/html/build/themes/contrib/oe_whitelabel/recipes/media_av_portal_photo_copyright"
+```
+
+Then:
+
+1. Fill `field_media_copyright` on your media items (or set a default/backfill).
+2. For gallery/carousel usage, ensure your view mode maps
+   `field_media_copyright` to the `copyright` field (like
+   `oe_w_pattern_gallery_item` does).
```
donquixote: I don't see that gallery/carousel integration working in oe_whitelabel. I don't see a code change here in this PR. Also not sure what this instruction means exactly, is it talking about a template or about view mode config via field UI? (especially as a regular reader I would not know)

donquixote (commented):
I am not sure if this is considered finished..

Some points:
- We probably want some tests, _but_ for me it's ok to get UAT feedback first.
- We should do the gallery paragraph integration.
- Perhaps we want to import the recipes in runner.yml.dist. Or is there a reason not to?

tibi2303:
@donquixote I added the test now just to be sure copyright shows okay. I added also the gallery/carousel integration of copyright field and the import in runner of recipes.

README.md
tibi2303: i removed that part. it was missleading and not correct.

modules/oe_whitelabel_helper/src/Plugin/Field/FieldFormatter/MediaGalleryFormatter.php
```diff
@@ -99,6 +99,16 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
...
+
+      // Keep gallery copyright support simple: if the standard media copyright
+      // field is present and not explicitly mapped already, expose it.
+      if (
+        !isset($gallery_items[$delta]['copyright']) &&
+        $entity->hasField('field_media_copyright') &&
+        !$entity->get('field_media_copyright')->isEmpty()
+      ) {
+        $gallery_items[$delta]['copyright'] = $entity->get('field_media_copyright')->value;
+      }
```
donquixote, line 101: It's a bit odd, we are creating that feature where you map a field, but then we ignore it.
And we will have to support this oddity in the future.

Instead, we should set this pattern mapping in the recipe.

BUT, what if at the time the recipe is imported, that gallery view mode does not exist?
There are two possible solutions:
- Separate recipes for the gallery mapping. (not convenient)
- A custom recipe action `simpleConfigUpdateIfExists`, which only updates a config entity if it exists. I can help with that.

donquixote (dismissed):

modules/oe_whitelabel_paragraphs/config/install/field.field.paragraph.oe_list_item.field_oe_image_copyright.yml
donquixote, line 1: For the media types we used recipes to make this field optional.
Here we use `config/install/`.

At first I thought this is odd.
But I see that we have more fields like this.

So I'd say it's ok :)

donquixote (dismissed):
I approve, but you may look at my comment anyway.

AaronGilMartinez (approved):

donquixote (approved):

PR #414, opened 2026-02-04
OEL-4032: Use 'gap' for icon spacing in button and link

tibi2303:
- See https://github.com/openeuropa/oe_bootstrap_theme/pull/537
- See https://github.com/openeuropa/bootstrap-component-library/pull/733

### DIFF
--- a/modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
+++ b/modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
@@ -49,7 +49,6 @@ function oe_whitelabel_multilingual_preprocess_links__oe_multilingual_content_la
       'label' => $languages[$translation->language()->getId()]->getName(),
       'current' => TRUE,
       'icon_position' => 'before',
-      'remove_icon_spacers' => FALSE,
       'icon' => [
         'name' => 'check-lg',
         'path' => $variables['bcl_icon_path'],
--- a/modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
+++ b/modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
@@ -53,7 +53,6 @@ function oe_whitelabel_multilingual_preprocess_links__oe_multilingual_content_la
         'name' => 'check-lg',
         'path' => $variables['bcl_icon_path'],
         'size' => 'xs',
-        '#attributes' => ['class' => ['me-2']],
       ],
     ];
   }
--- a/modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
+++ b/modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
@@ -230,7 +230,7 @@ class SearchBlockTest extends KernelTestBase {
     $this->assertStringContainsString('btn', $button->attr('class'));
     $this->assertStringContainsString('btn-primary', $button->attr('class'));
     $this->assertStringContainsString('btn-md', $button->attr('class'));
-    $icon = $button->filter('.bi.icon--s.me-md-2-5.bcl-search-form__btn_icon');
+    $icon = $button->filter('.bi.icon--s.bcl-search-form__btn_icon');
     $this->assertCount(1, $icon);
     $label = $button->filter('span.d-none.d-lg-inline-block');
     $this->assertCount(1, $label);
--- a/modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
+++ b/modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
@@ -231,7 +231,7 @@ class SearchBlockTest extends KernelTestBase {
     $this->assertStringContainsString('btn-primary', $button->attr('class'));
     $this->assertStringContainsString('btn-md', $button->attr('class'));
     $this->assertStringContainsString('gap-2-5', $button->attr('class'));
-    $icon = $button->filter('.bi.icon--s.bcl-search-form__btn_icon');
+    $icon = $button->filter('.bi.icon--fluid.bcl-search-form__btn_icon');
     $this->assertCount(1, $icon);
     $label = $button->filter('span.d-none.d-lg-inline-block');
     $this->assertCount(1, $label);
--- a/modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
+++ b/modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
@@ -230,6 +230,7 @@ class SearchBlockTest extends KernelTestBase {
     $this->assertStringContainsString('btn', $button->attr('class'));
     $this->assertStringContainsString('btn-primary', $button->attr('class'));
     $this->assertStringContainsString('btn-md', $button->attr('class'));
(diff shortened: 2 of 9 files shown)


modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
donquixote: I see we are removing the `'remove_icon_spacers' => FALSE,` because that is already the default.

But I notice this `me-2` class which is odd.
In fact in oesa I saw the output for this block and the 'me-2' class is nowhere to be found.
Maybe with the '#' in front of '#attributes' it cannot work.

(Maybe we can clean this up another time, I just noticed it and at first it seemed relevant but then not really)

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
```diff
@@ -230,7 +230,7 @@ public function testHeaderSearchBlockRendering(): void {
     $this->assertStringContainsString('btn', $button->attr('class'));
     $this->assertStringContainsString('btn-primary', $button->attr('class'));
     $this->assertStringContainsString('btn-md', $button->attr('class'));
-    $icon = $button->filter('.bi.icon--s.me-md-2-5.bcl-search-form__btn_icon');
+    $icon = $button->filter('.bi.icon--s.bcl-search-form__btn_icon');
```
donquixote: The 'me-md-2-5' is gone, but now we expect there to be 'gap' class on the button, should we check for that?

modules/oe_whitelabel_multilingual/oe_whitelabel_multilingual.module
tibi2303: we should clean up stuff when we find, i deleted it.. it didnt worked

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
tibi2303: added

modules/oe_whitelabel_search/tests/src/Kernel/SearchBlockTest.php
```diff
@@ -230,7 +230,7 @@ public function testHeaderSearchBlockRendering(): void {
     $this->assertStringContainsString('btn', $button->attr('class'));
     $this->assertStringContainsString('btn-primary', $button->attr('class'));
     $this->assertStringContainsString('btn-md', $button->attr('class'));
-    $icon = $button->filter('.bi.icon--s.me-md-2-5.bcl-search-form__btn_icon');
+    $icon = $button->filter('.bi.icon--s.bcl-search-form__btn_icon.gap-2-5');
```
donquixote: This causes failing test.
The `.gap-2-5` is on the parent element.
So
```
$this->assertStringContainsString('gap-2-5', $button->attr('class'));
```
(I find it strange how we assert these classes with "contains string", this would also pass if we have a longer class like "abc-gap-2-5-xyz". But we already do it for other classes.)

templates/overrides/search/input--submit--header--oe-whitelabel-search-form.html.twig
donquixote, line 13: Do we still need to render the icon + text into a separate variable?
What would happen if we just pass icon data to the 'button' template?

I think we need the `<span>` wrapper on the text so it gets hidden in mobile, but maybe that does not need to include the icon.

templates/overrides/navigation/block--oe-authentication-login-block.html.twig
```diff
@@ -29,7 +29,6 @@
       size: 'fluid',
     },
     icon_position: 'before',
-    remove_icon_spacers: true,
```
donquixote, line 32: Now that we remove this and rely on the default implementation, there is also some CSS we can remove.
In `resources/sass/components/_header.scss` there is:
```css
top-navigation-link {
  display: flex;
  align-items: center;
  gap: 0.75rem;
```
Some of it now redundant.

donquixote: Although.. maybe we still need it because other items might have that class?
What do you think?

templates/overrides/navigation/links--language-block.html.twig
```diff
@@ -38,7 +38,6 @@
     size: 'fluid',
   },
   icon_position: 'before',
-  remove_icon_spacers: true,
```
donquixote, line 41: Same here: Should we clean up css, or remove the "top-navigation-link" class from this component, or leave it as is for maximum BC?

templates/overrides/navigation/block--oe-authentication-login-block.html.twig
tibi2303: we still need, the other elements are top-navigation-link still

resources/sass/components/_header.scss
```diff
@@ -1,7 +1,5 @@
 .top-navigation-link {
-  display: flex;
   align-items: center;
-  gap: 0.75rem;
```
donquixote, line 4: Ok, fingers crossed!
It will work fine for our known existing links where we have the utility classes from bcl-link template.

templates/overrides/search/input--submit--header--oe-whitelabel-search-form.html.twig
```diff
@@ -7,25 +7,20 @@
...
 {{ pattern('button', {
   'variant': 'primary',
   'label': label,
+  'icon': {
+    'name': 'search',
+    'path': bcl_icon_path,
+    'attributes': create_attribute().addClass(['bcl-search-form__btn_icon']),
+  },
+  'icon_position': 'before',
+  'icon_display': 'inline-flex',
```
donquixote: This one is a button, not a link. so do we need the `icon_display` parameter?

tests/src/Kernel/MultilingualBlockTest.php
```diff
@@ -25,6 +25,7 @@ class MultilingualBlockTest extends KernelTestBase {
     'system',
     'ui_patterns',
     'ui_patterns_library',
+    'ui_patterns_settings',
```
donquixote, line 28: Strange, why was it needed?

tests/src/Kernel/AuthenticationBlockTest.php
```diff
@@ -25,6 +25,7 @@ class AuthenticationBlockTest extends KernelTestBase {
     'system',
     'ui_patterns',
     'ui_patterns_library',
+    'ui_patterns_settings',
```
donquixote, line 28: Again, what has changed that makes this necessary?

templates/overrides/search/input--submit--header--oe-whitelabel-search-form.html.twig
tibi2303: removed.

tests/src/Kernel/MultilingualBlockTest.php
tibi2303: to be able to add the params from the yml file into the page that it's testing in

donquixote (approved):
All looks good now!
Of course we have to merge it one by one starting with BCL.

donquixote:
Also we have to fix CI for 1.x.

PR #416, opened 2026-02-05
OEL-4455: Require (dev) drupal/modal_page:^6.0-beta11

donquixote:
This make sure the new patch version will apply.

- follows #413 

other:
- follows #408 
- follows #390

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "drupal/extra_field": "^2.3 || ^3.0@beta",
         "drupal/facets_form": "^1.0.0-alpha6",
         "drupal/field_group": "^3.4",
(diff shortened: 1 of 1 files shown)


donquixote:
actually.. before we merge:
This is require-dev, so it does not really make a difference.
if we care about projects using this, we could set a "conflict". But we generally don't do this.
So do we even care now?

AaronGilMartinez:
> actually.. before we merge: This is require-dev, so it does not really make a difference. if we care about projects using this, we could set a "conflict". But we generally don't do this. So do we even care now?

Any project using modal_page will see the patch not applying, we apply the patch, I think the change it's ok, it's more informative change.

AaronGilMartinez (approved):

PR #418, opened 2026-02-06
UCPKN-928: Add option to configure required state of search block input.

peterserfozo:


tibi2303 (approved):
Changes seems fine to me, but please let's wait for business also. We will talk about it in next refinement.

tibi2303:
@AaronGilMartinez what do you think of this changes?

AaronGilMartinez (approved):

PR #421, opened 2026-02-15
OEL-0000: Update GalleryParagraphTest for changed behavior of avportal

donquixote:
Some behavior in media_avportal has changed:
- Thumbnail dimensions
- Remote image url.

Not sure why, we should find out before we merge.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "drupal/extra_field": "^2.3 || ^3.0@beta",
         "drupal/facets_form": "^1.0.0-alpha6",
         "drupal/field_group": "^3.4",
-        "drupal/modal_page": "^6.0-beta10",
+        "drupal/modal_page": "^6.0-beta11",
         "drupal/paragraphs": "^1.17",
         "drupal/search_api": "^1.29",
         "drupal/search_api_autocomplete": "^1.7",


modules/oe_whitelabel_paragraphs/tests/src/Functional/GalleryParagraphTest.php
```diff
@@ -80,13 +80,13 @@ public function testRendering(): void {
...
           'rendered' => sprintf(
-            '<img loading="lazy" src="%s" width="639" height="426" alt="Euro with miniature figurines" class="img-fluid">',
+            '<img loading="lazy" src="%s" width="800" height="533" alt="Euro with miniature figurines" class="img-fluid">',
             $fn_get_filepath($avportal_photo, 'thumbnail')
           ),
         ],
         'media' => [
           'caption_title' => 'Euro with miniature figurines',
-          'rendered' => '<img class="avportal-photo img-fluid" alt="Euro with miniature figurines" data-src="https://ec.europa.eu/avservices/avs/files/video6/repository/prod/photo/store/store2/4/P038924-352937.jpg">',
+          'rendered' => '<img class="avportal-photo img-fluid" alt="Euro with miniature figurines" data-src="https://ec.europa.eu/avservices/avs/files/video6/repository/prod/photo/store//03/89/24/P-038924_00-15_02-HIGH-625541.jpg">',
```
donquixote: I suspect that this is not stable, because it contains what could be a date (00-15_02).

PR #423, opened 2026-02-16
OEL-4522: Stabilize media sizes for AV Portal media tests.

tibi2303:


tests/src/FunctionalJavascript/ContentProjectRenderTest.php
```diff
@@ -594,4 +624,31 @@ protected function assertTeaserStatusBadge(NodeElement $wrapper, string $status_
     $this->assertTrue($badges[0]->hasClass($status_class));
   }
 
+  /**
+   * Gets image dimensions for a media field, falling back to the file itself.
+   */
+  private function getImageDimensions($entity, string $field): array {
```
AaronGilMartinez: Since we are duplicating the function I think would be good to add a trait for this. Remember to check if this fits in a existing or it's better a dedicated.

modules/oe_whitelabel_paragraphs/tests/src/Functional/GalleryParagraphTest.php
```diff
@@ -57,44 +57,67 @@ public function testRendering(): void {
...
 
+    $media_storage = \Drupal::entityTypeManager()->getStorage('media');
+    $image = $media_storage->load($image->id());
+    $avportal_photo = $media_storage->load($avportal_photo->id());
+    $avportal_video = $media_storage->load($avportal_video->id());
+    $video = $media_storage->load($video->id());
+
     $file_url_generator = \Drupal::service('file_url_generator');
     $fn_get_filepath = static fn($entity, $field) => $file_url_generator->generate($entity->get($field)->entity->getFileUri())->toString();
+    [$image_width, $image_height] = $this->getImageDimensions($image, 'oe_media_image');
```
donquixote, line 76: These should be stable, no?

tests/src/FunctionalJavascript/ContentProjectRenderTest.php
donquixote: I see we are duplicating this method, should we move it to a trait?

modules/oe_whitelabel_paragraphs/tests/src/Functional/GalleryParagraphTest.php
```diff
@@ -213,4 +238,31 @@ protected function renderParagraph(ParagraphInterface $paragraph, ?string $langc
     return (string) $this->container->get('renderer')->renderRoot($render);
   }
 
+  /**
+   * Gets image dimensions for a media field, falling back to the file itself.
+   */
+  private function getImageDimensions($entity, string $field): array {
```
donquixote: We should have `@param` and `@return` docs, and we can put parameter type in php.

For `$entity`, I first thought we should rename it to `$media`, but now I think it is actually fine because this method works for any entity with an image reference field.

The field should be an image field, no? Not a media field. (that is, the reference target is image (file entity) not media.)

(Please don't use this 1:1, review if I am missing something)

```suggestion
  /**
   * Gets image dimensions for an image field, falling back to the file itself.
   * 
   * @param \Drupal\...EntityInterface $entity
   * @param string $field
   *   The name of an image reference field on the entity.
   * 
   * @return array{int, int}
   *    Image width and height.
   */
  private function getImageDimensions(EntityInterface $entity, string $field): array {
```

modules/oe_whitelabel_paragraphs/tests/src/Functional/GalleryParagraphTest.php
```diff
@@ -213,4 +238,31 @@ protected function renderParagraph(ParagraphInterface $paragraph, ?string $langc
...
+      return [$width, $height];
+    }
+
+    $uri = $item->entity?->getFileUri();
+    if ($uri) {
+      $image = \Drupal::service('image.factory')->get($uri);
+      if ($image->isValid()) {
+        return [$image->getWidth(), $image->getHeight()];
+      }
+    }
```
donquixote: We could have some inline comments explaining the different cases.
- The field is empty, return `[0, 0]`.
- The width and height are stored on the image field item, return that.
- The image field has no dimensions stored for the image, instead we look at the image field directly. (why and when does this happen?)

modules/oe_whitelabel_paragraphs/tests/src/Functional/GalleryParagraphTest.php
```diff
@@ -213,4 +238,31 @@ protected function renderParagraph(ParagraphInterface $paragraph, ?string $langc
...
+
+    return [$width, $height];
```
donquixote: At this point it is just `[0, 0]`.
But do we even support this case?
Otherwise we could say `$this->fail('The image dimensions could not be determined.');`.

modules/oe_whitelabel_paragraphs/tests/src/Functional/GalleryParagraphTest.php
```diff
@@ -213,4 +238,31 @@ protected function renderParagraph(ParagraphInterface $paragraph, ?string $langc
     return (string) $this->container->get('renderer')->renderRoot($render);
   }
 
+  /**
+   * Gets image dimensions for a media field, falling back to the file itself.
+   */
+  private function getImageDimensions($entity, string $field): array {
+    $item = $entity->get($field)->first();
+    if (!$item) {
+      return [0, 0];
```
donquixote: Is this a case we want to support?
We could say `$this->fail('The image dimensions could not be determined.');`.
We don't expect the dimensions to be 0, 0 in the output, so..

modules/oe_whitelabel_paragraphs/tests/src/Functional/GalleryParagraphTest.php
```diff
@@ -57,44 +57,67 @@ public function testRendering(): void {
...
     $paragraph->save();
 
+    $media_storage = \Drupal::entityTypeManager()->getStorage('media');
+    $image = $media_storage->load($image->id());
+    $avportal_photo = $media_storage->load($avportal_photo->id());
+    $avportal_video = $media_storage->load($avportal_video->id());
+    $video = $media_storage->load($video->id());
+
     $file_url_generator = \Drupal::service('file_url_generator');
     $fn_get_filepath = static fn($entity, $field) => $file_url_generator->generate($entity->get($field)->entity->getFileUri())->toString();
```
donquixote:
```suggestion
    $fn_get_filepath = static fn ($entity, $field) => $file_url_generator->generate($entity->get($field)->entity->getFileUri())->toString();
```
EDIT: Sorry, I see this line already existed)

donquixote: Also when we get this file path (which is actually a url), should we assert that it is not empty and shaped like a url?

modules/oe_whitelabel_paragraphs/tests/src/Functional/GalleryParagraphTest.php
tibi2303: Yes. We reload the media after render, and getImageDimensions() now fails if it can’t determine dimensions.

tests/src/FunctionalJavascript/ContentProjectRenderTest.php
tibi2303: moved to mediaCreationTrait, should we do it in another one? i think it's fine in there..

modules/oe_whitelabel_paragraphs/tests/src/Functional/GalleryParagraphTest.php
tibi2303: Seems okay like you said, i implemented like this.

modules/oe_whitelabel_paragraphs/tests/src/Functional/GalleryParagraphTest.php
tibi2303: I changed how dimensions are get and I put fails in all cases

modules/oe_whitelabel_paragraphs/tests/src/Functional/GalleryParagraphTest.php
tibi2303: added.

modules/oe_whitelabel_paragraphs/tests/src/Functional/GalleryParagraphTest.php
tibi2303: and also added comments.

modules/oe_whitelabel_paragraphs/tests/src/Functional/GalleryParagraphTest.php
tibi2303: changed

modules/oe_whitelabel_paragraphs/tests/src/Functional/GalleryParagraphTest.php
tibi2303: asserted for all media types.

tests/src/FunctionalJavascript/ContentProjectRenderTest.php
AaronGilMartinez: I think it's fine.

modules/oe_whitelabel_paragraphs/tests/src/Functional/GalleryParagraphTest.php
AaronGilMartinez: Ok to me, resolving.

modules/oe_whitelabel_paragraphs/tests/src/Functional/GalleryParagraphTest.php
AaronGilMartinez: I think it's better now with multiple conditions where the assertion fails and docs.

modules/oe_whitelabel_paragraphs/tests/src/Functional/GalleryParagraphTest.php
AaronGilMartinez: I think its fine like this, asserting the URL seems a bit too much, this should be done at other level.

AaronGilMartinez (approved):

modules/oe_whitelabel_paragraphs/tests/src/Functional/GalleryParagraphTest.php
```diff
@@ -57,44 +57,75 @@ public function testRendering(): void {
...
-    $fn_get_filepath = static fn($entity, $field) => $file_url_generator->generate($entity->get($field)->entity->getFileUri())->toString();
+    $fn_get_file_url = static fn($entity, $field) => $file_url_generator->generate($entity->get($field)->entity->getFileUri())->toString();
+    $image_src = $fn_get_file_url($image, 'oe_media_image');
+    $avportal_photo_thumb_src = $fn_get_file_url($avportal_photo, 'thumbnail');
+    $avportal_video_thumb_src = $fn_get_file_url($avportal_video, 'thumbnail');
+    $video_thumb_src = $fn_get_file_url($video, 'thumbnail');
+    foreach ([$image_src, $avportal_photo_thumb_src, $avportal_video_thumb_src, $video_thumb_src] as $src) {
+      $this->assertNotSame('', $src);
+      $this->assertNotFalse(parse_url($src));
+    }
```
donquixote, line 75: I was thinking this assert could be part of $fn_get_filepath.
But ok like this.

tests/src/Traits/MediaCreationTrait.php
```diff
@@ -206,4 +207,51 @@ protected function createMediaByBundle(string $bundle, array $values = []): Medi
...
+   *
+   * Uses stored dimensions when available, otherwise falls back to the image
+   * file to keep test expectations stable.
+   *
+   * @param \Drupal\Core\Entity\FieldableEntityInterface $entity
+   *   The entity containing the image field.
+   * @param string $field
+   *   The image field machine name.
+   *
+   * @return int[]
```
donquixote, line 221:
```suggestion
   * @return array{int, int}
```
would be more precise, but ok.

modules/oe_whitelabel_paragraphs/tests/src/Functional/GalleryParagraphTest.php
donquixote: the point was if it is empty for some reason we may never know.

PR #428, opened 2026-03-03
OEL-4510: Update to Drupal 11.3.x.

tibi2303:
- New PR: #434

### DIFF
--- a/modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.info.yml
+++ b/modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.info.yml
@@ -5,6 +5,7 @@ package: OpenEuropa Whitelabel Theme
 core_version_requirement: ^10 || ^11
 dependencies:
   - description_list_field:description_list_field
+  - drupal:node_storage_body_field
   - oe_content:oe_content_timeline_field
   - oe_paragraphs:oe_paragraphs
   - oe_paragraphs:oe_paragraphs_description_list
--- a/modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.info.yml
+++ b/modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.info.yml
@@ -5,7 +5,6 @@ package: OpenEuropa Whitelabel Theme
 core_version_requirement: ^10 || ^11
 dependencies:
   - description_list_field:description_list_field
-  - drupal:node_storage_body_field
   - oe_content:oe_content_timeline_field
   - oe_paragraphs:oe_paragraphs
   - oe_paragraphs:oe_paragraphs_description_list
--- a/modules/oe_whitelabel_paragraphs/tests/src/Functional/ParagraphsTest.php
+++ b/modules/oe_whitelabel_paragraphs/tests/src/Functional/ParagraphsTest.php
@@ -25,6 +25,9 @@ class ParagraphsTest extends BrowserTestBase {
   protected static $modules = [
     'node',
     'composite_reference',
+    'oe_content',
+    'oe_content_timeline_field',
+    'oe_paragraphs_timeline',
     'oe_paragraphs_carousel',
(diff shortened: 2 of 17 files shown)


tests/src/Functional/WhitelabelBrowserTestBase.php
```diff
@@ -25,6 +25,11 @@ abstract class WhitelabelBrowserTestBase extends BrowserTestBase {
    * {@inheritdoc}
    */
   protected function setUp(): void {
+    if (is_dir(DRUPAL_ROOT . '/core/modules/node/modules/node_storage_body_field')) {
+      static::$modules[] = 'node_storage_body_field';
+      static::$modules = array_values(array_unique(static::$modules));
+    }
```
AaronGilMartinez: See https://github.com/openeuropa/oe_starter_content/pull/51 for body.

modules/oe_whitelabel_paragraphs/tests/src/Functional/ParagraphsTest.php
```diff
@@ -58,10 +67,11 @@ public function testLinksBlockParagraph(): void {
     $this->assertSession()->fieldExists('oe_w_paragraphs[0][subform][oe_w_links_block_background]');
     $this->assertSession()->fieldExists('oe_w_paragraphs[0][subform][oe_w_links_block_orientation]');
 
-    $this->submitForm([], 'Add another item');
+    $this->submitForm([], 'oe_w_paragraphs_0_subform_field_oe_links_add_more');
 
     $values = [
-      'title[0][value]' => 'Test Links block node title',
+      'title[0][value]' => $title,
+      'oe_content_content_owner[0][target_id]' => self::CONTENT_OWNER,
```
AaronGilMartinez: Probably none of this changes related to author are needed (same for sparql trait), they come as part of oe_content that is not needed either, we just need to fix the issue in oe_starter_content.: https://github.com/openeuropa/oe_starter_content/pull/51.

modules/oe_whitelabel_paragraphs/tests/src/Functional/ParagraphsTest.php
```diff
@@ -448,9 +467,40 @@ protected function createTestContentType() {
...
+      ->save();
+    $display_repository->getViewDisplay('node', 'paragraphs_test', 'full')
+      ->setComponent('oe_w_paragraphs', $component)
+      ->save();
+  }
+
+  /**
+   * Opens the created node page by title.
+   */
+  protected function openCurrentNodePage(string $title): void {
```
AaronGilMartinez: We can use build/core/modules/node/tests/src/Traits/NodeCreationTrait.php and getNodeByTitle() no need to create a new method.

AaronGilMartinez (changes requested):

modules/oe_whitelabel_paragraphs/tests/src/Functional/ParagraphsTest.php
tibi2303: removed it and using nodecreationtrait now

tests/src/Functional/WhitelabelBrowserTestBase.php
tibi2303: nice, i removed it.

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.info.yml
```diff
@@ -4,6 +4,7 @@ description: Companion module to add support and customisations for oe_paragraph
 package: OpenEuropa Whitelabel Theme
 core_version_requirement: ^10 || ^11
 dependencies:
+  - oe_content:oe_content
```
AaronGilMartinez: I think we don't need oe_content with drupal:node_storage_body_field changes.

AaronGilMartinez (dismissed):

donquixote:
We keep this as archive, but closed.
- Fixed in #434

PR #433, opened 2026-03-30
OEL-4500: Migrate description_list to SDC component

piotrsmykaj:


modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.module
```diff
@@ -488,7 +488,8 @@ function oe_whitelabel_preprocess_paragraph__oe_description_list(array &$variabl
   /** @var Drupal\paragraphs\Entity\Paragraph $paragraph */
   $paragraph = $variables['paragraph'];
   $variables['title'] = $paragraph->get('field_oe_title')->value ?? '';
-  $variables['orientation'] = $paragraph->get('oe_w_orientation')->value;
+  $orientation = $paragraph->get('oe_w_orientation')->value;
+  $variables['orientation'] = $orientation === 'horizontal' ? 'horizontal' : 'default';
```
donquixote, line 492: We could use `match()`.
Is it an improvement? Not sure. It uses more lines but looks more clean, or not?
```php
$variables['orientation'] = match ($paragraph->get('oe_w_orientation')->value) {
  'horizontal' => 'horizontal',
  default => 'default',
};
```

templates/content/node--oe-sc-event--full.html.twig
```diff
@@ -35,11 +35,13 @@
 {% endif %}
 {# Set inpage navigation links and content #}
 {% set inpage_navigation_fields %}
-  {{ pattern('description_list', {
+  {{ include('oe_bootstrap_theme:description_list', {
```
donquixote, line 38: I don't find this `include(` syntax being used in core templates.
More common is `include '...' with {...} only`.
The `with_context = false` is equivalent to `only`, or not?

templates/overrides/navigation/links--language-block.html.twig
```diff
@@ -42,6 +42,6 @@
   attributes: extra_attributes
 }) }}
 
-{% include '@oe_whitelabel/patterns/modal/modal-language.html.twig' with language.modal only %}
+{% include '@oe-bcl/bcl-language-list/language-list.html.twig' with language.modal.language_list  %}
```
donquixote, line 45: Is this change needed?
If a pattern is not converted yet, we can simply continue using the pattern.

templates/patterns/modal/modal-language.html.twig
donquixote, line 1: Why was this pattern in whitelabel and not oebt, if this is using a bcl component?

donquixote: We could create as sdc in oebt, or we can create as sdc here.
Either wway we should keep the pattern for BC.

templates/patterns/README.txt
donquixote, line 1: Let's keep the pattern dir in whitelabel, until we officially remove patterns.

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.module
piotrsmykaj: If you ask my opinion, it's cleaner to keep it as is.

templates/content/node--oe-sc-event--full.html.twig
piotrsmykaj: In the SDC documentation, it's used `include()` in the examples:
- https://www.drupal.org/docs/develop/theming-drupal/using-single-directory-components/using-your-new-single-directory-component
- https://www.drupal.org/docs/develop/theming-drupal/using-single-directory-components/api-for-single-directory-components
- https://www.drupal.org/docs/develop/theming-drupal/using-single-directory-components/what-are-props-and-slots-in-drupal-sdc-theming

Also in the UI Patterns v2 as best practice is to use `include()`:
- https://project.pages.drupalcode.org/ui_patterns/2-authors/2-best-practices/

Yes, `with_context = false` is exactly equivalent to `only`.

templates/overrides/navigation/links--language-block.html.twig
piotrsmykaj: The pattern `modal-language` seems broken, and it doesn't work. In the pattern directory was only one file `templates/patterns/modal/modal-language.html.twig`, and a missing definition file "ui_patterns.yml". So I fixed it by using the BCL template, and the change was necessary, but it's debatable whether it should be included in this branch (which is about migrating description_list to SDC).

templates/patterns/modal/modal-language.html.twig
piotrsmykaj: I don't know what it was, please check my comment above: https://github.com/openeuropa/oe_whitelabel/pull/433/changes#r3148259819

templates/patterns/README.txt
piotrsmykaj: Please check my comment: https://github.com/openeuropa/oe_whitelabel/pull/433/changes#r3148259819
It depends. If this pattern is considered broken and not working properly, then it means that we don't have any patterns in oe_whitelabel, and this directory can be deleted.

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.module
donquixote: let's leave as is.

templates/content/node--oe-sc-event--full.html.twig
donquixote: This is still open.

templates/paragraphs/paragraph--oe-description-list.html.twig
```diff
@@ -13,8 +13,10 @@
...
 
-{{ pattern('description_list', {
+{{ include('oe_bootstrap_theme:description_list', {
   'items': items,
   'orientation': orientation,
   'attributes': attributes,
-}) }}
+},
+with_context = false
+) }}
```
donquixote, line 22: As above, let's do `include ... with ... only`, instead of `include(..., {..., with_context = false}`.

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.module
donquixote, line 492: To clarify: The goal of this change is to avoid that we get unexpected values?
Was this not a problem with the pattern?

Maybe we can keep this change just to be more robust.

templates/content/node--oe-sc-event--full.html.twig
piotrsmykaj: i had pending "review" so you didn't see my comments. Now you should see it, right?

templates/paragraphs/paragraph--oe-description-list.html.twig
piotrsmykaj: please check my comment above https://github.com/openeuropa/oe_whitelabel/pull/433#discussion_r3148199716

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.module
piotrsmykaj: > Single-Directory Components allow to specify enum properties, which restricts the possible values in the element to our list. 
Source: https://www.drupal.org/node/3519574

So, I prefer passing a value that respects the schema than thinking or wondering how things make happen when the value is out of the range.

templates/content/node--oe-sc-event--full.html.twig
donquixote: Your observation is correct, and I was not aware of it. Now we have a contradiction of what actually exists in Drupal core templates (and ECL, BCL, or other places we could look at) and what is recommended in documentation.

Imo we should stick with old syntax (`{% include ... with ... %}`) until we see Drupal core changing to `include()`.

templates/overrides/navigation/links--language-block.html.twig
donquixote: I would prefer a separate branch/issue/PR if this needs to be fixed independently of sdc conversion.

templates/patterns/modal/modal-language.html.twig
donquixote: again, let's keep this scope focused on the specific sdc conversion.

Other fixes should go in their own branch. Also because those can be merged in 1.x whereas the work here would sit in an epic for a while.

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.module
donquixote: Alright.
It seems to me the same change would be beneficial already for the pattern even before going for sdc.
But since this is about description list, we can consider it to be "in scope" of this PR.
So ok to keep.

PR #434, opened 2026-04-28
OEL-4510: Minimal upgrade Drupal 11.3

donquixote:
We only add this too see what happens.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,7 +40,7 @@ jobs:
     strategy:
       matrix:
         php_version: ["8.3"]
-        core_version: ["11.2.0"]
+        core_version: ["11.3.0"]
     env:
       PHP_VERSION: ${{ matrix.php_version }}
     steps:
--- a/tests/src/FunctionalJavascript/ContentProjectRenderTest.php
+++ b/tests/src/FunctionalJavascript/ContentProjectRenderTest.php
@@ -427,6 +427,9 @@ class ContentProjectRenderTest extends WebDriverTestBase {
         'value' => '2000-05-10',
         'end_value' => '2010-05-15',
       ],
+      // The /node route lists promoted content.
+      // Prior to Drupal 11.3, these nodes would be promoted by default.
+      'promote' => 1,
     ]);
 
     // In order to render a teaser, we use the default node list view.
@@ -442,6 +445,7 @@ class ContentProjectRenderTest extends WebDriverTestBase {
         'value' => '2010-05-10',
         'end_value' => '2100-05-15',
       ],
+      'promote' => 1,
     ]);
     // And a planned project.
     $this->createProjectNode([
@@ -449,6 +453,7 @@ class ContentProjectRenderTest extends WebDriverTestBase {
         'value' => '2100-05-10',
         'end_value' => '2200-05-15',
       ],
+      'promote' => 1,
     ]);
 
     $this->drupalGet('/node');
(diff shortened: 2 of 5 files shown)


tests/src/FunctionalJavascript/ContentProjectRenderTest.php
```diff
@@ -427,6 +427,9 @@ public function testTeaserStatusBadge(): void {
         'value' => '2000-05-10',
         'end_value' => '2010-05-15',
       ],
+      // The /node route lists promoted content.
+      // Prior to Drupal 11.3, these nodes would be promoted by default.
+      'promote' => 0,
```
donquixote: @tibi2303 You had this in the NodeCreationTrait, but I think it is better here, because we want to keep that trait generic for other tests.

tests/src/Kernel/ContentRenderTestBase.php
```diff
@@ -39,6 +39,7 @@ abstract class ContentRenderTestBase extends AbstractKernelTestBase {
     'link',
     'media',
     'node',
+    'node_storage_body_field',
```
donquixote, line 42: @tibi2303 This is already required in composer and as module dependency via oe_content.
We need to explicitly enable it in kernel tests because these do not automatically install module dependencies.

tests/src/Kernel/MultilingualBlockTest.php
```diff
@@ -26,6 +26,7 @@ class MultilingualBlockTest extends KernelTestBase {
     'ui_patterns',
     'ui_patterns_library',
     'ui_patterns_settings',
+    'user',
```
donquixote, line 29: @tibi2303 You originally had user-related steps in the `setUp()` method.
I found that these were not needed to make the test pass.

.github/workflows/release.yml
```diff
@@ -40,7 +40,7 @@ jobs:
     strategy:
       matrix:
         php_version: ["8.3"]
-        core_version: ["11.2.0"]
+        core_version: ["11.3.0"]
```
AaronGilMartinez, line 43: I think 11.2 security support ends in 1 month and 2 weeks, I assume this was done to keep test execution shorter, shouldn't be a problem for this changes but we could have breaking changes (other) for previous minor 11.2.

AaronGilMartinez (changes requested):

.github/workflows/release.yml
donquixote: Good catch. I took this from the other PR without questioning.

donquixote: Actually.. this is to build the release, this is not CI.

AaronGilMartinez: True, gitlab triggers correct versions.

AaronGilMartinez (approved):

tests/src/FunctionalJavascript/ContentProjectRenderTest.php
tibi2303: ok, thanks

tests/src/Kernel/MultilingualBlockTest.php
tibi2303: ah okay, thx

tests/src/Kernel/ContentRenderTestBase.php
tibi2303: yeah, i was putting it just until the oe_content upgrade. thanks

PR #440, opened 2026-05-13
OEL-4458: Support multiple site owners (v2 with line breaks)

donquixote:
- Supercedes #407

### DIFF
--- a/templates/overrides/navigation/oe-corporate-blocks-neutral-footer.html.twig
+++ b/templates/overrides/navigation/oe-corporate-blocks-neutral-footer.html.twig
@@ -69,7 +69,10 @@
 
 {% set footer_top_left_content %}
   <p class="fw-bold mb-2">{{ site_specific_footer.site_identity }}</p>
-  <p> {{ site_owner ? 'This site is managed by the @name'|t({'@name': site_owner}) }} </p>
+
+  {% if site_owners %}
+    <p>{{ 'This site is managed by the @name'|t({'@name': site_owners|join(', ')}) }}</p>
+  {% endif %}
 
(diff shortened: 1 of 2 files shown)


templates/overrides/navigation/oe-corporate-blocks-neutral-footer.html.twig
```diff
@@ -69,7 +69,14 @@
 
 {% set footer_top_left_content %}
   <p class="fw-bold mb-2">{{ site_specific_footer.site_identity }}</p>
-  <p> {{ site_owner ? 'This site is managed by the @name'|t({'@name': site_owner}) }} </p>
+
+  {% if site_owner %}
+    {% if site_owner|length > 1 %}
+      <p>{{ 'This site is co-managed by:'|t }}<br>{{ site_owner|safe_join('<br>') }}</p>
```
donquixote, line 75: The result of `safe_join()` must be printed and captured.
It cannot be directly assigned with `=` or passed as replacement argument to `t(...)`, or the html inside will be escaped again.

templates/overrides/navigation/oe-corporate-blocks-neutral-footer.html.twig
```diff
@@ -69,7 +69,14 @@
...
+    {% else %}
+      <p>{{ 'This site is managed by the @name'|t({'@name': site_owner|first}) }}</p>
```
donquixote, line 77: For singular, we are keeping the old text which includes " the ", and we do not insert a line break, simply because we have not received any request to change it.

AaronGilMartinez (approved):

PR #441, opened 2026-05-19
OEL-4748: Add bluesky and mastodon icons.

AaronGilMartinez:


piotrsmykaj (dismissed):

piotrsmykaj (approved):

piotrsmykaj (dismissed):

tests/src/Kernel/PreprocessOrderTest.php
```diff
@@ -175,8 +175,17 @@ protected function assertPreprocessFunctions(array $expected, string $hook): voi
     if (version_compare(\Drupal::VERSION, '11.0', '<')) {
       $expected = ['template_preprocess', ...$expected];
     }
+    $preprocess_functions = $info['preprocess functions'] ?? [];
```
tibi2303, line 178: can we check that it's also the expected preprocess callback? I see here it's only checking that exists

composer.json
```diff
@@ -12,7 +12,7 @@
         "drupal/daterange_compact": "^2.0.1",
         "drupal/twig_field_value": "^2.0.2",
         "openeuropa/composer-artifacts": "^1.0.0-alpha1",
-        "openeuropa/oe_bootstrap_theme": "^1.35"
+        "openeuropa/oe_bootstrap_theme": "^1.37.1"
```
tibi2303: ^1.37 is enough i think

piotrsmykaj: ~~ok as we agreed let's revert to ^1.35~~
as we agreed  let's keep ^1.37.1 to enforce dependencies up-to-date for the whole release.

tests/src/Kernel/PreprocessOrderTest.php
piotrsmykaj: good catch

tibi2303 (approved):

piotrsmykaj (approved):

PR #443, opened 2026-05-20
OEL-4736: Add media copyright to content banners.

tibi2303:


tests/src/Functional/WhitelabelBrowserTestBase.php
```diff
@@ -34,4 +36,33 @@ protected function setUp(): void {
     $this->setUpSparql();
   }
 
+  /**
+   * Creates the media copyright field and attaches it to media bundles.
+   *
+   * @param array $bundles
+   *   The media bundles to attach the field to.
+   */
+  protected function createMediaCopyrightField(array $bundles = ['image']): void {
```
piotrsmykaj: 1) I think is better to move the method into MediaCreationTrait to make it visible also in the WebDriverTestBase (used by the ContentProjectRenderTest).
2) We can also cover the Project content type by adding a copyright assertion into the ContentProjectRenderTest.

piotrsmykaj (approved):

PR #448, opened 2026-06-03
OEL-4819: Move preprocess functions from module to theme

donquixote:


modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.module
AaronGilMartinez, line 1: I still see some hook preprocess left in this file like:
- oe_whitelabel_paragraphs_preprocess_paragraph__oe_gallery
- oe_whitelabel_paragraphs_preprocess_paragraph__oe_document
- oe_whitelabel_paragraphs_preprocess_paragraph__oe_quote

Is there a reason to keep them?

AaronGilMartinez (changes requested):

modules/oe_whitelabel_paragraphs/oe_whitelabel_paragraphs.module
donquixote: Yes.
Here the prefix is `oe_whitelabel_paragraphs_(preprocess_)`, not `oe_whitelabel_(preprocess_)`.
This means they are "owned" by the module, not the theme.

We can disagree with that, but this is replicating existing behavior.
See how `'oe_whitelabel_paragraphs_preprocess_paragraph__oe_gallery'` behaves in the test case.

donquixote: So in a follow-up we could either move everything back to the module and rename, and use a mechanism to only have them fire when the theme is active.

Or we do the reverse and move the remaining ones to the theme, rename, and add the attribute.

AaronGilMartinez: Ok, agreed to leave It as as is for the moment.

AaronGilMartinez (approved):

PR #454, opened 2026-08-14
OEL-4973: Harden npm installs and update theme dependencies.

tibi2303:


.npmrc
AaronGilMartinez: I think we can apply same rule to kits/default/.npmrc.

composer.json
```diff
@@ -75,11 +75,21 @@
     },
     "extra": {
         "patches": {
+            "drupal/entity_browser": {
```
AaronGilMartinez: I faced the same issue in update to 11.4, there I patched oe_content and pointed to the branch: OEL-4912
We can do this or wait, depends on urgency of these changes.

AaronGilMartinez (changes requested):

.npmrc
AaronGilMartinez: This for oe_bootstrap_theme, not here.

composer.json
tibi2303: We need this until the end of the month i think - maybe not though.. not sure when they will require the update of toolkit to ignore release/build scripts.. might not be this month because it's to short. Let's wait for oe_content.

AaronGilMartinez: We can drop this: https://github.com/openeuropa/oe_content/releases/tag/4.15.0

AaronGilMartinez (approved):

PR #456, opened 2026-08-24
OEL-4907: Update to Drupal 11.4.x

piotrsmykaj:


tests/src/FunctionalJavascript/SearchBlockAppearanceTest.php
```diff
@@ -126,6 +126,8 @@ protected function doTestSearchBlockInOtherRegions(): void {
       ->set('region', 'header')
       ->set('settings.form.region', 'header')
       ->save();
+    \Drupal::service('cache.page')->deleteAll();
+    \Drupal::service('cache.dynamic_page_cache')->deleteAll();
```
AaronGilMartinez: Can you check why?

tests/src/Kernel/FooterBlockTest.php
```diff
@@ -109,6 +109,7 @@ public function testEcFooterBlockRendering(): void {
       ->save();
 
     $builder->resetCache();
+    \Drupal::service('cache.render')->deleteAll();
```
AaronGilMartinez: Same here, why we need to reset cache?

composer.json
```diff
@@ -17,36 +17,36 @@
...
         "openeuropa/oe_content_extra": "^1.4",
-        "openeuropa/oe_corporate_blocks": "^5.0.0-beta5",
+        "openeuropa/oe_corporate_blocks": "^5.2.1",
         "openeuropa/oe_link_lists": "^1.0",
         "openeuropa/oe_list_pages": "^1.0",
         "openeuropa/oe_multilingual": "^1.14",
         "openeuropa/oe_paragraphs": "^1.33",
         "openeuropa/oe_starter_content": "^1.5",
-        "openeuropa/oe_subscriptions": "^1.0.0-beta3"
+        "openeuropa/oe_subscriptions": "^1.0.0-beta4"
```
AaronGilMartinez: Why updated minimum on those packages?

AaronGilMartinez (changes requested):

tests/src/FunctionalJavascript/SearchBlockAppearanceTest.php
piotrsmykaj: Drupal 11.4 (issue [#3341042](https://www.drupal.org/project/drupal/issues/3341042)) narrowed block cache tag invalidation to config:block_list. Using $this->config() will try to invalidate the same cache tag as the block name, but nothing is tagged with it anymore. Since invalidation doesn't reach the cached page, we clear the page caches manually.

An alternative would be to use the block entity API entity instead of the Config API, where the config:block_list. tag will be invalidated, and then custom cache clearing will not be needed. Let me know if you want to change it.

tests/src/Kernel/FooterBlockTest.php
piotrsmykaj: Drupal 11.4 (issue [#3341042](https://www.drupal.org/project/drupal/issues/3341042)) overrides BlockViewBuilder::getCacheTags():

```
  public function getCacheTags() {
    // Blocks do not use a view cache tag...
    return [];
  }
```
so `$builder->resetCache()` without arguments, invalidates nothing, but adding an $entity as argument, it uses $entity->getCacheTags(). Good you asked and I checked deeper, and used a better solution for it.

composer.json
piotrsmykaj: I only took care about alpha/beta versions to check if newer version exists, but if you prefer, I can update more require-dev packages. Let me know.

AaronGilMartinez (approved):

PR #457, opened 2026-08-26
OEL-4984: Respect the configured Slim Select version.

tibi2303:


modules/oe_whitelabel_helper/oe_whitelabel_helper.post_update.php
```diff
@@ -67,3 +67,17 @@ function oe_whitelabel_helper_post_update_00004(): string {
 
   return $report . "\nThe old navigation block was disabled.";
 }
+
+/**
+ * Preserve the effective Slim Select version used by existing sites.
+ */
+function oe_whitelabel_helper_post_update_00005(): string {
+  $config = \Drupal::configFactory()->getEditable('slim_select.settings');
+  if ($config->get('version') !== 'v2.10.0') {
```
AaronGilMartinez: This could be anything, like 2.9.0 anv would pass, to specific anyway not sure if we should change the version.

PR #461, opened 2026-09-03
OEL-4961: Increase icon size for Facts and figures

piotrsmykaj:


config/install/oe_whitelabel.settings.yml
```diff
@@ -10,6 +10,8 @@ logo:
 bootstrap_tables:
   enable: true
   responsive: always
+facts_figures:
+  icon_size: 3xl
```
AaronGilMartinez: Where is the setting used?
I think there is a test for settings too, can you check?

modules/oe_whitelabel_paragraphs/tests/src/Kernel/Paragraphs/FactsFiguresTest.php
```diff
@@ -82,7 +82,7 @@ public function testRendering(): void {
     $this->assertCount(1, $crawler->filter('div.bcl-fact-figures.bcl-fact-figures--default'));
     $this->assertCount(1, $crawler->filter('h2.bcl-heading'));
     $this->assertCount(1, $crawler->filter('div.row-cols-md-3.row'));
-    $this->assertCount(6, $crawler->filter('svg.bi.icon--l'));
+    $this->assertCount(6, $crawler->filter('svg.bi.icon--3xl'));
```
AaronGilMartinez: This works because BCL applies a default, since default is set at BCL level any icon that does not declare a size will have 3xl.

AaronGilMartinez (changes requested):

config/install/oe_whitelabel.settings.yml
AaronGilMartinez: My bad, I didnt see oe_bootstrap_theme changes.

AaronGilMartinez (approved):

AaronGilMartinez (approved):

